PR #9, opened 2025-10-14
GUM-140: now supporting the assetId parameter in GET /albums

taggartgorman:
OK, final PR - I swear!

The assetId parameter in GET /albums is used by Immich on the asset detail page to display the albums that contain the asset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,12 +180,24 @@ This is useful for:
 
 ### Immich API Integration
 
+Defining endpoint parameters:
+
+- Use `Annotated` to specify attributes, such as `Query()`, `Path()`, `Body()` functions, or numeric or string validations, but do not use `Default` - the default value should be specified as part of the Python declaration.
+- If a parameter is not required, use `| SkipJsonSchema[None]` after defining the type to allow Pydantic to accept the `None` type, but prevent `None` from being exposed in the OpenAPI schema. 
+- If the exposed parameter name needs to be camelCase, use `alias="camelCase"` within the function and then use an appropriate snake_case name for the parameter in the function signature.
+
+Example:
+```python
+asset_id: Annotated[UUID | SkipJsonSchema[None], Query(alias="assetId")] = None,
+```
+
 When implementing new endpoints:
 
 1. **Generate models**: Use `generate_immich_models.py` to create up-to-date Pydantic models
 2. **Import models**: Use generated models from `routers.immich_models` for type safety
-3. **Validate compatibility**: Run `validate_api_compatibility.py` to ensure correct implementation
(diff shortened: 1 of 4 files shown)


taggartgorman:
@ternarybits The issue Charlie flagged with `asset_id` being typed as `UUID` but then having a default of `None` is similar to issues I encountered when I first worked on the adapter: Pyright and the OpenAPI validation tool do not agree at times.

If we change the code to
`asset_id: UUID | None = Query(default=None, alias="assetId"),`
then Pyright is happy, but the validation tool will flag the the endpoint with a warning of "Parameter 'assetId' schema differs".

The only way I could keep both tools happy is how the code is - no optional type, and `# type: ignore` in the tests. I felt that preventing the validator warning was more important than making Pyright happy as both the tests and live code work, but I'd like your input.

ternarybits:
Okay, let me see if I understand what's going on here.  `get_all_albums` is an endpoint that takes an optional `asset_id` and an optional `shared`.  It corresponds to this endpoint in the Immich API documentation: https://api.immich.app/endpoints/albums/getAllAlbums

In the Immich OpenAPI specification, this is how it's defined:

```
    "/albums": {
      "get": {
        "operationId": "getAllAlbums",
        "parameters": [
          {
            "name": "assetId",
            "required": false,
            "in": "query",
            "description": "Only returns albums that contain the asset\nIgnores the shared parameter\nundefined: get all albums",
            "schema": {
              "format": "uuid",
              "type": "string"
            }
          },
          {
            "name": "shared",
            "required": false,
            "in": "query",
            "schema": {
              "type": "boolean"
            }
          }
        ],
```

Based on that, it sounds like our own endpoint should have `asset_id` defined as 
```
asset_id: UUID | None = Query(default=None, alias="assetId"),
```

What's the difference that the validation tool flags?

ternarybits:
Also, I just looked at the Immich API, and it looks like all of the API parameters are supposed to be camelCased.  In this PR, you changed one to be snake_cased.  We should probably put something in AGENTS.md about this, because it's easy to mess up.

taggartgorman:
@ternarybits 

> Also, I just looked at the Immich API, and it looks like all of the API parameters are supposed to be camelCased. In this PR, you changed one to be snake_cased. We should probably put something in AGENTS.md about this, because it's easy to mess up.

`asset_id: Annotated[UUID | SkipJsonSchema[None], Query(alias="assetId")] = None,`

The alias attribute is what the openAPI generator uses, so this allows us to use snake_case for our internal code and camelCase for the external API.

ternarybits (approved):

PR #13, opened 2025-10-27
GUM-152: Change adapter to use dependency injection for gumnut client

taggartgorman:


routers/utils/gumnut_client.py
```diff
@@ -1,25 +1,188 @@
...
+
+    Returns:
+        Gumnut: Configured Gumnut client instance with user's JWT
     """
     settings = get_settings()
-    api_key = settings.gumnut_api_key
 
-    if not api_key:
-        raise HTTPException(status_code=500, detail="GUMNUT_API_KEY not configured")
+    return Gumnut(
```
ternarybits: Am I missing something?  I don't see `http_client` as an argument to the Gumnut constructor...

ternarybits (approved):

PR #19, opened 2025-11-12
GUM-158: Enabling mobile OAuth

taggartgorman:
Fixes to allow mobile OAuth:
- Implemented mobile redirect endpoint (`api/oauth/mobile-redirect`) to support OAuth providers that do not support Immich's custom app scheme of `app.immich:///oauth-callback` in the redirect URI.
- Added uvicorn settings `--timeout-keep-alive`, `--limit-concurrency`, and `--backlog` to better support mobile clients. `--timeout-keep-alive` is particularly important as it prevents "Connection closed before full header was received" errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -211,3 +211,6 @@ __marimo__/
 static/*
 # But keep the directory itself
 !static/.gitkeep
+
+# PEM files should be ignored for security reasons
+*.pem
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,4 +103,8 @@ ENV LOG_LEVEL=info
 # Run the application with optimized uvicorn settings
 # Render sets PORT environment variable automatically (typically 10000)
 # Using exec form with shell for proper signal handling and variable substitution
-CMD ["sh", "-c", "uvicorn main:app --host 0.0.0.0 --port ${PORT:-8080} --log-level ${LOG_LEVEL} --timeout-graceful-shutdown 60 --timeout-keep-alive 75 --limit-concurrency 1000 --backlog 2048"]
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -210,4 +210,7 @@ __marimo__/
 # Ignore contents
 static/*
 # But keep the directory itself
-!static/.gitkeep
\ No newline at end of file
+!static/.gitkeep
+
+# PEM files should be ignored for security reasons
+*.pem
```
ternarybits: This addition doesn't seem related to the rest of the PR.

Dockerfile
```diff
@@ -103,4 +103,8 @@ ENV LOG_LEVEL=info
 # Run the application with optimized uvicorn settings
 # Render sets PORT environment variable automatically (typically 10000)
 # Using exec form with shell for proper signal handling and variable substitution
-CMD ["sh", "-c", "uvicorn main:app --host 0.0.0.0 --port ${PORT:-8080} --log-level ${LOG_LEVEL} --timeout-graceful-shutdown 60"]
+CMD ["sh", "-c", "uvicorn main:app --host 0.0.0.0 --port ${PORT:-8080} --log-level ${LOG_LEVEL} \
```
ternarybits: Just noticed that this is calling uvicorn directly.  Does it not work to call `fastapi` instead?

ternarybits (approved):

.gitignore
taggartgorman: In the process of working on this feature, I needed to generate local certs in order to run https on my local machine. I didn't want them to get accidentally checked in.

PR #20, opened 2025-11-12
GUM-158: Handle Render.com reverse proxy in redirect URI rewriting

taggartgorman:
Render uses a reverse proxy to handle HTTPS, so we need to check x-forwarded-* headers when rewriting the mobile redirect URI.


ternarybits (approved):
I think there should be a test to go along with this logic?

PR #26, opened 2025-11-19
GUM-154: primitive version of mobile sync - no checkpoints, always does full sync

taggartgorman:
This implementation is by no means meant to be complete. I have not tested it on large albums.


ternarybits (approved):
Looks fine for a v1.

PR #121, opened 2026-04-08
GUM-324: Return real asset ID for duplicate uploads

ternarybits:
## Summary
- Fix `POST /assets` to return the real asset ID (not a placeholder UUID) when a duplicate is detected
- Buffered path: use `with_raw_response.create()` to detect HTTP 200 (duplicate) vs 201 (new) from photos-api
- Streaming path: propagate HTTP status code from the pipeline so the adapter can detect duplicates
- Remove dead "duplicate"/"already exists" string-matching branch from `_handle_upload_error` — photos-api returns HTTP 200 for duplicates (not errors), so this code path was unreachable

## Linear
https://linear.app/gumnut-ai/issue/GUM-324/post-assets-duplicate-response-returns-placeholder-uuid-instead-of

## Test plan
- [x] Buffered path: new asset returns 201 + "created" status
- [x] Buffered path: duplicate returns 200 + "duplicate" status with real asset ID
- [x] Streaming path: duplicate returns 200 + "duplicate" status with real asset ID
- [x] All 647 tests pass
- [x] Pyright type check clean, ruff lint/format clean

### DIFF
--- a/routers/api/assets.py
+++ b/routers/api/assets.py
@@ -294,9 +294,18 @@ def _handle_upload_error(e: Exception) -> AssetMediaResponseDto | JSONResponse:
     """Map upload exceptions to Immich-compatible HTTP responses."""
     error_msg = str(e).lower()
     if "duplicate" in error_msg or "already exists" in error_msg:
+        # Try to extract the real asset ID from the exception body (SDK errors
+        # preserve the API response). Fall back to a placeholder UUID.
+        dup_id = "00000000-0000-0000-0000-000000000000"
+        body = getattr(e, "body", None)
+        if isinstance(body, dict) and body.get("id"):
+            try:
+                dup_id = str(safe_uuid_from_asset_id(body["id"]))
+            except Exception:
+                pass
         return JSONResponse(
             content={
-                "id": "00000000-0000-0000-0000-000000000000",
+                "id": dup_id,
                 "status": AssetMediaStatus.duplicate.value,
             },
             status_code=status.HTTP_200_OK,
@@ -471,7 +480,7 @@ async def _upload_buffered(
                 span.set_data("upload.filename", asset_data.filename)
                 span.set_data("upload.content_type", asset_data.content_type)
                 span.set_data("upload.strategy", "buffered")
-                gumnut_asset = await client.assets.create(
+                raw_response = await client.assets.with_raw_response.create(
                     asset_data=(
                         asset_data.filename,
                         asset_data.file,
@@ -483,7 +492,18 @@ async def _upload_buffered(
                     file_modified_at=file_modified_at,
                 )
 
+            gumnut_asset = await raw_response.parse()
             asset_uuid = safe_uuid_from_asset_id(gumnut_asset.id)
+
+            if raw_response.status_code == 200:
+                return JSONResponse(
+                    content={
+                        "id": str(asset_uuid),
+                        "status": AssetMediaStatus.duplicate.value,
+                    },
+                    status_code=status.HTTP_200_OK,
+                )
+
             await _emit_upload_events(gumnut_asset, current_user)
 
             return AssetMediaResponseDto(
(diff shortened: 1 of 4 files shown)


taggartgorman (approved):
LGTM — clean fix for GUM-324. Returning the real asset ID on duplicate is the right behavior, and the three-pronged approach (buffered raw-response, streaming `_http_status` propagation, error-body fallback) covers all the paths a duplicate can surface through.

## Pre-approve checklist
- [x] **Security/auth**: untouched — changes confined to upload response handling
- [x] **Payments/billing**: untouched
- [x] **Data deletion**: untouched
- [x] **Confidence**: High — each branch (201 success, 200 duplicate buffered, 200 duplicate streaming, error-body extraction, placeholder fallback) has direct unit coverage
- [x] **Purpose clarity**: Clear from PR body and linked Linear issue

Nice touch including the error-handler fallback — covers the case where photos-api surfaces the duplicate as an exception rather than a 200 response.

routers/api/assets.py
```diff
@@ -483,7 +492,18 @@ async def _upload_buffered(
                     file_modified_at=file_modified_at,
                 )
 
+            gumnut_asset = await raw_response.parse()
             asset_uuid = safe_uuid_from_asset_id(gumnut_asset.id)
+
+            if raw_response.status_code == 200:
```
taggartgorman: Use constants (HTTP_200_OK), not hard coded numbers (200) for response codes.

routers/api/assets.py
```diff
@@ -532,9 +552,9 @@ async def _upload_streaming(
         result = await pipeline.execute(_extract_upload_fields)
 
         asset_id = result.get("id", "")
-        asset_status = result.get("status", "created")
+        http_status = result.pop("_http_status", 201)
```
taggartgorman: Use a HTTP constant here, too.

routers/api/assets.py
```diff
@@ -532,9 +552,9 @@ async def _upload_streaming(
...
 
-        if asset_status == AssetMediaStatus.duplicate.value:
+        if http_status == 200:
```
taggartgorman: Use a HTTP constant here, too.

taggartgorman (changes requested):

taggartgorman:
The `/review-pr` skill jumped the gun and approved the PR, but it didn't catch the HTTP_200_OK vs 200 issue. I'm wondering if low Claude effort during the morning crunch time is resulting in lower quality reviews.

ternarybits:
## Feedback addressed

### Auto-fixed (two-way door)
- **HTTP constants** (@taggartgorman): Replaced bare `200`/`201` with `status.HTTP_200_OK` / `status.HTTP_201_CREATED` in buffered and streaming duplicate detection
- **Non-dict guard** (@charliecreates): Added `isinstance(result, dict)` guard before injecting `_http_status` into streaming pipeline result

### Skipped
- "Buffered path treats non-200 as created" — SDK validates response against `AssetResponse` schema; unexpected 2xx would raise during `.parse()`
- "Streaming relies solely on `_http_status`" — strictly better than the previous `result.get("status", "created")` which never worked (photos-api response has no `status` field). Default `HTTP_201_CREATED` is correct
- "Error body might be str/bytes" — Stainless SDK `.body` is already parsed JSON; `isinstance(body, dict)` handles this
- "Test brittleness with byte comparison" — matches existing patterns in the file; `JSONResponse` key ordering is deterministic

services/streaming_upload.py
```diff
@@ -288,7 +288,9 @@ def _sync_upload(
         if new_token:
             self.refreshed_token = new_token
 
-        return response.json()
+        result = response.json()
+        result["_http_status"] = response.status_code
+        return result
```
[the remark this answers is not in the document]
ternarybits: Fixed — added `isinstance(result, dict)` guard before injecting `_http_status`. Non-dict payloads are wrapped in `{"_http_status": ..., "data": ...}`.

routers/api/assets.py
ternarybits: Fixed — using `status.HTTP_200_OK` now.

routers/api/assets.py
ternarybits: Fixed — using `status.HTTP_201_CREATED` for the default and `status.HTTP_200_OK` for the comparison.

routers/api/assets.py
ternarybits: Fixed — covered by the same change as the thread above.

ternarybits:
Closing in favor of a new PR with cleaner history after substantial changes during review.

PR #123, opened 2026-04-13
Implement face and search stub endpoints with Gumnut SDK

ternarybits:
## Summary
- Implement face delete and reassign endpoints using `client.faces.delete()` and `client.faces.update()`
- Implement person search (`GET /api/search/person`) using `client.people.list(name=...)`
- Implement metadata search (`POST /api/search/metadata`) using `client.search.search()` with date range and person ID filters
- Implement search statistics (`POST /api/search/statistics`) using `client.assets.counts()`
- Update adapter architecture docs to reflect new endpoint status

## Linear
https://linear.app/gumnut-ai/issue/GUM-502/implement-face-and-search-stub-endpoints-in-immich-adapter

## Test plan
- [ ] Verify face delete removes a face via the adapter
- [ ] Verify face reassign moves a face to a different person
- [ ] Verify person search returns matching people by name
- [ ] Verify metadata search returns assets filtered by date range and person IDs
- [ ] Verify search statistics returns correct total asset count
- [ ] All existing tests pass (646/646)


ternarybits:
## Feedback Round 1

### Auto-fixed (two-way door)
- **DELETE face body required** (`faces.py:47`): Made `request: AssetFaceDeleteDto` optional with `Body(default=None)`. The parameter was unused and requiring it would cause 422 errors for Immich clients that don't send a body. Added test for no-body case.

### Skipped
- **Metadata search pagination** (`total`/`nextPage`): SDK doesn't provide total count or next-page tokens. Matches the existing `search_smart` pattern (line 223-231).
- **Statistics ignores request filters**: `_fetch_asset_counts` supports only single `album_id`/`person_id`, while `StatisticsSearchDto` expects lists. Most filters have no SDK equivalent. Adding partial support would be misleading.
- **People list unbounded scan**: Standard codebase pattern (same as `faces.list`). Name search results are naturally bounded by matching people count.
- **Statistics hard-coded limit**: Already fixed in self-review commit `b38c188` — now uses `_fetch_asset_counts` with proper pagination.

routers/api/faces.py
```diff
@@ -32,37 +33,35 @@
...
-    Deletes a specific face by ID.
-    This is a stub implementation that returns a empty response.
-    """
-    return
+    """Deletes a specific face by ID."""
+    try:
+        gumnut_face_id = uuid_to_gumnut_face_id(id)
+        await client.faces.delete(gumnut_face_id)
+    except Exception as e:
+        raise map_gumnut_error(e, "Failed to delete face") from e
```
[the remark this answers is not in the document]
ternarybits: Fixed — made `request` optional with `Body(default=None)`. Added test for the no-body case.

ternarybits:
## Feedback Round 2

### Skipped
- **Move `_fetch_asset_counts` to shared utility** (`search.py:150`): Moving a single function to a new file isn't proportionate for the coupling concern. The failing-fast suggestion (`raise 422 when filters provided`) would break Immich clients — they send the full `StatisticsSearchDto` and expect a response.
- **`withHidden` type annotation**: `bool = Query(default=None)` is a pervasive codebase-wide pattern (25+ instances across 7 files). Fixing one instance would be inconsistent; a codebase-wide cleanup is a separate task.
- **Statistics tests coupled to timeline internals**: Mocking `client.assets.counts` directly tests the full integration path more realistically than patching the helper function.
- **Docs "Fully implemented" vs gaps**: The docs list what's implemented vs. stubbed. The "Fully implemented" section header means "has real SDK calls" (not stubs). Notes already clarify per-area limitations.
- **Metadata search page/size 500s**: Pydantic validates `size` and `page` as `int | None` before the handler runs — invalid values are rejected at the framework level.
- **People search unbounded**: Repeat from round 1 — standard codebase pattern, name search naturally bounded.

routers/api/search.py
```diff
@@ -155,23 +142,60 @@ async def search_asset_statistics(
...
-    This is a stub implementation that returns zero counts.
-    """
-    return SearchStatisticsResponseDto(total=0)
+    """Get asset count statistics."""
+    try:
+        from routers.api.timeline import _fetch_asset_counts
+
+        buckets = await _fetch_asset_counts(client)
+        total = sum(bucket.count for bucket in buckets)
+        return SearchStatisticsResponseDto(total=total)
```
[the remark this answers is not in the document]
ternarybits: Skipping both suggestions:

1. **Shared utility**: Moving a single function to a new file isn't proportionate. The import was moved to the top of the file (commit 326006c) to follow the top-of-file import convention. The cross-router import is pragmatic for a single helper.

2. **Fail fast on filters**: Immich clients send the full `StatisticsSearchDto` body and expect a response. Returning 422 when filters are present would break client compatibility. The current behavior (returning unfiltered total) is more graceful.

routers/api/search.py
```diff
@@ -155,23 +143,58 @@ async def search_asset_statistics(
...
+                )
+
+        return SearchResponseDto(
+            albums=SearchAlbumResponseDto(count=0, facets=[], items=[], total=0),
+            assets=SearchAssetResponseDto(
+                count=len(immich_assets),
+                facets=[],
+                items=immich_assets,
+                nextPage="",
+                total=len(immich_assets),
```
taggartgorman, line 196: **[Should Fix]** Metadata search pagination is broken

`total=len(immich_assets)` only reports the current page count, not the true total. And `nextPage=""` is always empty, so clients can never load subsequent pages.

For example, if there are 200 matching assets with `limit=50`, the client sees "50 results" and has no way to load more.

If the SDK exposes `total` or `has_more` on the response, use those to populate these fields. If not, consider documenting this as a known limitation and filing a follow-up.

tests/unit/api/test_search.py
```diff
@@ -0,0 +1,280 @@
...
+
+        request = StatisticsSearchDto()
+        with pytest.raises(HTTPException) as exc_info:
+            await search_asset_statistics(request=request, client=mock_client)
+
+        assert exc_info.value.status_code == 500
+        assert "Failed to get search statistics" in exc_info.value.detail
+
+
+class TestSearchMetadata:
```
taggartgorman, line 190: **[Should Fix]** No test exercises the asset conversion path

All four `TestSearchMetadata` tests supply `search_response.data = []`. The code path that calls `convert_gumnut_asset_to_immich(item.asset, current_user)` is never exercised. A regression in asset conversion or the `item.asset` attribute access would go undetected.

Consider adding a test with a non-empty `search_response.data` containing a mock search result item (with an `.asset` attribute) to verify that converted assets appear in `result.assets.items` and `result.assets.count`.

routers/api/search.py
```diff
@@ -25,6 +27,7 @@
     StatisticsSearchDto,
     UserResponseDto,
 )
+from routers.api.timeline import _fetch_asset_counts
```
taggartgorman: **[Consider]** Cross-module import of private function

`_fetch_asset_counts` has a leading underscore signaling "module-private" by Python convention. Importing it from another router module creates coupling between search and timeline.

Suggestion: either move to a shared utility (e.g., `routers/utils/asset_counts.py`) and drop the underscore, or rename it to `fetch_asset_counts` (public) in `timeline.py`.

routers/api/search.py
```diff
@@ -6,7 +6,9 @@
 
 from routers.utils.gumnut_client import get_authenticated_gumnut_client
 from routers.utils.current_user import get_current_user
-from routers.utils.error_mapping import check_for_error_by_code
+from routers.utils.error_mapping import check_for_error_by_code, map_gumnut_error
```
taggartgorman, line 10: **[Consider]** Missing module logger

`search.py` has no `logger = logging.getLogger(__name__)` and no logging calls. The new real implementations perform SDK calls with no diagnostic logging, making failures harder to trace in Sentry. Comparable endpoints in `people.py` and `faces.py` both define module loggers.

Suggestion: add `import logging` and `logger = logging.getLogger(__name__)` at the top, and consider adding debug-level log calls for key operations.

taggartgorman (commented):
Good work replacing stubs with real SDK calls. The implementations follow established codebase patterns well (error mapping, ID conversion, person conversion). A couple of items to address around pagination and test coverage.

ternarybits:
## Feedback Round 3

### Auto-fixed (two-way door)
- **No test for asset conversion** (`test_search.py`): Added `test_converts_search_results_to_immich_assets` with non-empty `search_response.data` that exercises `convert_gumnut_asset_to_immich(item.asset, current_user)`.
- **Missing module logger** (`search.py`): Added `import logging` and `logger = logging.getLogger(__name__)` to match all other router modules.
- **Cross-module private import** (`search.py:30`): Renamed `_fetch_asset_counts` → `fetch_asset_counts` across `timeline.py`, `search.py`, and `test_timeline.py` to make the function explicitly public.

### Skipped
- **Metadata search pagination** (`search.py:193`): SDK doesn't expose `total` or `has_more`. Matches the existing `search_smart` pattern (lines 224-230). Same rationale as Round 1.

routers/api/search.py
ternarybits: Skipping — SDK does not expose `total` or `has_more` on search results. This matches the existing `search_smart` pattern (lines 224-230) which uses the same `total=len(immich_assets)` and `nextPage=""`.

tests/unit/api/test_search.py
ternarybits: Fixed — added `test_converts_search_results_to_immich_assets` with non-empty `search_response.data` containing a mock search result item with `.asset` attribute. Verifies converted assets appear in `result.assets.items` with correct count and fields.

routers/api/search.py
ternarybits: Fixed — renamed `_fetch_asset_counts` → `fetch_asset_counts` across `timeline.py`, `search.py`, and `test_timeline.py` to make the function explicitly public since it is used cross-module.

routers/api/search.py
ternarybits: Fixed — added `import logging` and `logger = logging.getLogger(__name__)` to `search.py`, matching all other router modules.

taggartgorman (approved):

PR #130, opened 2026-04-16
Verify payload FK refs against production (GUM-545)

ternarybits:
## Summary

- Fixes a cross-cycle FK-violation bug where mobile clients with a `PersonV1` checkpoint past a `person_deleted` event still received stale `face_updated` events referencing that person, causing `SqliteException(787) FOREIGN KEY constraint failed` on `asset_face_entity` insert and stuck sync.
- Adds a payload-FK verification step in `_stream_entity_type`: for each face/album batch, payload-referenced FK IDs (`face_updated.person_id`, `album_updated.album_cover_asset_id`) are collected via a new `extract_payload_fk_refs` helper and verified against production. Confirmed 404s populate `stats.not_found_ids`, which is now authoritative regardless of the client's checkpoint state — the old checkpoint-skip guard in `null_deleted_fk_references` is removed.
- Updates the `sync-stream-event-ordering` design doc with a Fix 5 section (including the concrete production timeline that surfaced the bug) and the `sync-stream-architecture` overview's person/cover handling notes.

## Linear

https://linear.app/gumnut-ai/issue/GUM-545/debug-another-sync-issue

## Test plan

- [x] `uv run pytest` — all 669 tests pass
- [x] `uv run ruff format` / `uv run ruff check` clean
- [x] New regression tests assert both face and album cross-cycle scenarios (`test_face_updated_nulls_person_id_when_person_deleted_across_cycles`, `test_album_updated_nulls_cover_when_asset_deleted_across_cycles`), including that `stats.not_found_ids` is populated as expected
- [x] Three existing tests updated whose default empty list mocks would now trigger the new null-out (verification step only treats an ID as 404 when the fetch confirms absence)

### DIFF
--- a/routers/api/sync/fk_integrity.py
+++ b/routers/api/sync/fk_integrity.py
@@ -3,7 +3,9 @@
 import logging
 from collections import defaultdict
 from dataclasses import dataclass, field
-from typing import Any
+from typing import Any, NamedTuple
+
+from gumnut.types.events_response import Data as EventData
 
 from routers.immich_models import SyncEntityType
 from routers.api.sync.types import EntityType
@@ -16,6 +18,41 @@ FK_REFERENCES: dict[str, list[tuple[str, str]]] = {
     "face": [("person_id", "person"), ("asset_id", "asset")],
     "album_asset": [("album_id", "album"), ("asset_id", "asset")],
     "album": [("album_cover_asset_id", "asset")],
+}
+
+
+class PayloadFKOverride(NamedTuple):
+    """Declares a payload key the adapter applies as a causally-consistent FK override.
+
+    event_type: the event_type on which this override is applied (e.g. "face_updated").
+    payload_key: the key in event.payload carrying the referenced ID.
+    referenced_type: the gumnut entity type that the ID refers to (e.g. "person").
+    """
+
+    event_type: str
+    payload_key: str
+    referenced_type: str
+
+
+# Payload keys that the adapter applies as causally-consistent FK overrides
+# (see payload_override). Used to collect referenced IDs up front so we can
+# verify they still exist in production, regardless of whether the referenced
+# entity type is being streamed in this cycle.
+PAYLOAD_FK_OVERRIDES: dict[str, list[PayloadFKOverride]] = {
+    "face": [
+        PayloadFKOverride(
+            event_type="face_updated",
+            payload_key="person_id",
+            referenced_type="person",
+        ),
+    ],
+    "album": [
+        PayloadFKOverride(
+            event_type="album_updated",
+            payload_key="album_cover_asset_id",
+            referenced_type="asset",
+        ),
+    ],
 }
 
 # Map gumnut entity type -> SyncEntityType(s) for FK checkpoint lookups.
@@ -119,15 +156,21 @@ def null_deleted_fk_references(
     gumnut_entity_type: str,
     entity: EntityType,
     stats: SyncStreamStats,
-    checkpoint_map: dict[SyncEntityType, Checkpoint],
     event_type: str,
     cursor: str,
 ) -> EntityType:
     """Null FK fields that reference entities confirmed deleted (404 during fetch).
 
-    Uses FK_REFERENCES to discover which fields to check. Skips fields whose
-    referenced entity type has a checkpoint — the entity may exist on the
-    client from a prior sync cycle.
+    Uses FK_REFERENCES to discover which fields to check. A 404 in
+    ``stats.not_found_ids`` is authoritative for this sync cycle — the entity
+    does not exist in production, so the client must not hold a reference to
+    it regardless of any prior-cycle checkpoint (the client may have processed
+    the delete in an earlier cycle and no longer has the entity locally).
+
+    Callers are responsible for populating ``stats.not_found_ids`` for payload-
+    referenced types that are not otherwise streamed (see
+    ``extract_payload_fk_refs`` and the verification step in
+    ``_stream_entity_type``).
 
     Returns the (possibly updated) entity.
     """
@@ -139,10 +182,6 @@ def null_deleted_fk_references(
     for attr_name, ref_type in refs:
         ref_id = getattr(entity, attr_name, None)
         if ref_id is None:
-            continue
-
-        ref_sync_types = _GUMNUT_TYPE_TO_SYNC_TYPES.get(ref_type)
-        if ref_sync_types and any(t in checkpoint_map for t in ref_sync_types):
             continue
 
         if ref_id in stats.not_found_ids.get(ref_type, set()):
@@ -164,3 +203,36 @@ def null_deleted_fk_references(
         entity = entity.model_copy(update=updates)
 
     return entity
+
+
+def extract_payload_fk_refs(
+    gumnut_entity_type: str,
+    events: list[EventData],
+) -> dict[str, set[str]]:
+    """Collect payload-referenced FK IDs from a batch of events.
+
+    For each event whose payload carries a causally-consistent FK override
+    (see ``PAYLOAD_FK_OVERRIDES``), extract the referenced ID and group by
+    referenced entity type. Returns ``{ref_type: {ref_ids...}}``.
+
+    The caller uses this set to verify those IDs still exist in production
+    before the face/album event is streamed to the client. Without this check
+    the adapter would send references to entities deleted between sync cycles,
+    causing FK violations on the mobile client.
+    """
+    overrides = PAYLOAD_FK_OVERRIDES.get(gumnut_entity_type)
+    if not overrides:
+        return {}
+
+    result: dict[str, set[str]] = defaultdict(set)
+    for event in events:
+        if not isinstance(event.payload, dict):
+            continue
+        for override in overrides:
+            if event.event_type != override.event_type:
+                continue
+            should_apply, value = payload_override(event.payload, override.payload_key)
+            if should_apply and value is not None:
+                result[override.referenced_type].add(value)
+
+    return result
--- a/routers/api/sync/fk_integrity.py
+++ b/routers/api/sync/fk_integrity.py
@@ -38,6 +38,11 @@ class PayloadFKOverride(NamedTuple):
 # (see payload_override). Used to collect referenced IDs up front so we can
 # verify they still exist in production, regardless of whether the referenced
 # entity type is being streamed in this cycle.
+#
+# Invariant: every (event_type, payload_key, referenced_type) entry here must
+# have a matching (field=payload_key, referenced_type) entry in FK_REFERENCES
+# for the same entity type — otherwise null_deleted_fk_references will not
+# null the payload-overridden field when its referenced entity 404s.
 PAYLOAD_FK_OVERRIDES: dict[str, list[PayloadFKOverride]] = {
     "face": [
         PayloadFKOverride(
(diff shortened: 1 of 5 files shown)


taggartgorman:
@ternarybits Charlie never added a review to this PR.

routers/api/sync/fk_integrity.py
```diff
@@ -141,10 +184,6 @@ def null_deleted_fk_references(
         if ref_id is None:
             continue
```
taggartgorman, line 191: **[Consider]** Nulling can hit non-nullable FK fields now that the checkpoint guard is gone

`FK_REFERENCES` contains non-nullable attributes — verified via Pydantic introspection: `FaceResponse.asset_id: str`, `AlbumAssetResponse.album_id: str`, `AlbumAssetResponse.asset_id: str`. If one of those parents ends up in `stats.not_found_ids` (e.g., an asset in the current window that was deleted mid-sync and 404'd during asset phase), this branch now nulls that field via `model_copy(update=...)` which bypasses Pydantic validation. The downstream converter calls e.g. `safe_uuid_from_asset_id(face.asset_id)` which does `.startswith(...)` on `None` and raises `AttributeError`, aborting the stream generator.

This is pre-existing — Fix 4 introduced the null-out walk over all of `FK_REFERENCES`. Before this PR the checkpoint guard made it a full-sync-only problem; after this PR it applies in incremental sync too. In practice it's narrow (requires a parent deletion that races past its own fetch but whose cascaded `*_deleted` event doesn't land in the same window). Worth a follow-up to scope nulling to fields that are actually `Optional`, or to tag `FK_REFERENCES` entries with a `nullable` flag and log-warn instead of nulling when the field is required.

routers/api/sync/stream.py
```diff
@@ -184,6 +185,31 @@ async def _stream_entity_type(
...
+        # production. The payload override captures the causally-consistent
+        # value at event time, but the referenced entity may have been deleted
+        # since. A confirmed 404 adds the ID to stats.not_found_ids so
+        # null_deleted_fk_references can strip the reference before the event
+        # reaches the client — otherwise the client hits a SQLite FK violation
+        # (e.g., asset_face_entity → person) on insert. See the
+        # sync-stream-event-ordering design doc, "Fix 5".
+        payload_refs = extract_payload_fk_refs(gumnut_entity_type, events)
+        for ref_type, ref_ids in payload_refs.items():
+            unknown_ids = (
```
taggartgorman, line 199: **[Consider]** Verified-present refs aren't memoized across event batches

`unknown_ids` subtracts `streamed_ids` and `not_found_ids`, but IDs confirmed-present by the verification fetch aren't tracked anywhere. A cycle that spans multiple `EVENTS_PAGE_SIZE=200` batches where the same live person is referenced by face_updated events across batches will re-fetch that person once per batch rather than once per cycle.

For most sessions this never triggers (one batch, one call), matching the PR description's characterization. For large re-clustering runs the multiplier is mild but real. If you care to close the gap, add `stats.verified_present_ids: dict[str, set[str]]` populated from `ref_map.keys()` and include it in the subtraction. Otherwise the design-doc cost line is slightly understated — accurate phrasing is "one call per event batch per referenced type."

routers/api/sync/stream.py
```diff
@@ -184,6 +185,31 @@ async def _stream_entity_type(
...
+                - stats.not_found_ids.get(ref_type, set())
+            )
+            if not unknown_ids:
+                continue
+            ref_map, _ = await fetch_entities_map(
+                gumnut_client, ref_type, list(unknown_ids)
+            )
+            missing = unknown_ids - ref_map.keys()
+            if missing:
+                stats.not_found_ids[ref_type].update(missing)
```
taggartgorman, line 211: **[Consider]** Log when verification discovers a 404

The downstream `null_deleted_fk_references` already logs when it nulls a field, which covers most of the observability — but a log at the discovery site would let you distinguish "caught by Fix 5 verification" from "caught by entity fetch 404" when triaging future sync incidents. Edge case that wouldn't otherwise log: the face itself is 404'd (event is skipped upstream) so `null_deleted_fk_references` never runs, yet the verification correctly added the ref to `not_found_ids`.

```suggestion
            missing = unknown_ids - ref_map.keys()
            if missing:
                stats.not_found_ids[ref_type].update(missing)
                logger.info(
                    "Payload FK verification found deleted entities",
                    extra={
                        "entity_type": gumnut_entity_type,
                        "referenced_type": ref_type,
                        "missing_count": len(missing),
                        "missing_ids": sorted(missing),
                    },
                )
```

routers/api/sync/fk_integrity.py
```diff
@@ -18,6 +20,41 @@
...
+        ),
+    ],
+    "album": [
+        PayloadFKOverride(
+            event_type="album_updated",
+            payload_key="album_cover_asset_id",
+            referenced_type="asset",
+        ),
+    ],
+}
```
taggartgorman, line 61: **[Consider]** `FK_REFERENCES` and `PAYLOAD_FK_OVERRIDES` encode overlapping (entity, field, ref_type) triples

Both mappings carry `(face, person_id, person)` and `(album, album_cover_asset_id, asset)` — `FK_REFERENCES` so `null_deleted_fk_references` knows which field to null, `PAYLOAD_FK_OVERRIDES` so `extract_payload_fk_refs` knows which payload key to collect. They encode different concerns, so merging them isn't obviously right, but a future maintainer who renames a field or adds a new payload override must update both for the null-out flow to work end-to-end.

A short cross-reference comment here noting that every `(event_type, payload_key, referenced_type)` entry in `PAYLOAD_FK_OVERRIDES` assumes a matching `(field, referenced_type)` entry in `FK_REFERENCES` would make the coupling explicit. Alternatively, a small import-time assert or a unit test could enforce the invariant.

taggartgorman (approved):
## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (4 Consider-level comments inline below)
- [x] **Security/auth**: PR does not touch authentication, authorization, or permission logic. Diff limited to `routers/api/sync/fk_integrity.py`, `routers/api/sync/stream.py`, and docs/tests; the `AuthUser` sync-type handling in `generate_sync_stream` is untouched.
- [x] **Payments/billing**: no billing-related files in the diff.
- [x] **Data deletion**: `null_deleted_fk_references` nulls fields on outgoing sync events only; no DELETE statements or server-side row removals changed.
- [x] **Confidence**: Correctness is well understood. Verified `FaceResponse.asset_id` / `AlbumAssetResponse.album_id` / `AlbumAssetResponse.asset_id` nullability via Pydantic introspection; reasoning about the checkpoint-guard removal held up.
- [x] **Purpose clarity**: PR description and design doc Fix 5 section clearly explain the cross-cycle failure mode and the verification-based remedy.

## Summary

The Fix 5 design is sound: a confirmed 404 from production is strictly stronger evidence than any checkpoint-based assumption, so dropping the checkpoint-skip guard in `null_deleted_fk_references` is the right call. The `created_at_lt=sync_started_at` vs live-fetch tradeoff is explicitly documented and the end-state impact is cosmetic-only. Nice job on the regression tests — both cross-cycle scenarios are asserted at the behavioral level (nulled reference + `stats.not_found_ids` populated).

Four non-blocking Consider comments inline: one on a pre-existing narrow failure mode whose scope this PR expands (non-nullable FK nulling), and three small quality suggestions (memoization across batches, observability at the verification site, and a cross-reference between `FK_REFERENCES` and `PAYLOAD_FK_OVERRIDES`).

ternarybits:
## Round 2: addressing review feedback

Thanks for the approval and the four Consider notes, @taggartgorman. Summary of what this round did:

### Auto-fixed (two-way door)

- **Log 404 discovery at the Fix 5 verification site** — `routers/api/sync/stream.py` now emits a `logger.info "Payload FK verification found deleted entities"` with `{entity_type, referenced_type, missing_count, missing_ids}` whenever the bulk verification fetch returns a 404. This plugs the observability gap for the case where the face/album itself is 404'd (so `null_deleted_fk_references` never runs yet the ref is still correctly added to `not_found_ids`).
- **Cross-reference comment between `PAYLOAD_FK_OVERRIDES` and `FK_REFERENCES`** — `routers/api/sync/fk_integrity.py` now states the invariant above the overrides dict: every `(event_type, payload_key, referenced_type)` entry requires a matching `(field=payload_key, referenced_type)` entry in `FK_REFERENCES` for that entity type, otherwise `null_deleted_fk_references` won't null the payload-overridden field when its referenced entity 404s.
- **Tighten Fix 5 cost wording in the design doc** — `docs/design-docs/sync-stream-event-ordering.md` now describes the cost accurately: "one extra bulk list call per event batch per referenced type" with an explicit note that verified-present IDs are not memoized across batches, collapsing to the "one per cycle per type" characterization for typical single-batch sessions.

### Skipped (with rationale)

- **Nulling non-nullable FK fields after checkpoint-guard removal** — skipping in this PR per your own "worth a follow-up" framing. The fix isn't mechanical — deciding how to handle a non-nullable FK whose parent 404d is a genuine design choice (null-and-break downstream `.startswith`, skip the event entirely, or leave-ref-and-SQLite-FK-fail). Happy to file a follow-up Linear issue to track it; the scenario is narrow (parent deletion races past its own fetch but `*_deleted` event doesn't land in the same window) and is pre-existing from Fix 4.
- **Memoize verified-present refs across batches** — skipping. Per your own note, "for most sessions this never triggers (one batch, one call)." The updated cost line in the design doc now accurately reflects the per-batch-per-type cost so the doc is no longer understated; the mild multiplier on large re-clustering runs isn't worth the `stats.verified_present_ids` bookkeeping.

CI is green (lint, type, tests, compatibility).

routers/api/sync/stream.py
ternarybits: Fixed in d868124 — added `logger.info "Payload FK verification found deleted entities"` with `entity_type`, `referenced_type`, `missing_count`, and sorted `missing_ids` at the verification site. Also covers the case where the face/album itself is 404d and `null_deleted_fk_references` never runs.

routers/api/sync/fk_integrity.py
ternarybits: Fixed in d868124 — added an invariant comment above `PAYLOAD_FK_OVERRIDES` stating that every `(event_type, payload_key, referenced_type)` entry requires a matching `(field=payload_key, referenced_type)` entry in `FK_REFERENCES` for the same entity type, so `null_deleted_fk_references` will actually null the payload-overridden field when its referenced entity 404s.

routers/api/sync/fk_integrity.py
ternarybits: Skipping in this PR — agreeing with your own framing as a follow-up. The fix isn't mechanical: choosing how to handle a non-nullable FK whose parent 404d is a genuine design decision (null-and-break downstream `.startswith`, skip the event entirely, or leave-ref-and-SQLite-FK-fail). The scenario is narrow (requires a parent deletion racing past its own fetch whose cascaded `*_deleted` event doesn't land in the same window) and is pre-existing from Fix 4. Will file a Linear follow-up.

routers/api/sync/stream.py
ternarybits: Skipping memoization — your own note ("for most sessions this never triggers, one batch one call") matches the typical shape. Instead I tightened the design-doc cost line in d868124 to accurately say "one extra bulk list call per event batch per referenced type" with an explicit note that verified-present IDs are not memoized across batches, so the doc is no longer understated. If large re-clustering runs become a real hotspot we can revisit with `stats.verified_present_ids`.

PR #133, opened 2026-04-20
Add design doc: adapter-side trash work (GUM-556)

taggartgorman:
## Summary
- Companion adapter-side design doc to the backend trash design in `photos/docs/design-docs/trash-soft-delete-retention.md` (GUM-556)
- Covers delete `force` mapping, `/trash/*` endpoint implementations, `isTrashed` / `deletedAt` DTO plumbing, sync-stream hydration via the backend's privileged fetch, `on_asset_trash` / `on_asset_restore` WebSocket events, and `trashDays` server config
- Registers the new doc in `AGENTS.md` under Design Docs

Linear: https://linear.app/gumnut-ai/issue/GUM-556/implement-trash-soft-delete-with-retention

## Test plan
- [ ] Doc renders correctly on GitHub
- [ ] File paths and line numbers referenced in the doc still match the current `main` checkout
- [ ] Cross-reference to backend doc resolves after review

### DIFF
--- a/docs/design-docs/trash-soft-delete-adapter.md
+++ b/docs/design-docs/trash-soft-delete-adapter.md
@@ -48,8 +48,8 @@ Non-goals:
 
 Split the flow in `delete_assets` (`routers/api/assets.py:609`):
 
-- `force=False` (Immich default): call the backend's `trash_assets` endpoint with the full id list in a single request. Backend returns the ids that actually transitioned (live → trashed). For each transitioned id, emit a WebSocket `on_asset_trash` event; see *WebSocket events* below.
-- `force=True`: call the backend's `purge_assets` endpoint with the full id list. Backend returns the ids that were actually purged. Emit `on_asset_delete` per purged id.
+- `force=False` (Immich default): call the backend's `trash_assets` endpoint with the full id list in a single request. Backend returns the ids that actually transitioned (live → trashed). Emit a single batched `on_asset_trash` WebSocket event with the full transitioned-id array; see *WebSocket events* below for the array-payload contract.
+- `force=True`: call the backend's `purge_assets` endpoint with the full id list. Backend returns the ids that were actually purged. Emit one `on_asset_delete` event per purged id (Immich's wire shape is single-id per event for permanent deletes).
 
 Both paths replace the current per-id loop: today's code issues one `client.assets.delete(...)` per id and swallows 404s individually. The new paths should issue one bulk call per request and preserve the "return 204 and keep going" semantics by letting the backend skip ids it cannot transition (it already no-ops on already-trashed / already-deleted, per the backend design's `UPDATE ... WHERE deleted_at IS NULL RETURNING id` contract). Partial failures at the network layer still map through `map_gumnut_error` and surface as a single adapter 5xx to the client.
 
@@ -59,9 +59,9 @@ The exact REST shape on the backend side — `POST /api/assets/trash` vs. `DELET
 
 `routers/api/trash.py` is replaced end to end:
 
-- `POST /api/trash/restore/assets` → call backend `restore_assets` with the DTO id list. Emit `on_asset_restore` for every id the backend reports restored. Return `TrashResponseDto(count=<restored>)`.
-- `POST /api/trash/restore` → restore **all** of the caller's trashed assets. Immich's native `TrashService.restore(auth)` takes no id list (see `immich/server/src/services/trash.service.ts:27`). The backend exposes `list_trashed_assets()` per-library; the adapter enumerates the caller's trashed ids (paginated) and calls `restore_assets` with them. Return the total restored count, and emit `on_asset_restore` in batches so the event payload stays bounded.
-- `POST /api/trash/empty` → purge **all** of the caller's trashed assets. Same enumeration pattern: list trashed ids, then call the backend's `purge_assets`. Return the count purged. Emit `on_asset_delete` per purged id.
+- `POST /api/trash/restore/assets` → call backend `restore_assets` with the DTO id list. Emit a single batched `on_asset_restore` WebSocket event carrying the full restored-id array. Return `TrashResponseDto(count=<restored>)`.
+- `POST /api/trash/restore` → restore **all** of the caller's trashed assets. Immich's native `TrashService.restore(auth)` takes no id list (see `immich/server/src/services/trash.service.ts:27`). The backend exposes `list_trashed_assets()` per-library; the adapter enumerates the caller's trashed ids (paginated) and calls `restore_assets` with them. Return the total restored count. Emit one batched `on_asset_restore` event per enumeration page (the restored-id array is bounded by the page size) rather than one per id or one mega-event across all pages.
+- `POST /api/trash/empty` → purge **all** of the caller's trashed assets. Same enumeration pattern: list trashed ids, then call the backend's `purge_assets`. Return the count purged. Emit one `on_asset_delete` event per purged id (Immich's wire shape is single-id per event for permanent deletes).
 
 Immich's native `POST /trash/empty` returns immediately and queues the deletions to a background job (`TrashService.empty` → `AssetEmptyTrash` queue). The backend's `purge_assets` is also async-safe: it commits the DELETE + outbox together, and storage/CDN cleanup runs on `on_commit` via `AssetStorageCleanupTask`. So the adapter can call `purge_assets` synchronously and return the count; the user-visible behavior (assets disappear from the trash view immediately, storage frees shortly after) is indistinguishable from Immich's native behavior.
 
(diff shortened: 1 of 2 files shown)


taggartgorman:
@CharlieHelps review please

taggartgorman:
## Review feedback addressed (round 1)

Both items from @charliehelps are auto-fixed (two-way door doc edits). No escalations.

### Auto-fixed

1. **[Blocking] Rollout sequencing for PR A vs backend PR 1** — rewrote the Migration / Rollout Plan to explicitly gate all three adapter PRs (A/B/C) on backend PR 1 deploy + Gumnut SDK regeneration. Removed the "safe to land early" claim on PR A. Kept the orthogonal distinction that PR A can land before or after backend PR 2 (default-delete cutover) — the privileged-fetch switch and DTO passthrough are correct no-ops while `deleted_at` is null everywhere, which is the pre-PR-2 regime.
2. **[Non-blocking] WebSocket emission wording** — tightened the delete/trash section bullets so "emit a single batched `on_asset_trash` event with the full transitioned-id array" matches the later WebSocket contract section. Restore endpoints follow the same pattern. Permanent-delete stays per-id (matching Immich's `on_asset_delete: [string]` shape). For the `/trash/restore` enumeration flow, clarified that a batched event fires per enumeration page — bounded by page size — rather than per-id or one mega-event.

### Skipped

None.

Commit: 5e05d0b

docs/design-docs/trash-soft-delete-adapter.md
```diff
@@ -0,0 +1,247 @@
...
+
+Suggested adapter PR order (single PR is fine if reviewers prefer; these split out if the diff gets large):
+
+**PR A — DTO conversions + sync stream hydration**
+
+1. Add `deleted_at` passthrough in `convert_gumnut_asset_to_immich`, `gumnut_asset_to_sync_asset_v1`, and `build_asset_upload_ready_payload`.
+2. Switch `fetch_entities_map`'s `asset` branch to the privileged (include-trashed) fetch for sync-stream callers. Keep the default `assets.list` for every other caller (`timeline.py`, `albums.py`, `people.py`, `assets.py`, `search.py`).
+3. Update `trashDays` in server config to read from env.
+
+Before backend PR 1 ships, `deleted_at` is always null on the wire, so all three changes are no-ops in production; safe to land early.
```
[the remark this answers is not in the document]
taggartgorman: Fixed in 5e05d0b. The Rollout section now explicitly gates all three adapter PRs on backend PR 1 + SDK regeneration and removes the safe-to-land-early claim for PR A. PR A can still land before or after backend PR 2 (default-delete cutover) — the privileged-fetch switch and DTO passthrough are correct no-ops in the pre-PR-2 regime.

docs/design-docs/trash-soft-delete-adapter.md
```diff
@@ -0,0 +1,245 @@
...
+status: draft
+created: 2026-04-20
+last-updated: 2026-04-20
+---
+
+# Trash: Soft-Delete with Retention (Adapter)
+
+## Context
+
+The backend half of this work is designed in `photos/docs/design-docs/trash-soft-delete-retention.md` (GUM-556) and already merged as a design. It adds a nullable `deleted_at` column on `assets`, splits `AssetService.delete_assets()` into `trash_assets` / `restore_assets` / `purge_assets`, emits new `ASSET_TRASHED` / `ASSET_RESTORED` events, filters trashed rows from every user-facing query, and schedules a retention-driven purge task. The backend also exposes a privileged ids-only fetch that includes trashed rows (for sync hydration) and exposes `deleted_at` on `AssetResponse`.
```
ternarybits: Immich-adapter should never refer to documents in the closed source Gumnut repos.  It can talk about the interface and guarantees that it's relying on, but not point to specific document paths.

docs/design-docs/trash-soft-delete-adapter.md
```diff
@@ -0,0 +1,245 @@
...
+
+This doc covers the adapter half. Today:
+
+- `DELETE /api/assets` (`routers/api/assets.py:609`) hard-deletes via `client.assets.delete`. The `force` parameter on `AssetBulkDeleteDto` is silently ignored (see `assets.py:617`).
+- `POST /api/trash/empty`, `POST /api/trash/restore`, and `POST /api/trash/restore/assets` (`routers/api/trash.py`) all return `TrashResponseDto(count=0)` as stubs.
+- `GET /api/timeline/buckets` and `GET /api/timeline/bucket` (`routers/api/timeline.py:79-84`, `timeline.py:255`) return an empty list whenever `isTrashed=true` and hardcode `isTrashed=False` on every bucket entry — so the web trash page is permanently empty.
+- `GET /api/assets/statistics` (`routers/api/assets.py:693`) accepts an `isTrashed` query param but ignores it.
+- `POST /api/search/metadata` and `POST /api/search/large-assets` accept `withDeleted` / `trashedBefore` / `trashedAfter` and ignore them.
```
ternarybits, line 28: This implies that we need the ability to return both delete and non-deleted assets in the same response.  I don't think our database indexes in your design doc support that well.  They're all partial indexes on deleted_at.

docs/design-docs/trash-soft-delete-adapter.md
```diff
@@ -0,0 +1,245 @@
...
+- `AssetResponseDto.isTrashed` is hardcoded to `false` in `convert_gumnut_asset_to_immich` (`routers/utils/asset_conversion.py:357`).
+- `SyncAssetV1.deletedAt` is hardcoded to `None` in both `gumnut_asset_to_sync_asset_v1` (`routers/api/sync/converters.py:162`) and `build_asset_upload_ready_payload` (`routers/utils/asset_conversion.py:266`).
+- The adapter emits `WebSocketEvent.ASSET_DELETE` (`on_asset_delete`) on every delete (`assets.py:630-643`), regardless of `force`. The Immich UI uses `on_asset_trash` for soft-deletes and `on_asset_delete` only for permanent deletes; conflating the two causes the web client to remove the asset from the timeline immediately even on a soft delete, where a restore should have been possible.
+
+The Immich wire contract is public and consumed by unmodified web and mobile clients — we cannot change endpoint shapes. Internal plumbing is ours.
+
+## Goals
+
+- `DELETE /api/assets` honors `force`: `force=true` permanently deletes, `force=false` (the default) soft-deletes via the backend's new `trash_assets` service.
```
ternarybits: I'm thinking about the photos-api side of this API some more.  Let's chat about this in office.

Notes:
* I want to make it hard for an agent to accidentally force-delete assets.  
* The MCP tool for moving to trash should be different than the MCP tool for force deleting or emptying the trash.  I should be able to "allow" the "soft delete" tool without also allowing the "force delete"
* Should the REST API for moving an asset to the trash be DELETE?  That doesn't follow normal resource semantics.

docs/design-docs/trash-soft-delete-adapter.md
taggartgorman: Good call — done in d71d01e. Rewrote the context and dependencies sections to describe the backend interface and guarantees (nullable deleted_at, trash/restore/purge service methods, new ASSET_TRASHED/RESTORED events, privileged ids-only fetch, retention-driven purge) without naming private-repo paths. Also stripped the Linear ID.

docs/design-docs/trash-soft-delete-adapter.md
taggartgorman: You're right that `withDeleted=true` semantically asks for both in one response — that is the case the partial indexes don't serve directly. I dug through the access patterns we actually have:

- **Timeline/list endpoints** never ask for both — live timeline and trash page are separate endpoints on the Immich wire, so each query hits a single-scope index.
- **Checksum dedup on upload** does need to see both; it uses the existing non-partial `uq_assets_library_checksum` unique constraint, which spans both.
- **The privileged ids-only fetch** for sync hydration is a PK lookup — index-irrelevant.
- **`withDeleted=true` on search** is the one combined-view case. Search is ordered by vector similarity, not by date, so including trashed rows is a filter change on the query — the partial date indexes aren't what drive search plans. Also: the adapter defers `withDeleted=true` today because no Immich UI passes it, so this is forward-looking.

So the partial-predicate design holds for our current access patterns, but you're right that the backend doc doesn't say any of this. I've opened [GUM-592](https://linear.app/gumnut-ai/issue/GUM-592) to update the backend design doc to walk through each combined-view query and what serves it, so this reasoning is visible instead of having to be re-derived.

— drafted by Claude (Opus 4.7), posted by @taggartgorman

docs/design-docs/trash-soft-delete-adapter.md
taggartgorman: GUM-592 has been opened to handle updates to the backend design, including a "safer" surface regarding deletions.

ternarybits (approved):

PR #138, opened 2026-04-24
Migrate sync stream from asset.exif to asset.metadata

taggartgorman:
## Summary

Switches the adapter from the legacy `asset.exif` field and `exif_*` events on the photos-api to the new `asset.metadata` field and `metadata_updated` events. Both surfaces are populated upstream during the dual-write/dual-emission migration window, so this is a clean cutover with `git revert` as the rollback.

- **Read path:** every site that read `gumnut_asset.exif` now reads `gumnut_asset.metadata` — `routers/utils/asset_conversion.py` (4 functions) and the sync-stream converter, which is renamed `gumnut_metadata_to_sync_exif_v1`. The converter hard-codes `profileDescription=None` because the new `Metadata` type intentionally drops that column upstream; the WebSocket-payload `extract_sync_exif` path already emitted `None` here, so client-visible behavior is unchanged.
- **Event subscriber:** `_SYNC_TYPE_ORDER`, `entity_fetch`, `events` dispatch, and `fk_integrity` swap their `"exif"` keys/branches to `"metadata"` so the events API filter selects `metadata_updated` records. Drops `"exif_deleted"` from `_SKIPPED_EVENT_TYPES` — filtering by `metadata` no longer surfaces exif-category events at all, so the explicit skip is dead code.
- **Immich-facing types unchanged:** `SyncAssetExifV1`, `SyncEntityType.AssetExifV1`, `SyncRequestType.AssetExifsV1`, `AssetUploadReadyV1Payload.exif`, and `ExifResponseDto` remain mobile/web client contracts.
- **SDK bump:** `gumnut-sdk` 0.66 → 0.74 to pick up the new `Metadata` type on `AssetResponse`.
- **Docs + tests** migrated to match (FK-order diagrams, mock fixtures, integration fixtures). One extra fixture-pattern note added to `docs/references/code-practices.md` after several `Mock`-based tests silently flipped branches when the renamed attribute defaulted to a truthy `Mock` instead of `None`.

## Linear

https://linear.app/gumnut-ai/issue/GUM-479/phase-1-asset-metadata-layers-asset_metadata-table-rename

## Test plan

- [x] `uv run pytest` — 675 passed
- [x] `uv run ruff check` / `ruff format` / `pyright` — clean
- [ ] Reviewer / post-merge: confirm a fresh sync against staging streams `AssetExifV1` events backed by `metadata_updated` cursors and that existing client checkpoints resume cleanly (cursors are global across event streams upstream, so prior `exif_*` cursors should still advance the stream)


taggartgorman:
@CharlieHelps Review this PR

ternarybits (approved):

PR #144, opened 2026-04-29
Use bulk SDK calls for single-album add/remove (GUM-650)

ternarybits:
Resolves [GUM-650](https://linear.app/gumnut-ai/issue/GUM-650). Replaces #143 with a clean rewrite from main.

## Summary

The single-album endpoints (`PUT`/`DELETE /api/albums/{id}/assets`) were fanning out one SDK call per asset. The SDK and upstream `/api/albums/{album_id}/assets` already accept a list of `asset_ids` and split added/duplicate server-side, so a single bulk call replaces the per-asset loop.

## Changes

- **`add_assets_to_album`**: one bulk SDK call. Map per-asset response from the `{added_assets, duplicate_assets}` body. On any bulk error (including a 404 from a missing/wrong-library asset), mark every requested asset with the classified error — the upstream validates before any DB write, so a 404 means the whole batch failed and there is nothing to recover per-item.
- **`remove_asset_from_album`**: one bulk SDK call. The upstream silently skips missing IDs and returns 204, so per-asset `not_found` errors aren't reachable; batch-level errors map to every entry.
- **Multi-album `add_assets_to_albums`**: unchanged (still sequential per-album fan-out, same as pre-PR).

## Latent bug fix (side effect)

The prior per-item code caught `ConflictError` to map duplicates to `error=duplicate`, but the upstream returns 200 with `duplicate_assets` populated (not 409), so the catch never fired and duplicates were silently reported as `success=True`. Reading the response body fixes this.

## Follow-ups

- [GUM-667](https://linear.app/gumnut-ai/issue/GUM-667) — make upstream `add_assets_to_album` silently skip missing assets and return them in a `not_found_assets` response field. Once that lands, the "mark all as not_found on bulk error" branch in this PR can be replaced with structured per-asset classification.
- [GUM-668](https://linear.app/gumnut-ai/issue/GUM-668) — parallelize the remaining per-item fan-out endpoints in the adapter (`delete_assets`, `update_people`, `delete_people`, `reassign_faces`, multi-album `add_assets_to_albums`) with a shared bounded-concurrency utility.
- [GUM-669](https://linear.app/gumnut-ai/issue/GUM-669) — cap `asset_ids` length on the upstream album add/remove bulk endpoints to avoid hitting the Postgres bind-param limit (32,767) at extreme batch sizes.
- [GUM-670](https://linear.app/gumnut-ai/issue/GUM-670) — chunk `asset_ids` in the adapter before calling the SDK so a power user adding ≥30k photos in one click doesn't trigger a 5xx. Blocked on GUM-669.

## Tests

- 36 album tests pass (`uv run pytest tests/unit/api/test_albums.py`)
- New tests cover: single-call assertion, response-body duplicate detection, 404 marks all as `not_found`, other-status-error fan-out (4xx/5xx), transport-error fan-out for both add/remove
- Removed: dead `ConflictError`-based duplicate test for `add_assets_to_album`

## Validation

- `uv run ruff format` / `ruff check` (clean)
- `uv run pyright` (clean)
- `uv run pytest` (all pass)

### DIFF
--- a/docs/architecture/adapter-architecture.md
+++ b/docs/architecture/adapter-architecture.md
@@ -317,7 +317,12 @@ Immich clients have no HTTP 429 handling — a rate limit response causes sync f
 
 ### Per-item error handling
 
-Bulk operations (delete assets, update people, add assets to albums) process items individually and track per-item results. A failure on one item doesn't abort the entire operation — the adapter continues processing remaining items and returns a result array with success/error status per item. The shared `classify_bulk_item_error()` helper in `routers/utils/error_mapping.py` maps `APIStatusError` subclasses to the canonical `not_found` / `no_permission` / `unknown` buckets; per-endpoint nuances (e.g. `ConflictError` → `duplicate`) are layered on top.
+Bulk operations expose per-item results to Immich clients via `BulkIdResponseDto[]`, but use one of two implementation shapes depending on whether the upstream Gumnut endpoint is itself bulk-aware:
+
+1. **Single bulk SDK call** (e.g. single-album `add_assets_to_album` / `remove_asset_from_album`). The upstream accepts a list and either splits the result server-side (`add` returns `{added_assets, duplicate_assets}`) or silently skips missing IDs (`remove` returns 204). The adapter maps the response body into per-item results on success; on a bulk error, the whole batch failed (the upstream validates before any DB write), so every requested asset is annotated with the same classified error.
+2. **Per-item fan-out** (e.g. `delete_assets`, `update_people`, multi-album `add_assets_to_albums`). The upstream operates one entity at a time; the adapter loops and tracks per-item results, continuing past per-item failures.
+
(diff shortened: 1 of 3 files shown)


taggartgorman (approved):
Approving via auto-approve fast path (zero Must Fix / zero Should Fix findings).

## Pre-approve checklist
- [x] **Security/auth**: Untouched. Endpoints retain `get_authenticated_gumnut_client` dep; no change to auth or permission logic.
- [x] **Payments/billing**: Untouched. Scope is `routers/api/albums.py` + its tests.
- [x] **Data deletion**: `remove_asset_from_album` only detaches associations (verified against the SDK docstring: "The assets remain in the library and in any other albums they belong to."). Reversible.
- [x] **Confidence**: High. SDK semantics (`added_assets`/`duplicate_assets` body, 404-on-missing for add, silent-skip for remove) are clearly documented in the SDK source, and the new tests pin every branch the endpoint logic walks (success, duplicate-from-response, 404 marks all, other 4xx/5xx, transport error).
- [x] **Purpose clarity**: PR description spells out the bulk migration, the latent duplicate-handling bug, the all-or-nothing trade-off, and the tracked follow-ups.

## Notes

- Nice catch on the latent duplicate-handling bug — the previous `except ConflictError` never fired because upstream returns 200 with `duplicate_assets` populated, so duplicates were silently reported as `success=True`. Reading the response body fixes it.
- The all-or-nothing batch failure on `add` (one bad asset 404s the whole batch) is a real behavior change vs. the previous partial-success per-item loop, but it's a faithful reflection of upstream's atomic validation and is acknowledged + tracked via the GUM-667 follow-up.
- The defensive `else` branch in `add_assets_to_album` (asset missing from both `added` and `duplicate`) is appropriately fail-loud (warn + `unknown` rather than silently succeeding).

PR #145, opened 2026-04-29
GUM-634: Plumb trashed_at through DTOs + sync hydration; trashDays from env

taggartgorman:
## Summary

Adapter PR A for the trash soft-delete feature — three independently-revertible changes that surface trash state to Immich clients.

- **DTO conversions**: `convert_gumnut_asset_to_immich`, `build_asset_upload_ready_payload`, and `gumnut_asset_to_sync_asset_v1` now read `trashed_at` from `AssetResponse` instead of hardcoding `False`/`None`. This populates `AssetResponseDto.isTrashed` (REST), `SyncAssetV1.deletedAt` (sync stream), and `AssetUploadReadyV1Payload.asset.deletedAt` (upload websocket).
- **Sync stream hydration**: `fetch_entities_map`'s asset branch now fetches with `state="all"` so `ASSET_TRASHED` events hydrate successfully — without this they're filtered from `page.data` and dropped before reaching the client. The same call site also handles album-cover payload-FK verification, so trashed cover assets are no longer nulled out by FK reconciliation. Metadata branch deliberately stays on the live-only default.
- **Server config**: `/api/server/config.trashDays` now reads from a new `TRASH_RETENTION_DAYS` env var via `get_settings()` (default 90, matching photos-api). Replaces the hardcoded `30` that misled the web trash page's "permanently deleted after N days" message.

Bumps `gumnut-sdk` to 0.77.0 for the new `trashed_at` field on `AssetResponse` and the `state` query parameter on `assets.list`.

## Linear

https://linear.app/gumnut-ai/issue/GUM-634/trash-pr-a-adapter-dto-conversions-sync-stream-hydration

Depends on backend PR (now deployed) that exposes `trashed_at` and `state=all`. Independently revertible — rolling back returns null-everywhere DTOs and the default sync hydration filter; trash events drop silently on hydration until this PR re-lands, but no persistent state is corrupted.

## Test plan

- [x] `uv run pytest` passes (731 tests, including 10 new)
- [x] `uv run ruff check`, `uv run ruff format`, `uv run pyright` clean
- [x] Manual: trigger an `asset_trashed` event, verify the sync stream emits a `SyncAssetV1` with non-null `deletedAt` (regression test exists; this is a smoke test against a real backend)
- [x] Manual: hit `GET /api/server/config` with `TRASH_RETENTION_DAYS=45` set in env — verify `trashDays: 45`
- [x] Manual: re-upload a file whose checksum matches a trashed asset; verify `isTrashed: true` on the response

## Notes

Test fixtures across `tests/conftest.py`, `tests/unit/api/sync/conftest.py`, `tests/unit/api/sync/test_datetime_conversion.py`, `tests/unit/api/test_assets.py`, and `tests/unit/api/test_search.py` were audited to explicitly set `trashed_at = None` on Mock-based asset constructions — unset Mock attributes return truthy child Mocks, which would silently flip Pydantic validation in any test exercising `convert_gumnut_asset_to_immich` or `build_asset_upload_ready_payload`.


taggartgorman:
@CharlieHelps review

routers/api/server.py
```diff
@@ -55,20 +55,29 @@
...
-    "isInitialized": True,
-    "isOnboarded": True,
-    "externalDomain": "",
-    "maintenanceMode": False,
-    "publicUsers": True,
-    "mapDarkStyleUrl": "https://tiles.immich.cloud/v1/style/dark.json",
-    "mapLightStyleUrl": "https://tiles.immich.cloud/v1/style/light.json",
-}
+
+def get_fake_config() -> dict:
```
ternarybits, line 59: Seems like we could replace with this with real config.  Create a linear ticket for this?  Or just do it now.

ternarybits (approved):

ternarybits (approved):
## Pre-approve checklist

- [x] **Findings**: 0 Must Fix / 0 Should Fix (1 Consider below)
- [x] **Security/auth**: PR does not touch authentication, authorization, or permission logic. The diff in `routers/api/server.py` only modifies the `fake_config` dict (renamed to `get_fake_config()`) and the `/config` endpoint; oauth/license/auth handlers in the same file are untouched. Other touched files have no auth surface.
- [x] **Payments/billing**: PR does not touch payments, billing, or subscription logic.
- [x] **Data deletion**: PR is read-side only — surfaces `trashed_at` on DTOs, switches sync hydration to `state="all"` (a read filter), adds `trashDays` config field. No delete/purge/trash backend calls — those land in PR B per the design doc.
- [x] **Confidence**: 17 files, 280/26 lines (mostly tests). Three DTO sites changed identically (read `trashed_at` instead of hardcoded null/false). One-line sync-fetch change with code comment matching the design doc. Settings change is a standard pydantic-settings field. 10 new tests cover the new branches; the existing fixture audit explicitly handles the Mock-truthiness pitfall called out in `code-practices.md`.
- [x] **Purpose clarity**: Approach matches PR A scope in `docs/design-docs/trash-soft-delete-adapter.md`.

---

**Why this is safe to land**: All three DTO conversions read the same nullable `trashed_at` and surface it identically (`bool(trashed_at)` → `isTrashed`, raw value → `deletedAt`). The sync-fetch widening to `state="all"` is scoped to `fetch_entities_map` (sync-stream + payload-FK verification only); the metadata branch deliberately stays on the default live-only filter, which is consistent with the design doc and the inline code comment. The `get_fake_config()` rename moves the dict construction from module-import time to request time so `lru_cache`d settings drive `trashDays`.

**Test fixture audit verified**: spot-checked `test_timeline.py:466,498` and `test_faces.py:35` — these Mock-based asset constructions don't flow through the changed DTO conversion sites, so they correctly don't need a `trashed_at = None` line. The audit covers every fixture that does.

**Independent revertability confirmed**: rolling this back returns the null-everywhere DTOs and live-only sync hydration described in the design doc; trash events would drop silently on hydration until re-landed, with no persistent state corruption.

## [Consider] Update design doc to reflect shipped SDK shape

`docs/design-docs/trash-soft-delete-adapter.md` (status: `draft`, last-updated 2026-04-20) references the pre-implementation SDK names — `deleted_at` field on `AssetResponse` and a separate `list_assets_including_trashed` method — that diverge from what shipped in gumnut-sdk 0.77.0 (`trashed_at` field; `state="all"` parameter on `assets.list`). Status is `draft` so this isn't strictly flag-worthy, but a small evolution note before PR B/C land would save the next reader from translating the API names mentally.

Non-blocking; feel free to handle in PR B's doc-touch or as a follow-up.

routers/api/server.py
taggartgorman: Filed as [GUM-684](https://linear.app/gumnut-ai/issue/GUM-684/replace-remaining-hardcoded-values-in-get-fake-config-with). Most of the remaining fields are either operator-tunable (worth env vars) or irrelevant (e.g. map URLs are dead given `map: False`); ticket has the field-by-field audit.

PR #146, opened 2026-04-30
GUM-635: Trash PR B: DELETE honors force + trash router + WebSocket events

taggartgorman:
## Summary

- `DELETE /api/assets` now honors `force`: `force=True` permanent-deletes via the backend's bulk `DELETE /api/assets`; `force=False` (or absent — Immich's native default) soft-deletes via `POST /api/assets/trash`. Replaces the per-id loop that silently ignored `force`.
- `routers/api/trash.py` rewritten end-to-end: `/restore/assets` bulk-restores by ids; `/restore` enumerates `state="trashed"` and chunked-restores all of the caller's trashed assets; `/empty` enumerates and chunked bulk-deletes. All three return real counts.
- `WebSocketEvent` gains `ASSET_TRASH = "on_asset_trash"` and `ASSET_RESTORE = "on_asset_restore"` to match Immich's wire contract — single batched event per chunk for soft transitions, per-id `on_asset_delete` for permanent deletes.

## Linear

https://linear.app/gumnut-ai/issue/GUM-635/trash-pr-b-adapter-delete-honors-force-trash-router-websocket-events

## Implementation notes

- Backend bulk endpoints (`POST /api/assets/{trash,restore}`, bulk `DELETE /api/assets`) accept `{"ids": [...]}` body and are capped at `MAX_BULK_GET_IDS = 100`. Adapter chunks via `itertools.batched(..., BULK_CHUNK_SIZE)` with `BULK_CHUNK_SIZE = 100` hoisted to `routers/utils/gumnut_client.py` for reuse.
- The Gumnut SDK does not yet expose typed methods for these endpoints, so calls go through the public raw HTTP layer (`AsyncGumnut.post()` / `.delete()` with `cast_to=type(None)`). This routes through the same JWT auth, retry, and response-hook plumbing as typed SDK methods. New code-practices doc section captures the pattern for future endpoints landing before SDK regen.
- Backend bulk endpoints are idempotent on already-transitioned rows, so the previous per-id 404 swallowing is intentionally removed; failures propagate via the global `GumnutError` handler.
- `/api/trash/empty` enumerates and chunked-deletes (rather than calling `POST /api/assets/empty-trash` once) to satisfy the design doc's "WebSocket events fire on every transition" requirement and to return an accurate count without a separate `counts` round-trip.
- The `server_features.trash` flag stays `False` in this PR — the trash page can't render until PR C (GUM-636) fixes the `GET /api/timeline/buckets?isTrashed=true` short-circuit. Flag flip will ship with PR C.

## Test plan

- [x] Unit tests pass: `uv run pytest` (747 passing)
- [x] Type check clean: `uv run pyright` on PR files (0 errors)
- [x] Lint/format clean: `uv run ruff check && uv run ruff format`
- [x] **Test 1 — `force=false` soft-delete (browser, `server_features.trash` temporarily flipped to True locally)**: selected assets in Immich web UI → clicked Delete → toast shown, assets removed from timeline. Adapter logs: single `POST /api/assets/trash → 204` (not per-id) and `emitting event "on_asset_trash"` to user room.
- [x] **Test 2 — restore by ids (curl)**: `DELETE /api/assets force=false` (2 ids) → `POST /api/trash/restore/assets` → `count=2`; both assets verified `isTrashed=false`, `deletedAt=null`.
- [x] **Test 3 — restore-all (curl)**: `POST /api/trash/restore` → `count=2`; backend `POST /api/assets/restore` and `on_asset_restore` event emitted.
- [x] **Test 4 — `force=true` permanent delete (curl)**: `DELETE /api/assets force=true` (1 id) → 204; subsequent `GET /api/assets/<id>` → 404 (asset purged, no trash hop).
- [x] **Test 5 — empty trash (curl)**: trashed 2 ids → `POST /api/trash/empty` → `count=2`; both ids → 404. Backend received bulk `DELETE /api/assets` after enumeration.
- [ ] Manual (deferred to PR C): trash page renders trashed assets with Restore affordance once `GET /api/timeline/buckets?isTrashed=true` is wired.
- [ ] Manual (deferred): Immich mobile delete lands asset in app's local trash collection via the sync-stream `SyncAssetV1` upsert (PR A's `deletedAt` plumbing).
- [ ] Manual (deferred, optional): WebSocket frame inspection in Firefox devtools confirming `on_asset_trash: [string[]]` payload shape vs `on_asset_delete: [string]`.

### DIFF
--- a/docs/architecture/websocket-implementation.md
+++ b/docs/architecture/websocket-implementation.md
@@ -159,7 +159,9 @@ Starting with `on_upload_success`, but designed for future extension:
 |---|---|---|---|---|
 | `on_upload_success` | `AssetResponseDto` | Yes | Legacy | photos-api thumbnails are synchronous |
 | `AssetUploadReadyV1` | `SyncAssetV1` + `SyncAssetExifV1` | No | v2 sync | Emit alongside `on_upload_success` |
-| `on_asset_delete` | `string` (assetId) | Yes | Yes | photos-api deletion is synchronous |
+| `on_asset_delete` | `string` (assetId) | Yes | Yes | One per id; force=true permanent delete |
+| `on_asset_trash` | `string[]` (assetIds) | Yes | Yes | Batched per chunk; force=false soft delete |
+| `on_asset_restore` | `string[]` (assetIds) | Yes | Yes | Batched per chunk; restore from trash |
 | `on_session_delete` | `string` (sessionId) | Yes | No | Sessions managed by immich-adapter |
 | `on_server_version` | `ServerVersionResponseDto` | Yes | No | Sent on connect (existing) |
 
@@ -175,7 +177,6 @@ These events require features that don't exist in photos-api:
 
 | Event | Reason |
 |---|---|
-| `on_asset_trash` / `on_asset_restore` | photos-api hard-deletes (no soft-delete) |
 | `on_asset_update` | Assets are immutable after creation |
 | `on_asset_stack_update` | Stacks not implemented |
 | `on_asset_hidden` | Asset visibility not supported |
@@ -186,7 +187,7 @@ These events require features that don't exist in photos-api:
 ### Implementation Priority
 
 1. **`on_upload_success` + `AssetUploadReadyV1`** - Core upload feedback for web and mobile
-2. **`on_asset_delete`** - Timeline synchronization
+2. **`on_asset_delete` + `on_asset_trash` + `on_asset_restore`** - Timeline synchronization for hard-delete, soft-delete, and restore flows
 3. **`on_session_delete`** - Security critical; forces web logout when session deleted
 
 ### 3.3 Mobile v2 Sync Protocol
@@ -218,6 +219,8 @@ class WebSocketEvent(Enum):
     UPLOAD_SUCCESS = "on_upload_success"
     ASSET_UPLOAD_READY_V1 = "AssetUploadReadyV1"
     ASSET_DELETE = "on_asset_delete"
+    ASSET_TRASH = "on_asset_trash"
+    ASSET_RESTORE = "on_asset_restore"
     SESSION_DELETE = "on_session_delete"
     SERVER_VERSION = "on_server_version"
 
@@ -298,6 +301,8 @@ class WebSocketEvent(Enum):
     UPLOAD_SUCCESS = "on_upload_success"
     ASSET_UPLOAD_READY_V1 = "AssetUploadReadyV1"
     ASSET_DELETE = "on_asset_delete"
+    ASSET_TRASH = "on_asset_trash"
+    ASSET_RESTORE = "on_asset_restore"
     SESSION_DELETE = "on_session_delete"
     SERVER_VERSION = "on_server_version"
 
@@ -416,14 +421,14 @@ async def emit_event(event: WebSocketEvent, user_id: str, payload: EventPayload 
 ```python
 # In routers/api/assets.py (after upload completes)
 
-from routers.api.websockets import emit_event, WebSocketEvent
+from services.websockets import emit_user_event, WebSocketEvent
 
 @router.post("/assets")
 async def upload_asset(...):
     # ... upload logic ...
 
-    # Notify connected clients
-    await emit_event(WebSocketEvent.UPLOAD_SUCCESS, current_user.id, asset_response_dto)
+    # Fire-and-forget; SocketIOError is swallowed inside emit_user_event.
+    await emit_user_event(WebSocketEvent.UPLOAD_SUCCESS, current_user.id, asset_response_dto)
 
     return asset_response_dto
 ```
@@ -435,9 +440,9 @@ async def upload_asset(...):
 async def delete_assets(...):
     # ... delete logic ...
 
-    # Notify connected clients for each deleted asset
+    # No try/except needed — emit_user_event swallows SocketIOError centrally.
     for asset_id in deleted_asset_ids:
-        await emit_event(WebSocketEvent.ASSET_DELETE, current_user.id, asset_id)
+        await emit_user_event(WebSocketEvent.ASSET_DELETE, current_user.id, asset_id)
 ```
 
 ---
@@ -446,9 +451,9 @@ async def delete_assets(...):
 
 ### 5.1 Error Handling
 
-- If `emit_upload_success` is called for a user with no connected clients, it's a no-op (no error)
+- If `emit_user_event` is called for a user with no connected clients, it's a no-op (no error)
 - Session lookup failures during connect result in connection rejection
-- Emit failures should be logged but not block the upload response
+- **Emit failures are swallowed centrally**: `emit_user_event` and `emit_session_event` catch `SocketIOError` from the underlying transport, log at WARN with `exc_info=True`, and return normally. Callers must NOT wrap these in `try/except SocketIOError` — the fire-and-forget contract is type-level so emit failures cannot break the request paired with them. If a caller needs to handle other exception types (e.g., DTO conversion before the emit), it can still wrap the broader block in its own try/except.
 
 ### 5.2 Payload Serialization
 
--- a/docs/architecture/websocket-implementation.md
+++ b/docs/architecture/websocket-implementation.md
@@ -427,7 +427,6 @@ from services.websockets import emit_user_event, WebSocketEvent
 async def upload_asset(...):
     # ... upload logic ...
 
-    # Fire-and-forget; SocketIOError is swallowed inside emit_user_event.
     await emit_user_event(WebSocketEvent.UPLOAD_SUCCESS, current_user.id, asset_response_dto)
 
     return asset_response_dto
@@ -440,7 +439,6 @@ async def upload_asset(...):
 async def delete_assets(...):
     # ... delete logic ...
 
-    # No try/except needed — emit_user_event swallows SocketIOError centrally.
     for asset_id in deleted_asset_ids:
         await emit_user_event(WebSocketEvent.ASSET_DELETE, current_user.id, asset_id)
 ```
--- a/routers/api/assets.py
+++ b/routers/api/assets.py
@@ -1,3 +1,4 @@
+from itertools import batched
 from typing import List, Literal, NamedTuple, cast
 from uuid import UUID, uuid4
 import base64
@@ -18,18 +19,13 @@ from fastapi import (
     status,
 )
 from fastapi.responses import JSONResponse, StreamingResponse
-from gumnut import APIStatusError, AsyncGumnut, GumnutError, NotFoundError
+from gumnut import AsyncGumnut
 from gumnut.types.asset_response import AssetResponse
 
 from config.settings import Settings, get_settings
 from routers.utils.cdn_client import DEFAULT_FORWARDED_HEADERS, stream_from_cdn
 from routers.utils.gumnut_client import get_authenticated_gumnut_client
-from routers.utils.error_mapping import (
-    log_bulk_transport_error,
-    log_upstream_response,
-    map_gumnut_error,
-    truncated_error_detail,
-)
+from routers.utils.error_mapping import map_gumnut_error
 from routers.utils.current_user import get_current_user, get_current_user_id
 from pydantic import ValidationError
 from socketio.exceptions import SocketIOError
@@ -77,6 +73,11 @@ router = APIRouter(
     tags=["assets"],
     responses={404: {"description": "Not found"}},
 )
+
+
+# Backend caps bulk-id endpoints at MAX_BULK_GET_IDS=100 per request; chunk
+# requests larger than that to stay under the cap.
+_BULK_CHUNK_SIZE = 100
 
 
 AssetVariant = Literal["thumbnail", "preview", "fullsize", "original"]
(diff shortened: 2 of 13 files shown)


taggartgorman:
@CharlieHelps review

tests/unit/api/test_assets.py
```diff
@@ -998,70 +998,65 @@ async def test_update_assets_success(self):
...
+    per id. ``False``/absent → bulk soft-delete via
+    ``client.post("/api/assets/trash", body=...)`` with one batched
+    ``on_asset_trash`` per chunk.
+    """
 
     @pytest.mark.anyio
-    async def test_delete_assets_success(self):
-        """Test successful assets deletion."""
-        # Setup - create mock client
+    async def test_delete_assets_force_false_calls_trash_endpoint(self):
```
taggartgorman, line 1011: **[Should Fix] Pin the no-swallow contract with a test**

The PR's central correctness change is removing the per-id `NotFoundError` swallowing in favor of the global `GumnutError` handler. None of the new `TestDeleteAssets` tests assert this — if a future refactor reintroduced a `try/except` around `client.post` / `client.delete`, the suite would still pass.

Repo precedent at `test_assets.py:1271` (`test_get_asset_statistics_propagates_sdk_error`) — one test per flow asserting `pytest.raises(APIStatusError)` would lock the contract:

```python
@pytest.mark.anyio
async def test_delete_assets_force_false_propagates_sdk_error(self):
    from tests.conftest import make_sdk_status_error
    mock_client = Mock()
    mock_client.post = AsyncMock(side_effect=make_sdk_status_error(500, "boom"))
    request = AssetBulkDeleteDto(ids=[uuid4()], force=False)
    with patch("routers.api.assets.emit_user_event", new_callable=AsyncMock):
        with pytest.raises(APIStatusError):
            await delete_assets(request, client=mock_client, current_user_id=uuid4())
```

Same gap applies to `test_trash.py` — `empty_trash` / `restore_trash` / `restore_assets` make the same no-swallow claim.

routers/api/trash.py
```diff
@@ -13,27 +46,135 @@
...
+    user_id = str(current_user_id)
+    trashed_gumnut_ids = await _list_trashed_ids(client)
+    for chunk in batched(trashed_gumnut_ids, BULK_CHUNK_SIZE):
+        await client.delete(
+            "/api/assets",
+            body={"ids": list(chunk)},
+            cast_to=type(None),
+        )
+        for gumnut_id in chunk:
+            asset_uuid = safe_uuid_from_asset_id(gumnut_id)
```
taggartgorman, line 68: **[Consider] `safe_uuid_from_asset_id` runs after the bulk mutation has committed**

Decoding happens inside the WS loop, after `client.delete(...)` (and same in `restore_trash` after `client.post(...)`). If the SDK ever returned a row whose `.id` lacks the `asset_` prefix, `ValueError` would fire after the destructive mutation succeeded — backend mutated, user gets a 500, WS events for that chunk and all subsequent chunks dropped.

SDK contract should hold, so this is theoretical — but cheap to harden:

```python
async def _list_trashed_ids(client: AsyncGumnut) -> list[tuple[str, UUID]]:
    return [
        (asset.id, safe_uuid_from_asset_id(asset.id))
        async for asset in client.assets.list(state="trashed", limit=BULK_CHUNK_SIZE)
    ]
```

Then both `empty_trash` and `restore_trash` consume already-decoded UUIDs and the WS emit becomes pure formatting.

routers/api/trash.py
```diff
@@ -13,27 +46,135 @@
...
+
+    Enumerates the caller's trashed ids, then issues bulk
+    ``DELETE /api/assets`` calls in chunks. Emits one ``on_asset_delete`` per
+    purged id, matching Immich's wire shape (single-id-per-event for permanent
+    deletes). Returns the total number of ids purged.
     """
-    return TrashResponseDto(count=0)
+    user_id = str(current_user_id)
+    trashed_gumnut_ids = await _list_trashed_ids(client)
+    for chunk in batched(trashed_gumnut_ids, BULK_CHUNK_SIZE):
```
taggartgorman, line 61: **[Consider] Log partial-progress count when a multi-chunk operation fails mid-flight**

If chunk 1 succeeds (100 ids permanently deleted) and chunk 2 raises, the exception propagates to the global handler. The user sees a 500, but no log line summarizes how many ids made it through — a Sentry investigator has nothing to reconcile against.

The design doc accepts the partial-failure *behavior* as intended; the gap is diagnostic visibility. Suggested:

```python
purged_so_far = 0
try:
    for chunk in batched(trashed_gumnut_ids, BULK_CHUNK_SIZE):
        await client.delete(...)
        purged_so_far += len(chunk)
        ...
except Exception:
    if purged_so_far < len(trashed_gumnut_ids):
        logger.warning("empty_trash partial failure", extra={
            "purged_count": purged_so_far,
            "total_planned": len(trashed_gumnut_ids),
            "user_id": user_id,
        })
    raise
```

Same applies to `restore_trash`.

tests/unit/api/test_trash.py
```diff
@@ -0,0 +1,307 @@
...
+        assert mock_emit.await_count == 3
+        for emit_call in mock_emit.await_args_list:
+            event, user_id, payload = emit_call.args
+            assert event == WebSocketEvent.ASSET_DELETE
+            assert user_id == str(current_user_id)
+            # payload is a single asset UUID string
+            assert payload in {str(safe_uuid_from_asset_id(gid)) for gid in gumnut_ids}
+
+    @pytest.mark.anyio
+    async def test_chunks_large_trash_lists(self):
```
taggartgorman, line 311: **[Consider] Align this chunk-size test to 250 ids for symmetry**

`TestRestoreTrash.test_chunks_large_trash_lists` and `delete_assets`'s chunking tests use 250 ids (3 chunks: 100/100/50). This one uses 150 (2 chunks: 100/50). Bumping to 250 gives uniform 3-chunk boundary coverage across the trio.

taggartgorman (commented):
Solid implementation of PR B from the trash design doc. Tests cover the new branching logic well; the design follows the doc closely. Four non-blocking comments below — one Should Fix on test coverage worth picking up, plus three Consider-level points. No blockers.

taggartgorman:
## Self-review feedback round 1

### Auto-fixed (two-way door)

- **[Should Fix] Pin no-swallow contract with tests** (`tests/unit/api/test_assets.py:1011`) — Added `test_*_propagates_sdk_error` tests for `delete_assets` (force=true/false), `restore_assets`, `restore_trash`, and `empty_trash`. Each mocks the bulk SDK call to raise `APIStatusError(500)` and asserts `pytest.raises(APIStatusError)`. Future refactors that reintroduce per-id swallowing will now break these tests.
- **[Consider] Align chunk-size test to 250 ids** (`tests/unit/api/test_trash.py:271`) — Bumped `TestEmptyTrash.test_chunks_large_trash_lists` from 150 → 250 so the trio (`delete_assets`, `restore_trash`, `empty_trash`) all exercise the same 100/100/50 chunking shape.
- Documented the propagation-test convention under *Bulk-ID Endpoints* in `docs/references/code-practices.md` so the pattern is enforced for future bulk flows.

### Skipped

- **[Consider] `safe_uuid_from_asset_id` runs after bulk mutation** (`routers/api/trash.py:69`) — Skipping. The decode failure requires the SDK to return an asset whose `.id` lacks the `asset_` prefix, which violates the SDK contract. Per the comment itself: "SDK contract should hold, so this is theoretical." Refactoring to tuple-returning helpers adds type-signature complexity for a state that can't occur today; preferred to wait for a demonstrated failure mode before defending against it.
- **[Consider] Log partial-progress count on mid-flight chunk failure** (`routers/api/trash.py:62`) — Skipping. The scenario is real (transport/5xx mid-batch), but the value is diagnostic-only — the underlying SDK error already propagates with the failed chunk's body in the global handler. Adding `try/except` + counter + warning log on each of three trash flows + `delete_assets` is non-trivial complexity for an edge case that hasn't actually been observed in Sentry. Easy to add later if forensics demand it.

tests/unit/api/test_assets.py
taggartgorman: Fixed — added `test_*_propagates_sdk_error` tests for `delete_assets` (force=true/false) and the three trash flows (`restore_assets`, `restore_trash`, `empty_trash`). Each mocks the bulk SDK call to raise `APIStatusError(500)` and asserts `pytest.raises(APIStatusError)`. Also documented the propagation-test convention under *Bulk-ID Endpoints* in `docs/references/code-practices.md`.

routers/api/trash.py
taggartgorman: Skipping — the comment itself notes this is theoretical ("SDK contract should hold"). The decode failure requires the SDK to return an asset whose `.id` lacks the `asset_` prefix, which violates the SDK contract. Per the feedback-evaluation guidance, theoretical risks with non-zero complexity cost (tuple-returning helper, more type signature) default to rejection. Easy to revisit if a real SDK contract violation surfaces.

routers/api/trash.py
taggartgorman: Skipping — the underlying SDK error already propagates to the global handler with the failed chunk body, so reviewers have *some* signal. Adding try/except + counter + warning log across all three trash flows + delete_assets is non-trivial complexity for diagnostic value on an edge case that has not actually been observed in Sentry. Will revisit if forensics on a real partial-failure incident demand it.

tests/unit/api/test_trash.py
taggartgorman: Fixed — bumped to 250 ids so all three chunk-boundary tests (`delete_assets`, `restore_trash`, `empty_trash`) exercise the same 100/100/50 shape.

routers/api/assets.py
```diff
@@ -595,71 +594,90 @@ async def delete_assets(
...
+    for chunk in batched(asset_uuids, BULK_CHUNK_SIZE):
+        gumnut_ids = [uuid_to_gumnut_asset_id(uid) for uid in chunk]
+        await client.post(
+            "/api/assets/trash",
+            body={"ids": gumnut_ids},
+            cast_to=type(None),
+        )
+        chunk_uuid_strs = [str(uid) for uid in chunk]
+        try:
+            await emit_user_event(
```
ternarybits: Do we always catch SocketIOErrors from `emit_user_event`, log, and ignore them?  If so, perhaps we should move the catching and logging into `emit_user_event`?

routers/api/assets.py
```diff
@@ -595,71 +594,90 @@ async def delete_assets(
...
+                chunk_uuid_strs,
             )
-        except GumnutError as asset_error:
-            # Immich expects best-effort partial deletion; record per-item.
-            log_bulk_transport_error(
-                logger,
-                context="delete_assets",
-                exc=asset_error,
+        except SocketIOError as ws_error:
+            logger.warning(
```
ternarybits: Include exception info in warnings?  I think you said in Slack that we could do that?

services/websockets.py
```diff
@@ -35,7 +35,12 @@ class WebSocketEvent(Enum):
     # Phase 1: Can implement now
     UPLOAD_SUCCESS = "on_upload_success"
     ASSET_UPLOAD_READY_V1 = "AssetUploadReadyV1"
+    # ASSET_DELETE carries a single id per event; ASSET_TRASH/ASSET_RESTORE
+    # carry an array of ids per event — matches Immich's wire contract:
+    # on_asset_delete: [string], on_asset_trash/on_asset_restore: [string[]].
     ASSET_DELETE = "on_asset_delete"
+    ASSET_TRASH = "on_asset_trash"
```
ternarybits, line 43: **[Should Fix]** Update stale `docs/architecture/websocket-implementation.md`

The architecture doc currently lists `on_asset_trash` and `on_asset_restore` under "Not Applicable (Feature Not Supported)" with the reason "photos-api hard-deletes (no soft-delete)" (lines 174-185). It also shows a `WebSocketEvent` enum example (lines 217-225) that omits the new variants, while the usage example below it uses `WebSocketEvent.ASSET_TRASH` — contradicting itself.

Since this PR contradicts that doc, worth a quick edit in the same change (move both events into the supported events table and update the enum snippet) so a future reader doesn't see "feature not supported" claims that conflict with the working code.

routers/api/trash.py
```diff
@@ -13,27 +46,135 @@
...
+async def empty_trash(
+    client: AsyncGumnut = Depends(get_authenticated_gumnut_client),
+    current_user_id: UUID = Depends(get_current_user_id),
+) -> TrashResponseDto:
+    """Permanently delete every trashed asset belonging to the caller.
+
+    Enumerates the caller's trashed ids, then issues bulk
+    ``DELETE /api/assets`` calls in chunks. Emits one ``on_asset_delete`` per
+    purged id, matching Immich's wire shape (single-id-per-event for permanent
+    deletes). Returns the total number of ids purged.
```
ternarybits: **[Should Fix]** Document the partial-success classification consistently across the three trash endpoints

The sibling `restore_assets` docstring is explicit about its count semantics:

> "Already-live ids are silently skipped on the backend; the returned count therefore reflects the request size, not the number of rows that actually transitioned (the backend's restore endpoint returns 204 with no count)."

But this docstring on `empty_trash` ("Returns the total number of ids purged") and the one on `restore_trash` at line 97 ("returns the total restored count") both claim a transition count. In practice they return `len(trashed_gumnut_ids)` — the count enumerated upfront. Between enumeration and the chunked bulk calls, an asset could be transitioned by another concurrent request, so the returned count can diverge from reality.

The race is small but the inconsistency between sibling endpoints is the real issue — readers will reasonably expect uniform semantics. Suggest adding the same caveat language used in `restore_assets` to both `restore_trash` and `empty_trash`.

ternarybits (commented):
This PR is faithful to the trash-soft-delete design doc, and pinning the no-swallow contract via the new `*_propagates_sdk_error` tests is worth carrying forward to other bulk endpoints. A couple of Should Fix items below; both are doc/comment only and neither blocks merge.

The deliberate divergences from the design doc (collect-then-chunk in `_list_trashed_ids` rather than streaming, and enumerate-then-chunked-delete for `/empty` rather than calling `POST /api/assets/empty-trash` once) are well-motivated and explained in code + PR description.

taggartgorman:
## Self-review feedback round 2

### Auto-fixed (two-way door)

- **[Should Fix] Centralize SocketIOError handling in `emit_user_event` / `emit_session_event`** — Moved the catch+log+continue pattern from 7 duplicate sites into the emit functions themselves. Added `exc_info=True` to the WARN log so the traceback is preserved. Updated tests to patch `services.websockets._emit_event` so the SocketIOError originates inside the centralized handler, and added unit tests for the swallow contract on each emit function.
- **[Should Fix] Update stale `docs/architecture/websocket-implementation.md`** — Moved `on_asset_trash` / `on_asset_restore` out of the "Not Applicable" table into supported events, added them to the `WebSocketEvent` enum snippet and Implementation Priority list, fixed the assets-router usage example to show the no-try/except style, and documented the central swallow in section 5.1.
- **[Should Fix] Document partial-success classification consistently across the trash trio** — Added the same upfront-count caveat already on `restore_assets` to `restore_trash` and `empty_trash` docstrings, noting that concurrent transitions between enumeration and the chunked bulk calls can make the returned count diverge from rows actually transitioned.
- Added a "WebSocket Emission" note to `docs/references/code-practices.md` so future call sites don't reintroduce per-site catches.

### Verification

- 754 tests pass, lint and type checks clean.
- Net diff: -120 / +152 across 12 files; +29 net lines, but ~30 lines of duplicated try/except removed and replaced with one centralized handler + 2 swallow tests + docstring/doc updates.

routers/api/assets.py
taggartgorman: Fixed — moved the catch+log+continue into `emit_user_event` / `emit_session_event`. Removed all 7 duplicate try/except blocks across `trash.py`, `assets.py`, `auth.py`, `sessions.py`. Added unit tests for the swallow contract on each emit function. The fire-and-forget contract is now type-level.

routers/api/assets.py
taggartgorman: Fixed — `logger.warning(..., exc_info=True)` is now used in both `emit_user_event` and `emit_session_event` so the WARN carries a traceback.

services/websockets.py
taggartgorman: Fixed — moved `on_asset_trash` and `on_asset_restore` into the supported events table (lines 162-164), added them to the `WebSocketEvent` enum snippet, updated the Implementation Priority list, and fixed the assets-router usage example to reflect the no-try/except style. Also documented the central SocketIOError swallow in section 5.1.

routers/api/trash.py
taggartgorman: Fixed — both `empty_trash` and `restore_trash` docstrings now carry the same upfront-count caveat as `restore_assets`: the returned count reflects the enumerated id list, not rows that actually transitioned, and concurrent transitions can make the two diverge. Backend idempotency on already-transitioned rows is also called out.

docs/architecture/websocket-implementation.md
```diff
@@ -416,14 +421,14 @@ async def emit_event(event: WebSocketEvent, user_id: str, payload: EventPayload
...
-from routers.api.websockets import emit_event, WebSocketEvent
+from services.websockets import emit_user_event, WebSocketEvent
 
 @router.post("/assets")
 async def upload_asset(...):
     # ... upload logic ...
 
-    # Notify connected clients
-    await emit_event(WebSocketEvent.UPLOAD_SUCCESS, current_user.id, asset_response_dto)
+    # Fire-and-forget; SocketIOError is swallowed inside emit_user_event.
```
ternarybits: I don't think this comment about SocketIOError is necessary.

docs/architecture/websocket-implementation.md
```diff
@@ -435,9 +440,9 @@ async def upload_asset(...):
 async def delete_assets(...):
     # ... delete logic ...
 
-    # Notify connected clients for each deleted asset
+    # No try/except needed — emit_user_event swallows SocketIOError centrally.
```
ternarybits: I don't think this comment about SocketIOError is necessary.

routers/api/auth.py
```diff
@@ -82,22 +81,17 @@ async def post_logout(
...
-                extra={"session_id": session_token, "error": str(ws_error)},
-            )
         except Exception as e:
             # Log but don't fail the logout - cookie clearing is more important
             logger.warning(
                 "Failed to delete session on logout",
                 extra={"error": str(e)},
                 exc_info=True,
             )
+        # emit_session_event swallows SocketIOError internally — fire-and-forget.
```
ternarybits: I don't think this comment about SocketIOError is necessary.

routers/api/sessions.py
```diff
@@ -199,16 +198,10 @@ async def delete_session(
...
-    try:
-        await emit_session_event(
-            WebSocketEvent.SESSION_DELETE, session_token, session_token
-        )
-    except SocketIOError as ws_error:
-        logger.warning(
-            "Failed to emit WebSocket event after session delete",
-            extra={"session_id": session_token, "error": str(ws_error)},
-        )
+    # emit_session_event swallows SocketIOError internally — fire-and-forget.
```
ternarybits: I don't think this comment about SocketIOError is necessary.

ternarybits (approved):

docs/architecture/websocket-implementation.md
taggartgorman: Removed unnecessary comments about SocketIOError

PR #147, opened 2026-04-30
Fix reversed path/body semantics in PUT /faces/{id} (GUM-679)

ternarybits:
## Summary

- **Bug fix**: `PUT /api/faces/{id}` had path and body decoded in inverted directions vs. Immich's actual contract. Per Immich's `face.controller.ts` and `person.service.ts:111` (`reassignFacesById(auth, personId, dto)`), `{id}` is the **target person** and `FaceDto.id` is the face being reassigned. The adapter was treating `{id}` as a face id and `request.id` as a person id, so every PUT from Immich web returned 404 (`"Face '...' not found in library"`).
- **Test fix**: `tests/unit/api/test_faces.py` baked in the same swap, so the existing test passed while the production behavior was wrong. Tests now lock in the correct shape; a class-level docstring documents the surprising semantics.
- **Doc sharpening**: The gotcha at `code-practices.md:97` was added in PR #124 (GUM-505) when the same swap was fixed in `PUT /people/{id}/reassign` — but it only listed the people endpoint as the example. The faces-side bug was introduced earlier the same day in PR #123 and survived. Now both reassign endpoints are listed by name.
- **Process learning**: Added an "audit siblings when fixing class-of-bug issues" step to the same doc, so future fixes for class-of-bug traps don't leave latent bugs in adjacent endpoints.

## Sibling-endpoint audit

To make sure no other adapter endpoint has the same path/body swap, I audited every `@router.{method}("/{id}")`-style handler in `routers/api/` against Immich's controllers and services. Verdict: **`PUT /faces/{id}` was the only remaining case** — no other adapter endpoint inverts path/body semantics relative to Immich's contract.

Confirmed clean:

| File | Endpoints checked | Result |
|------|------------------|--------|
| `faces.py` | `DELETE /faces/{id}`, `GET /faces` | OK — `{id}` is the face id, matches Immich |
| `people.py` | `PUT /people/{id}/reassign`, person CRUD endpoints | OK — already fixed in PR #124 |
| `albums.py` | `PUT/DELETE /albums/{id}/assets`, `PATCH /albums/{id}` | OK — `{id}` is the album, body items reference assets |
| `assets.py` | asset-by-id endpoints | OK |
| `tags.py`, `partners.py`, `memories.py`, `stacks.py`, `shared_links.py` | id-in-path handlers | OK — path id consistently matches the URL collection |

## Linear

https://linear.app/gumnut-ai/issue/GUM-679/failed-to-assign-face-from-brigi-to-pamela

## Test plan

- [x] `uv run ruff format && uv run ruff check` — clean
- [x] `uv run pytest` — 731 passed (including the reshaped face reassign tests)
- [ ] Manual smoke test from Immich web: navigate to a person's photo, reassign one of the faces to a different person; confirm the request returns 200 and the face moves to the target person.

## Notes for reviewer

The Immich endpoint name and URL are misleading — `PUT /faces/{id}` looks like it should modify the face named in the path. It does not. Both reassign-style endpoints in Immich (`PUT /people/{id}/reassign` and `PUT /faces/{id}`) accept the **target person** in the path. The fixed docstring quotes Immich's own description verbatim and cites the upstream source.

### DIFF
--- a/docs/references/code-practices.md
+++ b/docs/references/code-practices.md
@@ -100,6 +100,10 @@ Forgetting step 2 causes silent drift — the served web UI stays on the old Imm
 5. **Validate compatibility**: Run `validate_api_compatibility.py` to ensure correct implementation
 6. **Test endpoints**: Verify responses match Immich API expectations
 
+### Fixing Class-of-Bug Issues
+
+When a bug rooted in a recurring trap is found in one endpoint (e.g., a path/body semantic mismatch, an ID-prefix decoding mistake, or a missing source/target distinction), **audit sibling endpoints in the same router and adjacent routers for the same pattern** before closing the fix. Both reassign endpoints in the gotcha above shipped with the same swap on the same day in different PRs; fixing one without auditing the other left a production bug latent for weeks. A one-line search (`grep -rn` for the pattern, or read each handler in the affected router) is cheap insurance against the trap recurring.
+
 ### Exception Handling
 
 - Don't expose implementation details in exceptions thrown to consumers
--- a/docs/references/code-practices.md
+++ b/docs/references/code-practices.md
@@ -97,12 +97,9 @@ Forgetting step 2 causes silent drift — the served web UI stays on the old Imm
 4. **Verify parameter semantics**: Check the Immich OpenAPI spec (`https://api.immich.app/endpoints/`) or source code (`immich/server/src/controllers/*.controller.ts` and the matching service) to confirm what each URL path and body parameter represents. URL `{id}` parameters don't always refer to the entity in the URL collection — face/person reassign endpoints in particular swap the natural reading. Both of these accept the **target person** as `{id}` in the path:
    - `PUT /people/{id}/reassign` — `{id}` is the target person (reassign TO); body items are sources.
    - `PUT /faces/{id}` — `{id}` is the target person (reassign TO); body `FaceDto.id` is the face being reassigned.
+   - When fixing a path/body or ID-decoding bug in one handler, audit sibling handlers in the same router (and adjacent routers) for the same trap before closing the fix. A one-line search (`grep -rn` for the pattern) is cheap insurance against the same class-of-bug recurring.
 5. **Validate compatibility**: Run `validate_api_compatibility.py` to ensure correct implementation
 6. **Test endpoints**: Verify responses match Immich API expectations
-
-### Fixing Class-of-Bug Issues
-
-When a bug rooted in a recurring trap is found in one endpoint (e.g., a path/body semantic mismatch, an ID-prefix decoding mistake, or a missing source/target distinction), **audit sibling endpoints in the same router and adjacent routers for the same pattern** before closing the fix. Both reassign endpoints in the gotcha above shipped with the same swap on the same day in different PRs; fixing one without auditing the other left a production bug latent for weeks. A one-line search (`grep -rn` for the pattern, or read each handler in the affected router) is cheap insurance against the trap recurring.
 
 ### Exception Handling
 
--- a/docs/references/code-practices.md
+++ b/docs/references/code-practices.md
@@ -94,7 +94,9 @@ Forgetting step 2 causes silent drift — the served web UI stays on the old Imm
 1. **Generate models**: Use `generate_immich_models.py` to create up-to-date Pydantic models (see [development tools](development-tools.md))
 2. **Import models**: Use generated models from `routers.immich_models` for type safety
 3. **Define parameters**: Follow the parameter conventions above
-4. **Verify parameter semantics**: Check the Immich OpenAPI spec (`https://api.immich.app/endpoints/`) or source code to confirm what each URL path and body parameter represents. URL `{id}` parameters don't always refer to the entity being queried — e.g., in `PUT /people/{id}/reassign`, `{id}` is the target person (reassign TO), not the source.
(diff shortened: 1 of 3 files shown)


tests/unit/api/test_faces.py
```diff
@@ -262,7 +262,12 @@ async def test_sdk_error_propagates(self):
 
 
 class TestReassignFace:
-    """Test the reassign_faces_by_id endpoint."""
+    """Test the reassign_faces_by_id endpoint.
+
+    Immich's PUT /faces/{id} contract is non-obvious: the URL `{id}` is the
+    target person (reassign TO) and the body `id` is the face being reassigned.
+    See routers/api/faces.py for the citation; these tests lock in that shape.
+    """
```
taggartgorman, line 270: **[Consider]** Class docstring largely restates the production docstring on `reassign_faces_by_id`. A reader investigating these tests will already see the canonical contract one click away in `routers/api/faces.py`. Optional shrink:

```suggestion
    """Test the reassign_faces_by_id endpoint.

    See routers/api/faces.py for the path/body contract these tests lock in.
    """
```

Non-blocking — keep as-is if you want the contract restated locally for grep visibility.

docs/references/code-practices.md
```diff
@@ -94,10 +94,16 @@ Forgetting step 2 causes silent drift — the served web UI stays on the old Imm
...
+   - `PUT /people/{id}/reassign` — `{id}` is the target person (reassign TO); body items are sources.
+   - `PUT /faces/{id}` — `{id}` is the target person (reassign TO); body `FaceDto.id` is the face being reassigned.
 5. **Validate compatibility**: Run `validate_api_compatibility.py` to ensure correct implementation
 6. **Test endpoints**: Verify responses match Immich API expectations
 
+### Fixing Class-of-Bug Issues
+
+When a bug rooted in a recurring trap is found in one endpoint (e.g., a path/body semantic mismatch, an ID-prefix decoding mistake, or a missing source/target distinction), **audit sibling endpoints in the same router and adjacent routers for the same pattern** before closing the fix. Both reassign endpoints in the gotcha above shipped with the same swap on the same day in different PRs; fixing one without auditing the other left a production bug latent for weeks. A one-line search (`grep -rn` for the pattern, or read each handler in the affected router) is cheap insurance against the trap recurring.
```
taggartgorman: **[Consider]** The new section is mostly incident-narrative ("Both reassign endpoints in the gotcha above shipped with the same swap on the same day") wrapped around one prescriptive sentence. The "gotcha above" is also a soft cross-reference that will silently break if the doc is restructured. Two ways to tighten:

1. Fold the prescription into step 4 of "Implementing New Endpoints" as a sub-bullet ("When fixing a path/body or ID-decoding bug in one handler, audit sibling handlers in the same router for the same trap") — avoids a new section and the fragile back-reference.
2. Or keep the section but strip the historical detail and replace "the gotcha above" with a direct anchor to the bullets.

Non-blocking — the guidance is reasonable; it just reads more like a PR retrospective than a durable convention right now.

taggartgorman (approved):
Approving on the fast path — the fix itself is a minimal, correct two-line swap, the tests are reshaped to lock in the right contract, and the sibling-endpoint audit in the PR description is the right level of rigor for a class-of-bug trap.

## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (2 Consider items inline below — non-blocking)
      Rationale: A previously broken endpoint now works; the test that was passing against the wrong behavior is corrected; sibling endpoint at `routers/api/people.py:401` already followed the correct pattern, so this brings `faces.py` in line. Smaller-scoped alternative considered (just the swap) — rejected because the docstring captures non-obvious upstream contract worth preserving in code.
      Evidence: Diff limited to `routers/api/faces.py`, `tests/unit/api/test_faces.py`, `docs/references/code-practices.md`; no auth modules touched.
      Evidence: Endpoint performs `client.faces.update(face_id, person_id=...)` — a CRUD update, not deletion. Behavior moves from "always 404" to "reassign as documented by Immich," which is the intended product behavior.
      Rationale: Reassignment changes which person owns a face; this matches Immich's documented contract and the sibling `PUT /people/{id}/reassign` endpoint.
      Rationale: 2-line swap, sibling endpoint at `people.py:401` confirms the path/body shape, decoders have distinct prefixes (`face_` vs `person_`) so a misuse fails loudly, unit tests cover success + error propagation.

ternarybits:
## Feedback addressed

**Auto-fixed (two-way door)**

- **`docs/references/code-practices.md`** — Folded the standalone "Fixing Class-of-Bug Issues" section into step 4 of "Implementing New Endpoints" as a sub-bullet. Removes the fragile "the gotcha above" back-reference and the incident narrative; keeps the durable audit-siblings convention co-located with the path/body semantics gotcha. (commit 9c9e40b)

**Skipped**

- **`tests/unit/api/test_faces.py:264-270`** — Class docstring left as-is. The reviewer flagged this as non-blocking and explicitly endorsed either approach ("keep as-is if you want the contract restated locally for grep visibility"). Locking the non-obvious Immich contract directly in the test file provides grep value when investigating test failures, and the docstring already includes a pointer to `routers/api/faces.py` for the canonical citation.

docs/references/code-practices.md
ternarybits: Fixed in 9c9e40b — folded into step 4 of `Implementing New Endpoints` as a sub-bullet. Drops the `the gotcha above` back-reference and the incident narrative; keeps the actionable audit-siblings rule co-located with the path/body semantics gotcha.

tests/unit/api/test_faces.py
ternarybits: Skipping per your endorsement of either approach — keeping the contract restated locally for grep visibility when investigating test failures. The docstring already includes the pointer to `routers/api/faces.py`, so a reader still has the canonical citation one click away.

PR #150, opened 2026-05-01
Add Max-Age to auth cookies; make validateToken actually validate

taggartgorman:
## Summary

- Auth cookies (`immich_access_token`, `immich_auth_type`, `immich_is_authenticated`, plus the refresh path) are now set with `Max-Age=400 days` instead of as session cookies. iOS `HTTPCookieStorage` only persists cookies that carry an explicit Max-Age; without it, iOS drops them when it reaps the backgrounded app, forcing a re-login on the next cold launch. The 400-day window matches what the upstream Immich server and the iOS client encode as the contract.
- `POST /api/auth/validateToken` was a stub that returned `authStatus=True` regardless of input. Because the iOS auth guard calls this on app launch and trusts the result, an unauthenticated client was let past the login gate and the missing-cookie failure only surfaced on the next API call (presenting as a sudden mid-session expiry). It now requires `request.state.jwt_token` and 401s otherwise.
- Documents the 400-day cookie contract in `docs/architecture/adapter-architecture.md` and adds a "fail closed on auth/authz checks" rule for stub endpoints in `docs/references/code-practices.md`.

## Linear

https://linear.app/gumnut-ai/issue/GUM-690/immich-adapter-auth-cookies-missing-max-age-cause-ios-to-lose-session

## Test plan

- [x] Unit tests assert `Max-Age=34560000` on all four `Set-Cookie` headers (3 from `set_auth_cookies`, 1 from `update_access_token_cookie`).
- [x] Unit tests cover `validate_access_token` for JWT-present (returns `authStatus=True`), missing `jwt_token` attribute (401), and `jwt_token=None` (401).
- [x] Full pytest (768 + 23 new) green; ruff + pyright clean.
- [ ] iOS manual verification: fresh login on the iOS Immich client, inspect `Set-Cookie` on `/api/oauth/callback` and confirm `Max-Age=34560000`.
- [ ] iOS manual verification: force-quit the app, relaunch, tap Backup; confirm the request includes the `immich_access_token` cookie and the sync stream returns 200.
- [ ] iOS manual verification: clear the cookie (logout or app data wipe), reopen the app, confirm the auth guard sends the user back to login instead of letting them into the home screen.

### DIFF
--- a/docs/architecture/adapter-architecture.md
+++ b/docs/architecture/adapter-architecture.md
@@ -83,7 +83,7 @@ Session storage is ~3KB per device, enabling horizontal scaling of the adapter.
 ### Session lifecycle
 
 - **TTL**: Session Redis keys are set to expire based on the underlying JWT's expiry time. When a JWT is refreshed, the TTL is updated. Sessions with no expiry persist until stale cleanup (90+ days inactive).
-- **Cookie flags**: `HttpOnly`, `Secure` (protocol-aware — disabled for local HTTP dev), `SameSite=lax`, `Max-Age=400 days` (`COOKIE_MAX_AGE_SECONDS` in `routers/utils/cookies.py`).
+- **Cookie flags**: `Secure` (protocol-aware — disabled for local HTTP dev), `SameSite=lax`, `Max-Age=400 days` (`COOKIE_MAX_AGE_SECONDS` in `routers/utils/cookies.py`); `ACCESS_TOKEN` and `AUTH_TYPE` are `HttpOnly`, `IS_AUTHENTICATED` is intentionally JS-readable as a frontend-visible flag.
 - **Why 400 days**: Without `Max-Age`/`Expires`, browsers and `HTTPCookieStorage` on iOS treat `Set-Cookie` as a session cookie that lives only in memory. iOS drops in-memory cookies when it reaps the backgrounded app, forcing a re-login on every cold launch. The upstream Immich server and the iOS client both encode a 400-day lifetime; the adapter matches so the contract is consistent across the stack. If you ever bound session lifetime, change `COOKIE_MAX_AGE_SECONDS` and the Redis session TTL together.
 - **Logout**: Deletes the Redis session key, clears cookies, emits `on_session_delete` WebSocket event to notify connected clients
 - **JWT refresh failure**: If the backend cannot refresh an expired JWT, the next request using that session returns 401. The client must re-authenticate via OAuth.
--- a/docs/architecture/adapter-architecture.md
+++ b/docs/architecture/adapter-architecture.md
@@ -83,7 +83,8 @@ Session storage is ~3KB per device, enabling horizontal scaling of the adapter.
 ### Session lifecycle
 
 - **TTL**: Session Redis keys are set to expire based on the underlying JWT's expiry time. When a JWT is refreshed, the TTL is updated. Sessions with no expiry persist until stale cleanup (90+ days inactive).
-- **Cookie flags**: `HttpOnly`, `Secure` (protocol-aware — disabled for local HTTP dev), `SameSite=lax`
+- **Cookie flags**: `HttpOnly`, `Secure` (protocol-aware — disabled for local HTTP dev), `SameSite=lax`, `Max-Age=400 days` (`COOKIE_MAX_AGE_SECONDS` in `routers/utils/cookies.py`).
+- **Why 400 days**: Without `Max-Age`/`Expires`, browsers and `HTTPCookieStorage` on iOS treat `Set-Cookie` as a session cookie that lives only in memory. iOS drops in-memory cookies when it reaps the backgrounded app, forcing a re-login on every cold launch. The upstream Immich server and the iOS client both encode a 400-day lifetime; the adapter matches so the contract is consistent across the stack. If you ever bound session lifetime, change `COOKIE_MAX_AGE_SECONDS` and the Redis session TTL together.
 - **Logout**: Deletes the Redis session key, clears cookies, emits `on_session_delete` WebSocket event to notify connected clients
 - **JWT refresh failure**: If the backend cannot refresh an expired JWT, the next request using that session returns 401. The client must re-authenticate via OAuth.
 
--- a/routers/utils/cookies.py
+++ b/routers/utils/cookies.py
@@ -41,7 +41,8 @@ def set_auth_cookies(
                  If not provided, secure flag defaults to True.
 
     Security flags:
-        - HttpOnly: True (prevents JavaScript access, XSS protection)
+        - HttpOnly: True for ACCESS_TOKEN and AUTH_TYPE; IS_AUTHENTICATED is
+          intentionally JS-readable so the frontend can branch on it
         - Secure: True by default, or based on secure parameter
         - SameSite: "lax" (CSRF protection while allowing some cross-site navigation)
         - Max-Age: COOKIE_MAX_AGE_SECONDS (persists across iOS app process death)
--- a/routers/utils/cookies.py
+++ b/routers/utils/cookies.py
@@ -1,6 +1,13 @@
 from fastapi import Response
 from enum import Enum
 
+# 400-day Max-Age on auth cookies. Without an explicit Max-Age these become
+# session cookies, which iOS HTTPCookieStorage holds in memory only and drops
+# on app process death — the upstream Immich server and iOS client both encode
+# a 400-day cookie lifetime, so the adapter must match to avoid forcing
+# re-login every time iOS reaps the backgrounded app.
+COOKIE_MAX_AGE_SECONDS = 400 * 24 * 60 * 60
+
 
 class ImmichCookie(str, Enum):
     ACCESS_TOKEN = "immich_access_token"
@@ -37,11 +44,13 @@ def set_auth_cookies(
         - HttpOnly: True (prevents JavaScript access, XSS protection)
         - Secure: True by default, or based on secure parameter
         - SameSite: "lax" (CSRF protection while allowing some cross-site navigation)
+        - Max-Age: COOKIE_MAX_AGE_SECONDS (persists across iOS app process death)
     """
     # Set access token cookie (most sensitive, always HttpOnly)
     response.set_cookie(
         key=ImmichCookie.ACCESS_TOKEN.value,
         value=access_token,
+        max_age=COOKIE_MAX_AGE_SECONDS,
         httponly=True,
         secure=secure,
         samesite="lax",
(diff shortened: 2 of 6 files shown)


routers/utils/cookies.py
```diff
@@ -37,11 +44,13 @@ def set_auth_cookies(
         - HttpOnly: True (prevents JavaScript access, XSS protection)
         - Secure: True by default, or based on secure parameter
         - SameSite: "lax" (CSRF protection while allowing some cross-site navigation)
+        - Max-Age: COOKIE_MAX_AGE_SECONDS (persists across iOS app process death)
```
taggartgorman, line 48: This "Security flags" block now lists `Max-Age` alongside `HttpOnly` / `Secure` / `SameSite` as if all four applied to every cookie this function sets — but `IS_AUTHENTICATED` is intentionally not `HttpOnly` (lines 67-72 omit `httponly=True` and the inline comment calls it a "convenience flag for frontend" — only useful if JS can read it). Since this PR is already editing this block, worth scoping the `HttpOnly` line so a future reader doesn't "fix" `IS_AUTHENTICATED` to match the docstring.

Suggestion:

```
        - HttpOnly: True for ACCESS_TOKEN and AUTH_TYPE; IS_AUTHENTICATED is
          intentionally JS-readable so the frontend can branch on it
```

docs/architecture/adapter-architecture.md
```diff
@@ -83,7 +83,8 @@ Session storage is ~3KB per device, enabling horizontal scaling of the adapter.
 ### Session lifecycle
 
 - **TTL**: Session Redis keys are set to expire based on the underlying JWT's expiry time. When a JWT is refreshed, the TTL is updated. Sessions with no expiry persist until stale cleanup (90+ days inactive).
-- **Cookie flags**: `HttpOnly`, `Secure` (protocol-aware — disabled for local HTTP dev), `SameSite=lax`
+- **Cookie flags**: `HttpOnly`, `Secure` (protocol-aware — disabled for local HTTP dev), `SameSite=lax`, `Max-Age=400 days` (`COOKIE_MAX_AGE_SECONDS` in `routers/utils/cookies.py`).
```
taggartgorman: Same issue as the `cookies.py` docstring: this bullet now reads "HttpOnly, Secure, SameSite=lax, Max-Age=400 days" as if every auth cookie carried all four flags, but `IS_AUTHENTICATED` is deliberately JS-readable. Since the line is being edited to add `Max-Age`, this is the natural moment to scope the `HttpOnly` claim correctly — the architecture doc is evergreen and this is what future audits will read.

Suggestion:

```
- **Cookie flags**: `Secure` (protocol-aware — disabled for local HTTP dev), `SameSite=lax`, `Max-Age=400 days` (`COOKIE_MAX_AGE_SECONDS` in `routers/utils/cookies.py`); `ACCESS_TOKEN` and `AUTH_TYPE` are `HttpOnly`, `IS_AUTHENTICATED` is intentionally JS-readable as a frontend-visible flag.
```

taggartgorman (commented):
The functional change is clean — Max-Age plumbing is consistent across all four `set_cookie` sites, `validateToken` correctly checks `request.state.jwt_token`, the 401 path flows through `_immich_http_exception_handler` and emits the Immich shape, and tests cover all new branches. Both findings below are doc-tightening on paragraphs the PR is already editing — not blocking, but worth a single follow-up commit while the context is fresh.

taggartgorman:
## Round 1 — review feedback addressed

Both findings from the self-review were the same issue (a docstring + the architecture doc both globally listed `HttpOnly` as a flag applied to every auth cookie the helper sets, while `IS_AUTHENTICATED` is intentionally JS-readable as a frontend convenience flag). Both fixed in `1e6d5d6` using the suggested wording.

### Auto-fixed (two-way door)

- **`routers/utils/cookies.py:47` docstring** — Scoped `HttpOnly` to `ACCESS_TOKEN` and `AUTH_TYPE`; called out `IS_AUTHENTICATED` as intentionally JS-readable.
- **`docs/architecture/adapter-architecture.md` Cookie flags bullet** — Same scoping in the architecture doc.

No behavior change; existing tests still cover the per-cookie security flags.

routers/utils/cookies.py
taggartgorman: Fixed in 1e6d5d6 — used the suggested wording.

docs/architecture/adapter-architecture.md
taggartgorman: Fixed in 1e6d5d6 — used the suggested wording.

tests/unit/utils/test_cookies.py
```diff
@@ -98,6 +99,24 @@ def test_different_auth_types(self, client):
         set_cookie_header = response.headers.get("set-cookie", "")
         assert "Secure" not in set_cookie_header
 
+    def test_all_cookies_have_max_age(self, client):
```
ternarybits: **[Consider]** Max-Age assertion only exercises the `secure=True` branch

This test calls `/test/set-auth-cookies` (default `secure=True`). The `secure=False` branch is exercised by `test_different_auth_types` via `/test/set-auth-cookies-with-secure`, but that test only asserts on `Secure not in header` and never checks for `Max-Age`. If a future refactor wraps `max_age=...` in a conditional that depends on `secure`, or drops it from one of the three cookie writes, the suite won't catch it.

Cheap to close — extend this test to also fetch the `_with_secure` endpoint, or add a parallel `_when_insecure` test (~5 lines).

tests/unit/utils/test_cookies.py
```diff
@@ -140,3 +159,20 @@ def test_with_request_object(self, client):
         # Access token should have secure flag set to False
         set_cookie_header = response.headers.get("set-cookie", "")
         assert "Secure" not in set_cookie_header
+
+    def test_refreshed_token_has_max_age(self, client):
```
ternarybits: **[Consider]** Same Max-Age `secure=False` gap on `update_access_token_cookie`

Parallel to the suggestion on `test_all_cookies_have_max_age`. This test only covers `secure=True`. `test_with_request_object` exercises `secure=False` but doesn't assert `Max-Age`. Suggest extending this test to also fetch `/test/update-token-with-secure` (~3 lines).

tests/unit/api/test_auth.py
```diff
@@ -210,3 +211,46 @@ async def test_logout_websocket_error_does_not_fail_logout(
             # Logout should still succeed despite WebSocket error
             assert result.successful is True
             mock_session_store.delete.assert_called_once()
+
+
+class TestValidateAccessToken:
```
ternarybits, line 221: **[Consider]** No integration test exercising `validateToken` through the auth middleware

These tests build `request.state` with `Mock(spec=...)` and call the route function directly — proving the endpoint's branch logic in isolation. But the load-bearing claim of this PR (per the docstring and the new `code-practices.md` guidance) is that returning `authStatus=True` from a stub lets unauthenticated clients past the iOS auth guard. The middleware is what's responsible for populating `request.state.jwt_token` correctly, and that wiring is exactly what the unit tests mock.

A middleware regression that fails to populate `jwt_token` on a real authenticated request would silently 401 every login; one that populates it for unauthenticated requests would re-introduce the original bug. Neither is caught here.

Suggest adding a TestClient-based integration test that hits `POST /api/auth/validateToken` end-to-end: one authed (200 + `authStatus=True`), one unauthed (401). Matches the existing TestClient pattern in `test_cookies.py`.

ternarybits (approved):
Solid two-fer fix — the iOS HTTPCookieStorage reap behavior and the auth-guard trust assumption surface as the same user-visible "mid-session expiry" symptom, so packaging them is the right call. The new code-practices.md guidance on fail-closed stubs is load-bearing and likely to prevent a class of future regressions.

No Must Fix or Should Fix issues. Three test-coverage suggestions below — all Consider-level and non-blocking. The 400-day Max-Age choice matches the RFC 6265bis cap and the upstream Immich/iOS contract; the endpoint-level JWT check correctly sits behind the auth middleware (which 401s on session-lookup failure and sets `jwt_token=None` when no session token is provided, so the endpoint covers the "no session token at all" case).

taggartgorman:
## Round 2 — review feedback addressed

All three Consider-level test-coverage suggestions from @ternarybits addressed in `631d4c6`. No production code changes; all 772 tests pass, ruff/pyright clean.

### Auto-fixed (two-way door)

- **`tests/unit/utils/test_cookies.py:102` — `test_all_cookies_have_max_age` `secure=False` gap.** Parametrized over `/test/set-auth-cookies` and `/test/set-auth-cookies-with-secure` so all three cookies' `Max-Age` is asserted on both branches.
- **`tests/unit/utils/test_cookies.py:163` — `test_refreshed_token_has_max_age` `secure=False` gap.** Same parametrize pattern over `/test/update-token` and `/test/update-token-with-secure`.
- **`tests/unit/api/test_auth.py:216` — no integration test for `validateToken` through middleware.** Added `TestValidateAccessTokenIntegration` wiring `AuthMiddleware` + `auth_router` + `configure_exception_handlers` via TestClient. Authed Bearer → 200 + `authStatus=True`; unauth → 401 in Immich error shape (`statusCode`, `message`).

A round-2 self-review then trimmed the new fixture: dropped an unused `TEST_ENCRYPTED_JWT` constant (`stored_jwt` is set but `get_jwt` is mocked, so the encrypted value is dead) and added `spec=SessionStore` to the `AsyncMock` to match the sibling `TestPostLogout` fixture.

tests/unit/utils/test_cookies.py
taggartgorman: Fixed in `631d4c6` — parametrized over `/test/set-auth-cookies` and `/test/set-auth-cookies-with-secure` so the Max-Age assertion runs on both `secure=True` and `secure=False`.

tests/unit/utils/test_cookies.py
taggartgorman: Fixed in `631d4c6` — parametrized over `/test/update-token` and `/test/update-token-with-secure` so Max-Age is asserted on both `secure=True` and `secure=False` for `update_access_token_cookie`.

tests/unit/api/test_auth.py
taggartgorman: Fixed in `631d4c6` — added `TestValidateAccessTokenIntegration` in `tests/unit/api/test_auth.py` that wires `AuthMiddleware` + `auth_router` + `configure_exception_handlers` via TestClient. Authed Bearer → 200 + `authStatus=True`, unauth → 401 in Immich shape, matching the existing TestClient pattern in `test_cookies.py` / `test_auth_middleware.py`.

PR #151, opened 2026-05-04
GUM-688: Normalize asset width/height to display orientation

taggartgorman:
## Summary

Some assets in immich web were rendered stretched — most visibly portrait photos shown in landscape layout boxes. Root cause: the adapter emitted `asset.width`/`asset.height` in raw sensor orientation while Immich's wire contract expects display orientation (post-EXIF-rotation), and the immich web client's `getAssetRatio` helper consults width/height directly without applying the orientation tag.

This PR normalizes width/height to display orientation at every adapter boundary that emits them, and also nulls the wire `orientation` field whenever a swap occurred (matching upstream Immich and preventing immich web's companion `getDimensions` helper from re-applying the rotation on already-rotated dims).

- New `display_dimensions(width, height, orientation)` helper in `routers/utils/asset_conversion.py` swaps width/height for orientations 5–8 (90°/270° rotations).
- New `wire_orientation(orientation, width, height)` helper returns `None` whenever `display_dimensions` would swap.
- Applied at all seven emit sites: `AssetResponseDto`, `SyncAssetV1`, `ExifResponseDto`, `SyncAssetExifV1`, `build_asset_upload_ready_payload`, `gumnut_asset_to_sync_asset_v1`, `gumnut_metadata_to_sync_exif_v1`, and the timeline `ratio` field.
- Codebase convention documented in `docs/references/code-practices.md`.

Out of scope: `routers/api/faces.py` emits raw image dims paired with raw bounding-box coords — fixing that requires also rotating the bounding boxes and is a separate change.

## Linear

https://linear.app/gumnut-ai/issue/GUM-688/some-images-in-immich-web-are-not-displayed-correctly

## Test plan

- [x] Unit tests for `display_dimensions` and `wire_orientation` (parametrized over orientations 1–8 and None).
- [x] Per-call-site regression tests asserting both swapped dims and nulled orientation for orientation=6, and pass-through for orientation=1.
- [x] Timeline regression test asserting `ratio` is computed from display dims.
- [x] Full suite: 810 passing; ruff format/check and pyright clean.
- [ ] Manual verification: load an affected asset (`width > height` + `orientation in {5,6,7,8}`) in immich web; confirm the preview is no longer stretched.

### DIFF
--- a/tests/unit/api/sync/test_orientation.py
+++ b/tests/unit/api/sync/test_orientation.py
@@ -72,6 +72,9 @@ def test_sync_exif_v1_swaps_for_flipped_orientation():
 
     assert result.exifImageWidth == 2268
     assert result.exifImageHeight == 4032
+    # Orientation must be nulled when dims are swapped — see GUM-688
+    # (otherwise immich web's getDimensions re-applies the rotation).
+    assert result.orientation is None
 
 
 def test_sync_exif_v1_passes_through_unflipped_orientation():
@@ -86,3 +89,4 @@ def test_sync_exif_v1_passes_through_unflipped_orientation():
 
     assert result.exifImageWidth == 4032
     assert result.exifImageHeight == 2268
+    assert result.orientation == "1"
--- a/tests/unit/api/sync/test_orientation.py
+++ b/tests/unit/api/sync/test_orientation.py
@@ -44,8 +44,8 @@ def test_sync_asset_v1_passes_through_unflipped_orientations(orientation):
 
 @pytest.mark.parametrize("orientation", [5, 6, 7, 8])
 def test_sync_asset_v1_swaps_for_flipped_orientations(orientation):
-    """Regression: GUM-688 — Pixel-shot landscape buffers tagged orientation=6
-    were emitted unswapped, causing immich web to render portrait pixels into a
+    """Regression: Pixel-shot landscape buffers tagged orientation=6 were
+    emitted unswapped, causing immich web to render portrait pixels into a
     landscape layout box."""
     asset = create_mock_asset_data(UPDATED_AT)
     asset.width = 4032
@@ -72,8 +72,8 @@ def test_sync_exif_v1_swaps_for_flipped_orientation():
 
     assert result.exifImageWidth == 2268
     assert result.exifImageHeight == 4032
-    # Orientation must be nulled when dims are swapped — see GUM-688
-    # (otherwise immich web's getDimensions re-applies the rotation).
+    # Orientation must be nulled when dims are swapped, otherwise immich
+    # web's getDimensions re-applies the rotation on top of swapped dims.
     assert result.orientation is None
 
 
--- a/tests/unit/api/sync/test_orientation.py
+++ b/tests/unit/api/sync/test_orientation.py
@@ -1,0 +1,88 @@
+"""Tests for orientation-aware width/height normalization in sync converters.
+
+The sync stream's ``SyncAssetV1.width``/``height`` and
+``SyncAssetExifV1.exifImageWidth``/``exifImageHeight`` must reflect display
+orientation (post-rotation), matching upstream Immich's wire contract — clients
+read these directly without consulting the orientation tag.
+"""
+
+from datetime import datetime, timezone
+
+import pytest
+
+from routers.api.sync.converters import (
+    gumnut_asset_to_sync_asset_v1,
+    gumnut_metadata_to_sync_exif_v1,
+)
+from tests.unit.api.sync.conftest import (
+    create_mock_asset_data,
+    create_mock_metadata_data,
+)
+
+
+OWNER_UUID = "22222222-2222-2222-2222-222222222222"
+UPDATED_AT = datetime(2026, 5, 1, tzinfo=timezone.utc)
+
+
+@pytest.mark.parametrize("orientation", [None, 1, 2, 3, 4])
+def test_sync_asset_v1_passes_through_unflipped_orientations(orientation):
+    asset = create_mock_asset_data(UPDATED_AT)
+    asset.width = 4032
+    asset.height = 2268
+    if orientation is None:
+        asset.metadata = None
+    else:
+        metadata = create_mock_metadata_data(UPDATED_AT)
+        metadata.orientation = orientation
+        asset.metadata = metadata
+
+    result = gumnut_asset_to_sync_asset_v1(asset, owner_id=OWNER_UUID)
+
+    assert result.width == 4032
+    assert result.height == 2268
+
+
+@pytest.mark.parametrize("orientation", [5, 6, 7, 8])
+def test_sync_asset_v1_swaps_for_flipped_orientations(orientation):
+    """Regression: GUM-688 — Pixel-shot landscape buffers tagged orientation=6
+    were emitted unswapped, causing immich web to render portrait pixels into a
+    landscape layout box."""
+    asset = create_mock_asset_data(UPDATED_AT)
+    asset.width = 4032
+    asset.height = 2268
+    metadata = create_mock_metadata_data(UPDATED_AT)
+    metadata.orientation = orientation
+    asset.metadata = metadata
+
+    result = gumnut_asset_to_sync_asset_v1(asset, owner_id=OWNER_UUID)
+
+    assert result.width == 2268
+    assert result.height == 4032
+
+
+def test_sync_exif_v1_swaps_for_flipped_orientation():
+    asset = create_mock_asset_data(UPDATED_AT)
+    asset.width = 4032
+    asset.height = 2268
+    metadata = create_mock_metadata_data(UPDATED_AT)
+    metadata.orientation = 6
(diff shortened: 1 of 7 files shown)


taggartgorman (commented):
**Review verdict: ready to merge** (auto-approve path; cannot self-approve via API).

## Pre-approve checklist

- [x] **Findings**: 0 Must Fix / 0 Should Fix (4 Consider items below)
      Rationale: Fixes a visible rendering bug (portrait pixels stretched into landscape boxes for EXIF orientations 5–8). The two helpers (`display_dimensions`, `wire_orientation`) consolidate a non-obvious wire-contract rule across 7 emit sites and add a documented convention. Alternatives considered: (1) "do nothing" — leaves bug; (2) fix only `AssetResponseDto` — leaves the rest of the wire inconsistent and the `getDimensions` re-rotation case still wrong; (3) inline the swap at each site — duplicates the orientation-set membership check 7×; (4) put the swap in a single shared converter — would require restructuring the call graph. Helper-pair approach beats each.
      Evidence: Diff is limited to width/height/orientation field assignments and pure helpers. No changes to `auth*`, `current_user`, `get_authenticated_gumnut_client`, or any guard. Pre-existing `auth/trash` references in the same files are read-only and untouched.
      Evidence: No `delete`, `trash`, `permanent` paths touched. `deletedAt`/`trashed_at` references in same files are pre-existing pass-throughs and untouched.
      Rationale: Helpers are pure functions with parametrized tests over orientations 1–8 + None and missing dims. Each of the 7 emit sites has a regression test. CI green (lint, type, test, API compatibility). The metadata=None fallback case (Consider #1) preserves pre-PR behavior, not a regression.

## Consider-level (non-blocking) follow-ups

### 1. Silent fallback when `asset.metadata is None`
All five new normalization sites use `asset.metadata.orientation if asset.metadata else None`, and `display_dimensions(w, h, None)` returns raw dims unchanged. That fall-through is exactly the broken visual state. Concrete scenarios where metadata=None can reach these sites:
- `_emit_upload_events` (`routers/api/assets.py:278`) calls `build_asset_upload_ready_payload` immediately after upload, before metadata extraction has run.
- Sync stream's asset event fetcher (`routers/api/sync/entity_fetch.py:62`) does not filter on metadata presence (the metadata event fetcher at lines 83-98 does), so an asset event for an asset whose metadata extraction hasn't completed still ships SyncAssetV1 with raw dims. SyncAssetV1 has no orientation field; clients typically cache width/height off SyncAssetV1, not the later SyncAssetExifV1.

Whether this actually fires depends on the photos-api event-emission contract. Suggest a `logger.warning` (or a counter) at emission sites when `asset.metadata is None` but width/height are present, so the residual rate is observable in production.

### 2. Repeated `metadata.orientation` + `display_dimensions` block at 6 sites
The 2-line lookup
```python
orientation = asset.metadata.orientation if asset.metadata else None
width, height = display_dimensions(asset.width, asset.height, orientation)
```
appears at 6 sites. The `metadata.orientation if metadata else None` part is the most likely to drift if someone adds a new emit site and forgets the None-guard. Could be a thin helper:
```python
def asset_display_dims(asset: AssetResponse) -> tuple[int | None, int | None, int | None]:
    orientation = asset.metadata.orientation if asset.metadata else None
    w, h = display_dimensions(asset.width, asset.height, orientation)
    return w, h, orientation
```
Mid-grain copy-paste — keep or consolidate is a judgment call.

### 3. Duplicate `display_dimensions` call after `extract_exif_info` / `extract_sync_exif`
`convert_gumnut_asset_to_immich` (line 392) and `build_asset_upload_ready_payload` (line 311) both compute `display_dimensions(...)` on the outer DTO after calling a helper that already computed and stored the same values on the nested DTO (`exif_info.exifImageWidth/Height`, `sync_exif.exifImageWidth/Height`). Cheap to recompute, but two sources of truth that could silently diverge if the rule ever changes. Could pull the second pair from the already-extracted DTO (`width, height = exif_info.exifImageWidth, exif_info.exifImageHeight`); the int→float coercion at the use sites still works.

### 4. Doc subsection placement
The new `### Asset dimensions and orientation` subsection in `docs/references/code-practices.md` is inserted between `### Implementing New Endpoints` and `### Stub endpoints — fail closed on auth/authz checks`, breaking the endpoint-flow continuity of the surrounding subsections. Consider moving it before `### Implementing New Endpoints` or after `### Stub endpoints`.

ternarybits:
Does the Gumnut API contract make sense?  Should we change how we return orientation / width / height from photos-api?

ternarybits (commented):
Reviewed the display-orientation normalization change. The implementation is focused and the helper captures a real Immich wire-contract invariant; CI is green across compatibility, lint/typecheck, tests, and Immich version sync.

I found two small convention issues in newly added regression comments/docstrings. They are worth cleaning up because this is a public repo and the code-practices doc explicitly asks us to avoid internal tracker IDs in code comments.

tests/unit/api/sync/test_orientation.py
```diff
@@ -0,0 +1,92 @@
...
+
+    result = gumnut_asset_to_sync_asset_v1(asset, owner_id=OWNER_UUID)
+
+    assert result.width == 4032
+    assert result.height == 2268
+
+
+@pytest.mark.parametrize("orientation", [5, 6, 7, 8])
+def test_sync_asset_v1_swaps_for_flipped_orientations(orientation):
+    """Regression: GUM-688 — Pixel-shot landscape buffers tagged orientation=6
```
ternarybits: **[Should Fix]** Remove internal issue IDs from test comments

This new test docstring references `GUM-688`, and the same file repeats it in the comment below about nulled orientation. The repo's code practices explicitly say not to put internal tracker IDs in code comments because this is a public repository and not every reader has access. The surrounding prose already explains the regression, so the ID adds inaccessible context without adding useful information.

Please remove the `GUM-688` references and keep the self-contained explanation of the orientation/dimension mismatch.

tests/unit/api/test_timeline.py
```diff
@@ -440,6 +440,42 @@ async def test_get_time_bucket_success(
                 extra_query=JANUARY_2024_DATE_RANGE
             )
 
+    @pytest.mark.anyio
+    async def test_get_time_bucket_ratio_uses_display_orientation(
+        self, multiple_gumnut_assets, mock_sync_cursor_page
+    ):
+        """Regression: GUM-688 — the timeline's per-asset ``ratio`` must be
```
ternarybits: **[Should Fix]** Remove internal issue ID from regression docstring

This new test docstring references `GUM-688`, which violates the repo convention against internal tracker IDs in code comments for this public repository. The rest of the docstring already describes the concrete rendering failure and reproduction shape, so the private ID is unnecessary.

Please drop `GUM-688` from the docstring and leave the regression context in plain, self-contained terms.

taggartgorman:
> Does the Gumnut API contract make sense? Should we change how we return orientation / width / height from photos-api?

@ternarybits Upon review, yes it makes sense to normalize dimensions upon ingest. I have created [GUM-720](https://linear.app/gumnut-ai/issue/GUM-720/photos-api-normalize-asset-widthheight-to-display-orientation-at) to track this.

This fix in this PR does not interfere with the suggested GUM-720 fix, nor will it be broken by the GUM-720 fix.

taggartgorman:
## Review feedback addressed (round 1)

### Auto-fixed (two-way door)
- **Removed internal tracker IDs from test docstrings/comments** — dropped the IDs from the new test docstrings in `tests/unit/api/sync/test_orientation.py` (regression docstring + the swapped-orientation comment) and `tests/unit/api/test_timeline.py` (timeline ratio regression docstring). The surrounding prose already explains the regression in self-contained terms.

### Documentation
- **Tightened the convention** in `docs/references/code-practices.md` so the existing "no internal tracker IDs in code comments" rule explicitly covers docstrings (and test docstrings) — the lapse here suggests the prior wording read as if docstrings were out of scope.

### Skipped (non-blocking, author-flagged)
The four "Consider"-level items in the original self-review (silent fallback when `asset.metadata is None`, repeated `metadata.orientation` + `display_dimensions` block, duplicate `display_dimensions` after the extract helpers, doc subsection placement) are intentionally not addressed in this round — they were posted as non-blocking follow-ups and remain judgment calls. Happy to revisit any of them in a follow-up if reviewers want them landed here.

CI is green. Re-requesting review.

tests/unit/api/sync/test_orientation.py
taggartgorman: Fixed — removed the tracker IDs from both the regression docstring and the swapped-orientation comment in this file. Replaced the comment with the self-contained explanation of why orientation must be nulled when dims are swapped (immich web getDimensions re-applies rotation otherwise).

tests/unit/api/test_timeline.py
taggartgorman: Fixed — removed the tracker ID from the regression docstring; the surrounding prose already describes the concrete failure mode.

ternarybits (approved):

PR #152, opened 2026-05-05
GUM-724: Implement read-only "On this day" memories

taggartgorman:
## Summary

- Replaces the `routers/api/memories.py` read-side stubs with a real implementation: `GET /api/memories`, `GET /api/memories/{id}`, and `GET /api/memories/statistics` synthesize OnThisDay memories at request time from per-day asset queries against the photos-api. Mutation endpoints stay no-op stubs (out of scope).
- Memory IDs are deterministic UUID-shaped tokens encoding `(marker, year, month, day, user_id_low_8_bytes)` so `GET /memories/{id}` decodes in O(1) without a UUIDv5 hash scan, and IDs minted for one user 404 for any other user.
- Flips `memories.enabled` to `True` in the user preferences stub — the Immich web carousel reads `$preferences?.memories?.enabled` and stays hidden otherwise.

## Implementation notes

- 30-year window, fanned out via `asyncio.gather`. Skips empty years before responding so the client never has to.
- Per-day query uses naive `local_datetime_after`/`local_datetime_before` so the photos-api compares against each asset's wall-clock capture time directly (matching how the user thinks about "the photos from May 4, 2024").
- The Immich web client encodes "today" by stamping its local wall-clock time as UTC (`setZone('utc', { keepLocalTime: true })`) — pulling `.year/.month/.day` off the parsed datetime as-is is correct; no tz math required at the adapter layer. Documented as a learning in `docs/references/code-practices.md`.
- Self-review caught one real performance bug: `async for asset in client.assets.list(..., limit=N)` does **not** cap iteration at N — `limit` is the SDK's per-page size and `async for` walks every page. Without an explicit `break`, `/statistics` (which uses `limit=1` to probe non-empty years) would burn one round-trip per matching asset. Fixed and documented.
- Mutation endpoints (`POST/PUT/DELETE /memories[/{id}[/assets]]`) intentionally remain no-op stubs. The full memory viewer's save/hide/remove actions will appear to succeed but won't persist; acceptable for the carousel-only goal. Adding write persistence is the only piece left, and only matters if save/hide UX is requested.

## Linear

https://linear.app/gumnut-ai/issue/GUM-724/immich-adapter-implement-read-only-on-this-day-memories

## Test plan

- [x] `uv run pytest` — 830 passed (20 in `tests/unit/api/test_memories.py`)
- [x] Hit `GET /api/memories?for=<today-local>` for a user with assets across multiple past years; one entry per non-empty year.
- [x] Open the Immich web client photos page; confirm the "On this day" carousel renders thumbnails with "N years ago" titles.
- [x] Click a carousel item; confirm `/memory/...` route loads.

### DIFF
--- a/docs/design-docs/immich-adapter-gap-analysis.md
+++ b/docs/design-docs/immich-adapter-gap-analysis.md
@@ -196,7 +196,7 @@ Write endpoints (`POST /memories`, `PUT /memories/{id}`, `DELETE /memories/{id}`
 
 **Effort**: **S remaining** — Adding write persistence is the only piece left, and only if save/hide UX matters before another consumer needs durable memories.
 
-**Recommendation**: **Revisit if/when save/hide UX is requested** — the read path covers the carousel.
+**Recommendation**: **Revisit later** — the read path covers the carousel; revisit if/when save/hide UX is requested.
 
 ---
 
@@ -656,7 +656,7 @@ These are architectural limitations documented in `docs/architecture/adapter-arc
 |-----|--------|------------|-----------|
 | #3a Map markers | S–M | Both | Good value if EXIF GPS data exists in backend |
 | #34 Performance (pagination) | L | Both | Scaling requirement |
-| #8 Memories (write path) | S | Adapter or backend | Read path shipped; only save/hide persistence remains |
+| #8 Memories (write path) | S | Both | Read path shipped; only save/hide persistence remains |
 | #2 Tags | L | Both | Power-user organization |
 
 ### Tier 3: Revisit Later (high effort or blocked)
--- a/docs/design-docs/immich-adapter-gap-analysis.md
+++ b/docs/design-docs/immich-adapter-gap-analysis.md
@@ -184,15 +184,17 @@ Immich activities allow users to add comments and reactions to shared albums.
 
 Immich's "memories" feature auto-generates "On This Day" collections and similar nostalgia-based groupings.
 
-**Current behavior**: All 8 endpoints return empty lists. The memories section in the Immich UI is empty.
+**Current behavior**: Read endpoints (`GET /memories`, `GET /memories/{id}`, `GET /memories/statistics`) synthesize OnThisDay memories at request time by querying the photos-api for assets captured on today's local month/day across the previous 30 years (one parallel call per year). Memory IDs encode `(user, year, month, day)` so they round-trip without persistence. The Immich web "On This Day" carousel renders correctly.
 
-**User impact**: **Medium** — "On This Day" is a popular engagement feature in photo apps. Users opening the app expect to see memories if they have photos from previous years.
+Write endpoints (`POST /memories`, `PUT /memories/{id}`, `DELETE /memories/{id}`, `POST/DELETE /memories/{id}/assets`) remain no-op stubs. The memory viewer's save/hide/remove actions appear to succeed but don't persist; this is acceptable for the carousel-only goal and is flagged in `routers/api/memories.py`.
 
-**Dependency**: **Both** — Gumnut Photos API would need a memory generation system (likely a background job that queries assets by date). Adapter translation is straightforward.
+**User impact**: **Resolved for the read path** — the engagement feature is now visible. Persistence for save/hide is a follow-up.
 
-**Effort**: **L** — Backend needs date-based query logic, memory entity CRUD, and a generation mechanism (could be a Celery task). Adapter adds S overhead.
+**Dependency**: **Adapter only** for read; write would need a backend persistence layer (or an adapter-side store).
 
-**Recommendation**: **Revisit later** — Nice engagement feature but requires non-trivial backend work.
+**Effort**: **S remaining** — Adding write persistence is the only piece left, and only if save/hide UX matters before another consumer needs durable memories.
+
+**Recommendation**: **Revisit if/when save/hide UX is requested** — the read path covers the carousel.
 
 ---
 
@@ -652,7 +654,7 @@ These are architectural limitations documented in `docs/architecture/adapter-arc
 |-----|--------|------------|-----------|
 | #3a Map markers | S–M | Both | Good value if EXIF GPS data exists in backend |
 | #34 Performance (pagination) | L | Both | Scaling requirement |
-| #8 Memories | L | Both | Engagement feature |
+| #8 Memories (write path) | S | Adapter or backend | Read path shipped; only save/hide persistence remains |
 | #2 Tags | L | Both | Power-user organization |
 
 ### Tier 3: Revisit Later (high effort or blocked)
--- a/docs/design-docs/immich-adapter-gap-analysis.md
+++ b/docs/design-docs/immich-adapter-gap-analysis.md
@@ -184,7 +184,9 @@ Immich activities allow users to add comments and reactions to shared albums.
 
 Immich's "memories" feature auto-generates "On This Day" collections and similar nostalgia-based groupings.
 
-**Current behavior**: Read endpoints (`GET /memories`, `GET /memories/{id}`, `GET /memories/statistics`) synthesize OnThisDay memories at request time by querying the photos-api for assets captured on today's local month/day across the previous 30 years (one parallel call per year). Memory IDs encode `(user, year, month, day)` so they round-trip without persistence. The Immich web "On This Day" carousel renders correctly.
+**Current behavior**: Read endpoints `GET /memories` and `GET /memories/{id}` synthesize OnThisDay memories at request time by querying the photos-api for assets captured on today's local month/day across the previous 30 years (one parallel call per year). Memory IDs encode `(user, year, month, day)` so they round-trip without persistence. The Immich web "On This Day" carousel renders correctly.
+
+`GET /memories/statistics` remains a `total=0` stub. No upstream Immich client (web or mobile) calls it, so synthesizing a real count would burn round-trips for a value nobody reads.
 
 Write endpoints (`POST /memories`, `PUT /memories/{id}`, `DELETE /memories/{id}`, `POST/DELETE /memories/{id}/assets`) remain no-op stubs. The memory viewer's save/hide/remove actions appear to succeed but don't persist; this is acceptable for the carousel-only goal and is flagged in `routers/api/memories.py`.
 
--- a/routers/api/memories.py
+++ b/routers/api/memories.py
@@ -80,8 +80,8 @@ def decode_memory_id(memory_id: UUID, user_id: UUID) -> tuple[int, int, int] | N
     return year, month, day
 
 
-def _local_today(for_param: datetime | None) -> tuple[int, int]:
-    """Return today's (month, day) in the user's local timezone.
+def _local_today(for_param: datetime | None) -> tuple[int, int, int]:
+    """Return today's (year, month, day) in the user's local timezone.
 
     The Immich web client encodes "today" by taking its local wall-clock time
     and pretending it's UTC (`setZone('utc', { keepLocalTime: true })`). The
@@ -89,13 +89,19 @@ def _local_today(for_param: datetime | None) -> tuple[int, int]:
     pulling them off the parsed datetime as-is gives us today in their tz —
     no offset arithmetic needed.
 
+    The year matters for the search window: a Sydney user just past midnight
+    on Jan 1 sees `for` carrying the new year while the server's UTC clock
+    still reads Dec 31 of the prior year. Threading the local year through
+    keeps the window aligned with what the user would call "the past 30
+    years" instead of slicing off the year just ended.
+
     When the client omits `for`, fall back to today UTC; the carousel always
     sends `for`, so this branch is only hit by direct API consumers.
     """
     if for_param is None:
         today = datetime.now(tz=timezone.utc)
-        return today.month, today.day
-    return for_param.month, for_param.day
+        return today.year, today.month, today.day
+    return for_param.year, for_param.month, for_param.day
 
 
 def _year_window(reference_year: int) -> list[int]:
@@ -121,20 +127,26 @@ async def _fetch_assets_for_day(
     Note: SDK `local_datetime_after` is exclusive, so passing the exact
     midnight boundary skips assets captured at exactly 00:00:00.000000 — we
     accept that microsecond edge case for symmetry with `timeline.py`.
+
+    The SDK's `limit` parameter is the per-page size, not a result cap —
+    `async for` walks every page until `has_more` is false. We break out
+    explicitly so callers actually receive at most `limit` assets; without
+    this, `/statistics` (which only needs to know each year is non-empty)
+    would burn a round-trip per asset on busy days.
     """
-    try:
-        day_start = datetime(year, month, day)
-    except ValueError:
-        return []
+    day_start = datetime(year, month, day)
     day_end = day_start + timedelta(days=1)
-    return [
-        asset
-        async for asset in client.assets.list(
-            local_datetime_after=day_start.isoformat(),
-            local_datetime_before=day_end.isoformat(),
-            limit=limit,
-        )
-    ]
+    assets: list[AssetResponse] = []
+    async for asset in client.assets.list(
+        local_datetime_after=day_start.isoformat(),
+        local_datetime_before=day_end.isoformat(),
+        state="live",
+        limit=limit,
+    ):
+        assets.append(asset)
+        if len(assets) >= limit:
+            break
+    return assets
 
 
 def _build_memory(
@@ -212,8 +224,8 @@ async def search_memories(
     ):
         return []
 
-    month, day = _local_today(for_param)
-    years = _year_window(datetime.now(tz=timezone.utc).year)
+    reference_year, month, day = _local_today(for_param)
+    years = _year_window(reference_year)
     year_assets = await _gather_year_assets(
         client, years, month, day, _ASSETS_PER_MEMORY
     )
@@ -260,8 +272,8 @@ async def memories_statistics(
     ):
         return MemoryStatisticsResponseDto(total=0)
 
-    month, day = _local_today(for_param)
-    years = _year_window(datetime.now(tz=timezone.utc).year)
+    reference_year, month, day = _local_today(for_param)
+    years = _year_window(reference_year)
     # Cap at 1 per year — we only need to know whether each year is non-empty.
     year_assets = await _gather_year_assets(client, years, month, day, limit=1)
     return MemoryStatisticsResponseDto(
--- a/routers/api/memories.py
+++ b/routers/api/memories.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 from datetime import date, datetime, timedelta, timezone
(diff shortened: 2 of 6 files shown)


routers/api/memories.py
```diff
@@ -24,19 +31,210 @@
...
+    midnight boundary skips assets captured at exactly 00:00:00.000000 — we
+    accept that microsecond edge case for symmetry with `timeline.py`.
+
+    The SDK's `limit` parameter is the per-page size, not a result cap —
+    `async for` walks every page until `has_more` is false. We break out
+    explicitly so callers actually receive at most `limit` assets; without
+    this, `/statistics` (which only needs to know each year is non-empty)
+    would burn a round-trip per asset on busy days.
+    """
+    day_start = datetime(year, month, day)
```
taggartgorman: **[Should Fix] Feb 29 deterministically crashes the whole window.**

When `for_param` lands on Feb 29 (e.g. 2024-02-29), `_local_today` returns `(2024, 2, 29)` and `_year_window(2024)` produces `[2023, 2022, ..., 1994]`. For every non-leap year in the window, `datetime(year, 2, 29)` here raises `ValueError: day is out of range for month`. With `asyncio.gather` defaulting to fail-fast, the first ValueError tanks the whole `/memories` (and `/statistics`) call → 500.

Fires deterministically every Feb 29 (~once every 4 years) for every user opening the carousel that day; the user expects to see leap-year memories from 2020, 2016, etc. and instead gets an error.

Suggestion: either skip non-leap years in `_year_window` when `(month, day) == (2, 29)`, or wrap the `datetime(year, month, day)` construction here in `try`/`except ValueError: return []`. Add a regression test for `for_param=2024-02-29`.

taggartgorman (commented):
Solid PR — read-only OnThisDay carousel works end-to-end, the synthetic-ID codec is well thought out (deterministic + cross-user 404 without a backend round-trip), and the two non-obvious learnings landed in `code-practices.md` (the `keepLocalTime` wire-format hack and the SDK `limit`-is-per-page quirk) are load-bearing. Tests cover the main shapes well.

Two `Should Fix` items worth addressing before merge:

1. **Feb 29 deterministically crashes the whole window.** `_year_window` produces non-leap years where `datetime(year, 2, 29)` raises, and `asyncio.gather`'s fail-fast propagates → 500 once every 4 years.
2. **`asyncio.gather` without `return_exceptions=True` makes any single transient backend failure tank all 30 years.** Memories are a soft-feature surface; partial results are friendlier than a 500.

The remaining items are `Consider`-level — comment density (some doc paragraphs now duplicate `code-practices.md`), test coverage gaps, and a fan-out concurrency-cap suggestion.

routers/api/memories.py
```diff
@@ -24,19 +31,210 @@
...
+
+async def _gather_year_assets(
+    client: AsyncGumnut,
+    years: list[int],
+    month: int,
+    day: int,
+    limit: int,
+) -> list[tuple[int, list[AssetResponse]]]:
+    """Fan out per-year `assets.list` calls and pair each result with its year."""
+    asset_lists = await asyncio.gather(
```
taggartgorman: **[Should Fix] `asyncio.gather` fail-fast tanks the whole window on one transient error.**

`asyncio.gather(*coros)` defaults to `return_exceptions=False`, so a single failing year's `client.assets.list(...)` (transient backend timeout, network blip, rate-limit reject) propagates and cancels the other 29 in-flight requests. The user gets a 500 instead of 29-out-of-30 years' memories. With 30 parallel calls, the probability that any one transiently fails is materially higher than for a single call. Compounds with the Feb 29 finding above.

Suggestion: pass `return_exceptions=True` and either log+drop failed years or substitute empty lists. Memories are a soft-feature surface — the carousel renders best-effort partial results gracefully, but a 500 hides everything.

routers/api/memories.py
```diff
@@ -24,19 +31,210 @@
...
+    if is_saved is True:
+        return True
+    if is_trashed is True:
+        return True
+    if memory_type is not None and memory_type != MemoryType.on_this_day:
+        return True
+    return False
+
+
+async def _gather_year_assets(
```
taggartgorman, line 189: **[Consider] 30-way fan-out has no concurrency cap.**

Both `/memories` and `/statistics` issue 30 simultaneous `client.assets.list(...)` calls with no `asyncio.Semaphore`. The carousel polls these when the timeline opens, so a fleet of clients produces a 30x backend amplification factor. There's no per-user rate limit visible at the adapter layer.

Suggestion: wrap with a small `asyncio.Semaphore(6)` (or similar) shared across the gather to bound photos-api fan-out. Trade-off: serialized portion within the cap adds latency. Worth weighing — the adapter doesn't have precedent for concurrency caps elsewhere, but no other endpoint fans out to this degree either.

routers/api/memories.py
```diff
@@ -24,19 +31,210 @@
...
+    is_trashed: bool | None,
+    memory_type: MemoryType | None,
+) -> bool:
+    """True when the requested filters select for memories we never synthesize.
+
+    Synthetic OnThisDay memories are always live (not trashed), unsaved, and of
+    type `on_this_day`. Any filter that excludes those guarantees an empty
+    result, so we can short-circuit before fanning out asset queries.
+    """
+    if is_saved is True:
```
taggartgorman, line 180: **[Consider] Filter false-path branches (`isSaved=False`, `isTrashed=False`) untested.**

`_filters_exclude_synthetic` short-circuits only when the boolean is explicitly `True`; `False` and `None` both pass through. The carousel sends `isSaved=false` regularly. A regression that flipped the predicate to `is_saved is not None` would silently break the main carousel call but no current test would fail.

Suggestion: add tests passing `isSaved=False` and `isTrashed=False` and assert the fan-out still happens (memories are returned, not short-circuited).

tests/unit/api/test_memories.py
```diff
@@ -0,0 +1,446 @@
...
+        # Every per-year call should target July 15.
+        assert client.assets.list.call_count == _YEAR_WINDOW
+        for call in client.assets.list.call_args_list:
+            assert call.kwargs["local_datetime_after"].endswith("-07-15T00:00:00")
+            assert call.kwargs["local_datetime_before"].endswith("-07-16T00:00:00")
+
+
+class TestFetchAssetsForDay:
+    @pytest.mark.anyio
+    async def test_caps_at_limit_across_pages(self):
```
taggartgorman, line 423: **[Consider] `test_caps_at_limit_across_pages` doesn't actually exercise pagination.**

The test name promises validation "across pages," but `MockSyncCursorPage.__aiter__` yields all items from a single in-memory list — no real page boundaries. The `break`-when-`len>=limit` logic works the same on a flat list as on a paged stream, so the test passes either way. The bug it would catch (someone removing the break and relying on SDK `limit` as a result cap) is real, but the in-memory iterator can't distinguish per-page-size vs result-cap semantics.

Suggestion: either rename to `test_caps_at_limit_in_iteration` to match what's verified, or extend `MockSyncCursorPage` to simulate multi-page iteration with a `has_more` transition so a regression dropping the explicit break would actually fail.

tests/unit/api/test_memories.py
```diff
@@ -0,0 +1,446 @@
...
+        assert decode_memory_id(uuid4(), USER_UUID) is None
+
+    def test_invalid_date_returns_none(self):
+        # Manually build a UUID with our marker but month=13.
+        marker = b"OTD\x00"
+        raw = marker + (2024).to_bytes(2, "big") + bytes([13, 1]) + USER_UUID.bytes[8:]
+        assert decode_memory_id(UUID(bytes=raw), USER_UUID) is None
+
+
+class TestSearchMemories:
```
taggartgorman, line 117: **[Consider] No test for one year's `assets.list` raising.**

No regression guard for the most likely production failure mode: a transient backend error in one of the 30 fan-out calls. Pairs with the `return_exceptions=True` finding — pin whichever behavior is chosen.

Suggestion: add a test that stubs `client.assets.list` to raise on one specific year and asserts the chosen behavior (currently: 500 propagates; after the gather fix: degraded result).

routers/api/memories.py
```diff
@@ -24,19 +31,210 @@
...
+    if not (1 <= month <= 12 and 1 <= day <= 31):
+        return None
+    try:
+        date(year, month, day)
+    except ValueError:
+        return None
+    return year, month, day
+
+
+def _local_today(for_param: datetime | None) -> tuple[int, int, int]:
```
taggartgorman, line 82: **[Consider] Docstring restates `code-practices.md`.**

The `keepLocalTime` wire-format derivation and Sydney-midnight worked example (paragraphs 1-2) now live in `docs/references/code-practices.md` under "Immich web 'today' wire format" (added in this same PR). The function name plus a one-line pointer would carry the load. Only the `for_param is None` UTC-fallback paragraph is unique to this site.

Suggestion: trim to a 1-2 line docstring naming the convention and pointing at the reference; keep only the UTC-fallback note.

routers/api/memories.py
```diff
@@ -24,19 +31,210 @@
...
+    Uses naive `local_datetime` bounds so the photos-api compares against each
+    asset's wall-clock capture time directly (matching what the user would
+    intuitively call "the photos from May 4, 2024"), regardless of the device
+    timezone the photo was originally captured in.
+
+    Note: SDK `local_datetime_after` is exclusive, so passing the exact
+    midnight boundary skips assets captured at exactly 00:00:00.000000 — we
+    accept that microsecond edge case for symmetry with `timeline.py`.
+
+    The SDK's `limit` parameter is the per-page size, not a result cap —
```
taggartgorman: **[Consider] SDK-`limit` paragraph duplicates `code-practices.md`.**

The per-page-vs-result-cap explanation here mirrors what's now in `docs/references/code-practices.md` "Counts and Aggregates" (the `assets: list[AssetResponse] = []` snippet, added in this PR). The break-out pattern below is exactly what the convention prescribes.

Suggestion: replace the SDK-limit paragraph with a one-liner pointing at the reference. Keep the `local_datetime_after`-is-exclusive note — that's local to this implementation choice.

docs/design-docs/immich-adapter-gap-analysis.md
```diff
@@ -184,15 +184,17 @@ Immich activities allow users to add comments and reactions to shared albums.
...
-**Dependency**: **Both** — Gumnut Photos API would need a memory generation system (likely a background job that queries assets by date). Adapter translation is straightforward.
+**User impact**: **Resolved for the read path** — the engagement feature is now visible. Persistence for save/hide is a follow-up.
 
-**Effort**: **L** — Backend needs date-based query logic, memory entity CRUD, and a generation mechanism (could be a Celery task). Adapter adds S overhead.
+**Dependency**: **Adapter only** for read; write would need a backend persistence layer (or an adapter-side store).
 
-**Recommendation**: **Revisit later** — Nice engagement feature but requires non-trivial backend work.
+**Effort**: **S remaining** — Adding write persistence is the only piece left, and only if save/hide UX matters before another consumer needs durable memories.
+
+**Recommendation**: **Revisit if/when save/hide UX is requested** — the read path covers the carousel.
```
taggartgorman: **[Consider] Vocabulary drift in section #8.**

Line 44 documents the fixed Recommendation vocabulary as `Close / Intentional gap / Revisit later`. This entry uses `**Revisit if/when save/hide UX is requested**`, which doesn't match — readers scanning the doc by section status will skim past it. The Tier 2 row at line 657 also uses `Adapter or backend` for Dependency, where the rest of the doc uses `Adapter-only` / `Both` / `N/A`.

Suggestion: use `**Revisit later**` for the Recommendation (move the condition into the bullet body), and pick `Both` (or `Adapter-only`) for the Dependency cell.

routers/api/memories.py
```diff
@@ -24,19 +31,210 @@
 )
 
 
+# Memory IDs are synthesized rather than persisted. The first 4 bytes of the
+# UUID act as a marker so we can recognize and decode our own IDs; the next 4
+# bytes encode the (year, month, day) the memory points at; the last 8 bytes
+# bind the ID to the user that generated it (low half of the user UUID), so a
+# memory ID minted for user A returns 404 when fetched by user B.
```
taggartgorman: **[Consider] File-top comment overlaps the encode/decode docstrings.**

Lines 34-38 explain the marker layout and the user-binding 404 behavior; `encode_memory_id` and `decode_memory_id` cover the same ground in their docstrings. Either the file-top block or the docstrings can shrink.

Suggestion: trim the marker block to a 1-line `# 4-byte marker so we recognize our own synthetic IDs.` Rely on the codec docstrings for layout details. Keep the `_YEAR_WINDOW` and `_ASSETS_PER_MEMORY` rationale comments — those are non-obvious choices.

taggartgorman:
## Round 2: self-review feedback addressed

### Fixed (auto, two-way door)

- **[Should Fix] Feb 29 crash** — `_fetch_assets_for_day` now wraps `datetime(year, month, day)` in try/except returning `[]`, so non-leap years in the window don't ValueError. Added regression test `test_feb_29_does_not_crash_window`.
- **[Should Fix] `asyncio.gather` fail-fast** — `_gather_year_assets` now passes `return_exceptions=True`. Per-year exceptions log at WARNING with `exc_info` and substitute `[]`. Filter on `Exception` (not `BaseException`) so `asyncio.CancelledError` propagates instead of being swallowed; non-Exception `BaseException` results are explicitly re-raised. Pinned by `test_one_year_raising_yields_degraded_result`.
- **[Consider] Filter False/None tests** — Parametrized `test_is_saved_false_or_none_does_not_short_circuit` and `test_is_trashed_false_or_none_does_not_short_circuit`.
- **[Consider] Pagination test mock** — Replaced flat `MockSyncCursorPage` usage with a local `_PaginatedListing` that tracks `pages_fetched` across page boundaries, so dropping the explicit `break` would walk extra pages and fail.
- **[Consider] Trim `_local_today` docstring** — Now points at `code-practices.md` "Immich web 'today' wire format" instead of restating it.
- **[Consider] Trim `_fetch_assets_for_day` SDK-limit paragraph** — Now points at `code-practices.md` "Counts and Aggregates"; kept the `local_datetime_after`-exclusive note (local choice).
- **[Consider] Trim file-top comment** — Codec docstrings already cover layout; trimmed to a one-liner pointing at `encode_memory_id`.
- **[Consider] Vocabulary drift in gap-analysis #8** — Recommendation now uses standard "Revisit later"; Tier 2 row Dependency uses "Both" instead of "Adapter or backend".

### Skipped (with rationale)

- **[Consider] No concurrency cap on 30-way fan-out** — Skipping. The reviewer themselves flagged the tradeoff ("the adapter doesn't have precedent for concurrency caps elsewhere, but no other endpoint fans out to this degree either"); there's no demonstrated production load problem; the carousel polls once per timeline open. Easy to add if metrics ever show the amplification matters.

### Documentation learnings

Added two `code-practices.md` entries based on this round:
- New "Parallel Fan-Out with `asyncio.gather`" subsection covering the `return_exceptions=True` pattern and the `Exception` vs `BaseException` gotcha.
- Testing bullet noting that `MockSyncCursorPage` can't distinguish per-page-size from result-cap; reference the local `_PaginatedListing` for the canonical shape.

Tests: 834 pass (24 in `test_memories.py`, +6 from this round). Lint, format, pyright clean.

routers/api/memories.py
taggartgorman: Fixed in 856b7bf — `_fetch_assets_for_day` now catches `ValueError` from `datetime(year, 2, 29)` and returns `[]`. Regression test: `test_feb_29_does_not_crash_window`.

routers/api/memories.py
taggartgorman: Fixed in 856b7bf — `_gather_year_assets` now passes `return_exceptions=True`, logs per-year failures at WARNING with `exc_info`, and substitutes `[]` for failed years. Filter on `Exception` (not `BaseException`) so `CancelledError` propagates; non-Exception BaseExceptions are re-raised. Pinned by `test_one_year_raising_yields_degraded_result`.

routers/api/memories.py
taggartgorman: Skipping. The tradeoff you flagged in the suggestion (no precedent for concurrency caps in the adapter; serialized portion adds latency) is the deciding factor: there is no demonstrated production load problem today, and the carousel only polls once per timeline open. Easy to add later if metrics show the 30x amplification matters.

routers/api/memories.py
taggartgorman: Fixed in 856b7bf — added parametrized tests `test_is_saved_false_or_none_does_not_short_circuit[False|None]` and `test_is_trashed_false_or_none_does_not_short_circuit[False|None]` asserting `client.assets.list.call_count == _YEAR_WINDOW` and a memory is returned. A predicate inversion to `is not None` would now fail four cases.

tests/unit/api/test_memories.py
taggartgorman: Extended in 856b7bf — built a local `_PaginatedListing` mock that increments a `pages_fetched` counter at each page boundary. The test now asserts `pages_fetched == 1` after capping at `limit=2` over 5 items with `page_size=2`; dropping the explicit `break` would walk into page 2 (and 3) and fail the assertion. Test renamed-in-spirit by tightening what it checks rather than what its name says.

tests/unit/api/test_memories.py
taggartgorman: Fixed in 856b7bf — added `test_one_year_raising_yields_degraded_result`: stubs `client.assets.list` to raise on year 2024 and asserts other years still return memories. Pins the `return_exceptions=True` behavior chosen above.

routers/api/memories.py
taggartgorman: Trimmed in 856b7bf — `_local_today` docstring is now 2 sentences pointing at `code-practices.md` "Immich web today wire format" + the `for_param is None` UTC-fallback note.

routers/api/memories.py
taggartgorman: Trimmed in 856b7bf — replaced the SDK-limit paragraph with a one-line pointer to `code-practices.md` "Counts and Aggregates". Kept the `local_datetime_after`-is-exclusive note (local implementation choice). Also added a paragraph documenting the new Feb 29 ValueError guard.

docs/design-docs/immich-adapter-gap-analysis.md
taggartgorman: Fixed in 856b7bf — Recommendation now reads "**Revisit later** — the read path covers the carousel; revisit if/when save/hide UX is requested" (the condition moved into the bullet body). Tier 2 row Dependency cell changed from "Adapter or backend" to "Both".

routers/api/memories.py
taggartgorman: Trimmed in 856b7bf — file-top now reads `# Synthesized (not persisted) memory IDs — see encode_memory_id for layout.` Codec docstrings carry the marker/year/month/day/user-low details. Kept the `_YEAR_WINDOW` and `_ASSETS_PER_MEMORY` rationale comments.

ternarybits (approved):

PR #154, opened 2026-05-05
GUM-670: chunk asset_ids on album add/remove

ternarybits:
## Summary
- Wrap the bulk SDK call in `add_assets_to_album` and `remove_asset_from_album` (`PUT`/`DELETE /api/albums/{id}/assets`) with chunking so a >cap request from the Immich web client can't 422 the entire batch against photos-api's `MAX_BULK_GET_IDS=100`.
- Per-chunk error isolation: a chunk-level `APIStatusError`/`GumnutError` only flips its own ids' results; other chunks still report `success`/`duplicate`/`not_found`. Sequential per the issue's "likely answer" — parallelization is GUM-668's scope.
- Extract a `chunked_per_item_bulk` async generator in `routers/utils/bulk.py` that owns the chunking + error classification + log shape (`chunk_size`, `request_size`); both album endpoints now share it. The trash-shape (errors propagate, no per-item mapping) is unaffected. Future Immich `BulkIdResponseDto` endpoints (tags, memories, duplicates stubs) get the pattern by-import.
- Bump `gumnut-sdk>=0.82.0` to consume the new `not_found_assets` field on the add response (photos-api #577 + SDK 0.82.0). `add_assets_to_album` maps `response.not_found_assets` → `Error1.not_found` per id; the contract-drift `unknown` warning fallback stays for ids absent from all three buckets.
- 9 endpoint tests covering chunking, input-order preservation, empty input, per-chunk error isolation, cross-chunk merging for both `duplicate` and `not_found`, and the success-with-not-found-bucket path; 7 helper tests covering the helper's own contract (chunking, error classification, transport-error log shape, mixed success/failure ordering).
- Document the new shape in `docs/references/code-practices.md` § Bulk-ID Endpoints, pointing at the helper.

## Linear
https://linear.app/gumnut-ai/issue/GUM-670/immich-adapter-chunk-asset-ids-on-album-addremove

## Test plan
- [x] \`uv run ruff format && uv run ruff check && uv run pyright\` pass
- [x] \`uv run pytest\` — 852 passed (9 endpoint tests, 7 helper tests, plus the rest of the suite)
- [ ] Manually exercise via the Immich web client adding >100 photos to an album end-to-end (deferred — staging verification)


routers/api/albums.py
```diff
@@ -269,34 +272,36 @@ async def remove_asset_from_album(
...
-            exc=bulk_error,
-            extra={"album_id": str(id), "asset_count": len(request.ids)},
-        )
-        return [
-            BulkIdResponseDto(id=str(uuid), success=False, error=Error1.unknown)
-            for uuid in request.ids
-        ]
 
-    return [BulkIdResponseDto(id=str(uuid), success=True) for uuid in request.ids]
+    # Upstream silently skips missing assets and 204s on success.
```
ternarybits: I think this has changed in the latest API/SDK.

ternarybits:
## Round 1 of PR feedback

**Auto-fixed (two-way door)**
- Dropped the `logger` parameter from `chunked_per_item_bulk`. The helper now uses a module-level `logger = logging.getLogger(__name__)` in `routers/utils/bulk.py` — standard Python convention and matches what `routers/api/albums.py` and `routers/utils/error_mapping.py` already do. Both call sites + the seven helper tests are updated.

**Skipped**
- _"I think this has changed in the latest API/SDK"_ on the `# Upstream silently skips missing assets and 204s on success.` comment. Verified against gumnut SDK 0.79.0 (`AsyncAssetsAssociationsResource.remove` still returns `None` with `cast_to=NoneType`) and photos-api main (`/api/albums/{id}/assets DELETE` still returns `status_code=204`; `UserAlbumService.remove_assets_from_album` uses `DELETE WHERE asset_id IN (...)` which silently no-ops on absent ids). Comment is still accurate — happy to revisit if you had something specific in mind.

850 tests pass; ruff/pyright clean.

ternarybits:
## Round 1 follow-up — corrected

Reopening the SDK comment thread I'd previously closed as "skipped." The reviewer was right: I'd checked the SDK at the version this PR was pinned at (0.77.0), not `latest`.

**What I missed**
- gumnut-sdk **0.82.0** (`cf54b72`, May 6) + photos-api **#577** (`a668eab`, May 5) added `not_found_assets: List[str]` to `AssetsAssociationAddResponse`. The add-album-assets endpoint no longer 404s on missing/different-library asset_ids — they're returned in the response.

**Fixed in `b7d1183`**
- Bumped `gumnut-sdk>=0.82.0`; lockfile refreshed.
- `add_assets_to_album` now reads `response.not_found_assets` and maps to `Error1.not_found` per id. The "shouldn't happen" warning branch is kept as a true contract-drift sentinel.
- New test `test_add_assets_not_found_assets_from_response` covers the success-with-not-found-bucket path.
- The `remove` endpoint is unchanged upstream — the line-276 comment is still accurate there, no change needed.

851 tests pass; ruff/pyright clean.

taggartgorman (approved):
## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (3 Consider items inline below — all minor tidiness/coverage suggestions, none blocking).
- [x] **Codebase value**: Change clearly improves on the before-state (large requests no longer 422 entirely; per-chunk error isolation; doc paragraph distinguishes this pattern from the trash-style `batched()` flow). Alternatives considered: (a) inline `batched()` per `routers/api/trash.py` — rejected because that pattern lets errors propagate, but these endpoints must catch and translate to per-id `Error1`; (b) status quo — rejected because it 422s the entire batch on >100 ids. The 107-line helper consolidates non-trivial try/except + classification + structured-log boilerplate that would otherwise duplicate across both call sites; per-endpoint composition (4-way for `add`, 2-way for `remove`) genuinely differs and stays at the call site, which is the right place for it.
- [x] **Security/auth**: No auth, authorization, or permission code touched. Diff limited to album asset-association add/remove handlers, a new utility, tests, and a doc paragraph.
- [x] **Data deletion**: `remove_asset_from_album` is album-membership removal, not asset destruction (verified — no `assets.delete`, `trash`, or destructive SDK calls). Trash flow lives in `routers/api/trash.py` and is not touched.
- [x] **Confidence**: Change is bounded to two endpoints + one new helper module + tests. Helper preserves order across chunks, isolates failures per chunk, augments transport-error logs with `chunk_size`/`request_size`. CI green (lint, typecheck, tests, version-sync, compatibility).
- [x] **Purpose clarity**: PR description matches the diff; no aspirational claims. Sequential chunking is intentionally deferred per the linked follow-up issue.

Approving via the auto-approve fast path. Three Consider-level suggestions inline — all two-way doors.

taggartgorman (commented):
## Three Consider-level suggestions (non-blocking)

GitHub's inline-comment API rejected these as line comments (a known intermittent), so capturing them here with explicit anchors.

### 1. Redundant uuid→gumnut_id reconversion — `routers/api/albums.py:201`

`chunked_per_item_bulk` already converts every uuid to its gumnut id once (`routers/utils/bulk.py:76`), so the line

```python
gumnut_asset_id = uuid_to_gumnut_asset_id(asset_uuid)
```

does the same translation a second time per asset. Two cheap shapes:

- Have the helper yield `chunk_gumnut_ids` alongside `chunk_uuids` so callers can reuse them. Cleanest if more callers land.
- Or build a `gumnut_id_by_uuid: dict[UUID, str]` once after the async loop and look up here.

`remove_asset_from_album` doesn't need this — only `add` pays for it, because only `add` has to look ids up in the SDK's `added`/`duplicate`/`not_found` sets. The caller-local map is the smaller change.

### 2. Key `errors_by_uuid` on `UUID` rather than `str` — `routers/api/albums.py:181` (and `:284`)

UUIDs are already hashable, and the loop only needs `str(asset_uuid)` for the `BulkIdResponseDto.id` field. Typing this as `dict[UUID, Error1]` and storing/looking up via the UUID directly drops one `str()` per write and one per read, and tightens the type. Same pattern duplicated at line 284 in `remove_asset_from_album`.

### 3. Add exact-boundary chunking tests — `tests/unit/api/test_albums.py:596` (and `:939`)

The new chunking tests use sizes `BULK_CHUNK_SIZE * 2`, `BULK_CHUNK_SIZE * 2 + 5`, and `BULK_CHUNK_SIZE * 3` — none pin the two boundary cases:

- `total = BULK_CHUNK_SIZE` — must produce exactly one SDK call with all ids.
- `total = BULK_CHUNK_SIZE + 1` — must produce exactly two SDK calls, second receiving a single-element list.

These are the cases most likely to regress if `batched(..., BULK_CHUNK_SIZE)` is ever replaced with hand-rolled slice math or an `if len(ids) > BULK_CHUNK_SIZE` split. Two short parametrizations would close the gap. (`tests/unit/utils/test_bulk.py::test_splits_oversized_input_into_ordered_chunks` has the same shape and would benefit similarly.)

PR #155, opened 2026-05-06
GUM-703: Fix Android asset search (limit clamp + nextPage None)

ternarybits:
## Summary

Fixes two distinct bugs in `/api/search/metadata` that crashed asset search on the Immich Android client. Same response-shape fix mirrored to `/api/search/smart`, and `/smart` now also forwards `size`/`page` (previously dropped silently).

- **HTTP 422 from oversized `limit`**: Immich client sends `size=1000` by default; photos-api caps `limit` at 200 (`MAX_PAGE_SIZE`). Now clamped at the adapter boundary via `min(int(request.size), PHOTOS_API_MAX_PAGE_SIZE)` (constant from `routers/api/constants.py`, same import already used by `timeline.py` / `sync/stream.py`). Can't tighten the DTO instead — `routers/immich_models.py` is regenerated from the upstream Immich OpenAPI spec.
- **`FormatException` on every successful response**: adapter set `nextPage=""`. Dart's null-aware `?.toInt()` short-circuits only on `null`, so `int.parse("")` crashed the mobile client. Changed to `nextPage=None` in both `/metadata` and `/smart`.
- **Forward `size`/`page` to `/smart`**: previously only `query` was forwarded, so the client's `size=1000` was silently dropped and the user got photos-api's default page size back. Now mirrors the `/metadata` kwargs-building pattern.
- **Drive-by**: stop substituting hardcoded adapter defaults for `limit`/`page` — forward only what the client provided so photos-api owns the defaults (20 / 1) as the single source of truth. The previous adapter-side default of 50 silently disagreed with photos-api's 20.

Out of scope: real pagination (`nextPage` is still always `None`, stub today).

`docs/references/code-practices.md` updated with three generic learnings (don't tighten generated DTOs — consolidated into the "Forwarding pagination parameters" subsection, mobile null-aware string parsing, forwarding pagination optionality) so the next person hits these less hard.

## Linear

https://linear.app/gumnut-ai/issue/GUM-703/asset-search-broken-on-android

## Test plan

- [x] Unit tests: 6 new tests across `TestSearchMetadata`/`TestSearchSmart` covering pagination kwarg omission, size clamp at 200, and `nextPage is None` on both endpoints.
- [x] Full suite: `uv run pytest` → all pass.
- [x] Lint/format/type: `uv run ruff format`, `ruff check`, `pyright` all clean (0 errors).
- [ ] Verify on Android client against this branch — search no longer 422s and the carousel no longer crashes on the response.


taggartgorman (commented):
Two distinct production crashes (`/metadata` HTTP 422 from `size=1000`; Android `FormatException` from `nextPage=""`) fixed cleanly with regression tests, and learnings captured in `code-practices.md`. The response-shape fix is correctly mirrored to `/smart`.

Two **Should Fix** items are about reusing the existing `PHOTOS_API_MAX_PAGE_SIZE` constant (already in `routers/api/constants.py` and imported by `timeline.py` / `sync/stream.py`) rather than introducing a third site that hard-codes `200`. Three **Consider** items: two doc-structure tweaks, plus an out-of-band note about extending the `size`/`page` story to `/smart`.

---

**[Consider] Extend `size`/`page` forwarding to `/api/search/smart`** — `routers/api/search.py:203` (out-of-band; the SDK call line isn't in the diff)

`SmartSearchDto` has the same `size` (`le=1000.0`) and `page` fields as `MetadataSearchDto`, and the Immich mobile client sends `size=1000` to smart search the same way it does to metadata search. `/smart` currently forwards only `query`, silently dropping `request.size` / `request.page` — so today there's no 422 (photos-api never sees `size`), but the client thinks it asked for 1000 results and gets only the photos-api default page size back. The PR description marks this as out-of-scope; flagging so the trade-off is explicit and not lost between branches. If extended later, mirror the same kwargs-building pattern and add a regression test analogous to `test_clamps_size_to_photos_api_ceiling` against `SmartSearchDto`.

docs/references/code-practices.md
```diff
@@ -92,7 +92,7 @@ Forgetting step 2 causes silent drift — the served web UI stays on the old Imm
 
 ### Implementing New Endpoints
 
-1. **Generate models**: Use `generate_immich_models.py` to create up-to-date Pydantic models (see [development tools](development-tools.md))
+1. **Generate models**: Use `generate_immich_models.py` to create up-to-date Pydantic models (see [development tools](development-tools.md)). The generated `routers/immich_models.py` is overwritten on every Immich version bump, so **don't tighten field constraints in that file** (e.g., dropping `Field(le=1000.0)` to `le=200.0` because the backend caps lower). The next regen restores upstream's constraint and silently reintroduces the bug. Constrain in endpoint code instead, with a comment pointing at the backend cap — e.g., `min(int(request.size), 200)` for the photos-api `MAX_PAGE_SIZE` ceiling.
```
taggartgorman: **[Consider]** Step 1 of "Implementing New Endpoints" was a one-liner about regenerating models; the new addition appends a multi-sentence rule about not tightening generated-model field constraints. That rule is a standalone concern that applies any time the generated DTOs are over-permissive — it gets buried under a numbered step here, and a reader searching for "don't edit generated files" or "field constraints" wouldn't find it.

Consider lifting it to its own subsection (e.g., "Don't tighten constraints in generated `immich_models.py`") or folding it into the new "Forwarding pagination parameters" section since the running example is the same. Step 1 then stays scannable and the rule has a heading.

docs/references/code-practices.md
```diff
@@ -152,6 +152,30 @@ for asset_uuid in request.ids:
...
 
+### Forwarding pagination parameters
+
+When forwarding pagination params (`size`, `page`, `limit`) from an Immich request to a Gumnut SDK call, forward only what the client provided — don't substitute an adapter-side default. The SDK uses an `Omit` sentinel; just leave the kwarg out of the call so photos-api applies its own default:
+
+```python
+search_kwargs: dict[str, Any] = {"query": request.description, ...}
+if request.size is not None:
+    # Clamp at the photos-api per-page ceiling (MAX_PAGE_SIZE = 200).
+    search_kwargs["limit"] = min(int(request.size), 200)
```
taggartgorman: **[Should Fix]** Use the named constant in the example (and surrounding prose).

`routers/api/constants.py` already defines `PHOTOS_API_MAX_PAGE_SIZE = 200` and is the import the codebase actually uses (`timeline.py:9`, `sync/stream.py:28`). A code-practices doc that hard-codes `200` trains readers to write the literal in new endpoints rather than reusing the constant. Same `200` literal also appears in the prose at line 95 ("`MAX_PAGE_SIZE` ceiling") and line 171 ("photos-api enforces 200") — worth referencing the symbol in those too.

```suggestion
    # Clamp at the photos-api per-page ceiling (PHOTOS_API_MAX_PAGE_SIZE).
    search_kwargs["limit"] = min(int(request.size), PHOTOS_API_MAX_PAGE_SIZE)
```

docs/references/code-practices.md
```diff
@@ -152,6 +152,30 @@ for asset_uuid in request.ids:
...
+if request.page is not None:
+    search_kwargs["page"] = int(request.page)
+gumnut_results = await client.search.search(**search_kwargs)
+```
+
+Substituting an adapter-side default (e.g., `limit = int(request.size) if request.size else 50`) fragments the source of truth — photos-api's `DEFAULT_PAGE_SIZE = 20` and an adapter-hardcoded 50 silently disagree, and a future change to photos-api's default won't propagate. Same principle for any optional kwarg passed through the adapter: preserve the optionality, don't normalize.
+
+Generated Immich DTO constraints can exceed the backend's per-page cap (e.g., `MetadataSearchDto.size` allows `le=1000.0` while photos-api enforces 200). The Immich mobile client uses these high values by default — clamp at the adapter, otherwise photos-api 422s and the user sees a generic "Failed to ..." surface.
```
taggartgorman: **[Consider]** This paragraph ("Generated Immich DTO constraints can exceed the backend's per-page cap…") restates the rule the same PR added to step 1 of "Implementing New Endpoints" (line 95 — "don't tighten field constraints in [generated models]… `min(int(request.size), 200)` for the photos-api `MAX_PAGE_SIZE` ceiling"). The unique value of the new "Forwarding pagination parameters" section is the *don't substitute an adapter-side default* rule; the DTO-cap point is a separate, already-stated fact.

Consider dropping this paragraph or shortening to one sentence cross-referencing the step-1 / new-subsection guidance.

routers/api/search.py
```diff
@@ -159,17 +159,20 @@ async def search_assets(
...
+    search_kwargs: dict[str, Any] = {
+        "query": request.description,
+        "captured_after": request.takenAfter,
+        "captured_before": request.takenBefore,
+        "person_ids": person_ids,
+    }
+    if request.size is not None:
+        # Clamp at the photos-api per-page ceiling (MAX_PAGE_SIZE = 200). The
+        # Immich client default is 1000; without this, photos-api 422s.
+        search_kwargs["limit"] = min(int(request.size), 200)
```
taggartgorman: **[Should Fix]** Use the existing `PHOTOS_API_MAX_PAGE_SIZE` constant instead of the magic `200`.

`routers/api/constants.py:3` already defines `PHOTOS_API_MAX_PAGE_SIZE = 200` (with the comment `Keep in sync with MAX_PAGE_SIZE in photos-api`), and it's already imported by `routers/api/timeline.py:9` and `routers/api/sync/stream.py:28`. Hard-coding `200` here introduces a third site that has to be remembered if the cap ever moves; the inline `MAX_PAGE_SIZE = 200` in the comment also duplicates what the constant name already encodes.

Also add `PHOTOS_API_MAX_PAGE_SIZE` to the existing `from routers.api.constants import …` block (or add a new import line if there isn't one yet).

```suggestion
    if request.size is not None:
        # Clamp at the photos-api per-page ceiling. The Immich client default
        # is 1000; without this, photos-api 422s.
        search_kwargs["limit"] = min(int(request.size), PHOTOS_API_MAX_PAGE_SIZE)
```

docs/references/code-practices.md
ternarybits: Done — restored step 1 of "Implementing New Endpoints" to its original one-liner and folded the constraint rule into the existing "Forwarding pagination parameters" subsection (your second suggested option).

docs/references/code-practices.md
ternarybits: Done — code example now imports `PHOTOS_API_MAX_PAGE_SIZE` and uses it in the `min(...)` call. Prose at the consolidated paragraph also references the symbol instead of the `200` literal.

docs/references/code-practices.md
ternarybits: Done — addressed by the same restructuring as the step-1 thread. The consolidated paragraph in "Forwarding pagination parameters" now states the DTO-cap point once with the don't-tighten-the-DTO rule attached, instead of restating step 1.

routers/api/search.py
ternarybits: Done — using `PHOTOS_API_MAX_PAGE_SIZE` from `routers/api/constants.py`, added to the import block.

taggartgorman (approved):
## Pre-approve checklist
- [x] **Codebase value**: Change leaves the codebase better than it found.
      Rationale: fixes a real Android-client crash on every successful search response (Dart `?.` doesn't short-circuit on `""`), eliminates a silent default disagreement between the adapter (50) and photos-api (20), and adds three reusable learnings to `docs/references/code-practices.md`. Alternatives considered: tightening the generated DTO (explicitly ruled out — `routers/immich_models.py` is regenerated from upstream); leaving `nextPage=""` (crashes the mobile client); narrower scope without the `/smart` mirror (same bug is present and ships in the same DTO).
- [x] **Security/auth**: PR does not touch authentication, authorization, or permission logic.
      Evidence: changes limited to `routers/api/search.py` (pagination kwarg construction, response-shape `nextPage=None`), `tests/unit/api/test_search.py`, and `docs/references/code-practices.md`. `current_user` plumbing untouched; no auth helpers modified.
- [x] **Data deletion**: PR does not touch destructive operations or data removal.
      Evidence: read-only `/api/search/{metadata,smart}` endpoints; no `delete_*`, `trash_*`, or write operations in the diff.
- [x] **Confidence**: Correctness is well understood.
      Rationale: narrow change to two endpoints sharing one pattern; the size clamp matches the existing `PHOTOS_API_MAX_PAGE_SIZE` usage in `timeline.py` / `sync/stream.py`; `nextPage` field typed `str | None` per `SearchAssetResponseDto`; 6 new tests cover omitted kwargs, clamp at 1000→200, and `nextPage is None` on both endpoints; full CI green (lint, type-check, tests, version sync, compatibility).
- [x] **Purpose clarity**: PR purpose and approach are clear from description and code. The PR body lays out both bugs, the drive-by default-substitution fix, the explicit out-of-scope (real pagination), and the docs additions; each claim verified against the diff.

LGTM — auto-approving.

PR #156, opened 2026-05-06
Parallelize per-item SDK fan-out with bounded concurrency (GUM-668)

ternarybits:
## Summary

- Add a shared `gather_with_concurrency` helper (`routers/utils/concurrency.py`) wrapping `asyncio.gather` with a `BULK_FANOUT_CONCURRENCY_LIMIT = 10` semaphore — order-preserving, propagates the first exception. Coroutines waiting on the semaphore at cancellation time are explicitly closed so cancellation doesn't leak `RuntimeWarning: coroutine was never awaited`.
- Roll the helper out to the per-item SDK fan-out sites that lacked a bulk SDK variant: `update_people`, `delete_people`, `reassign_faces` (outer per-`(asset, sourcePerson)` loop only — inner per-face stays sequential), and `add_assets_to_albums` multi-album. Per-item error mapping stays inside the per-item coroutine for the `BulkIdResponseDto`-returning endpoints; `delete_people` keeps its 204-on-all-success contract by letting the first failure propagate to the global `GumnutError` handler.
- Extract `classify_bulk_item_call` next to `chunked_per_item_bulk` in `routers/utils/bulk.py` so the per-item `APIStatusError` / `GumnutError` mapping isn't re-rolled at each fan-out call site.
- Add `emit_user_event_per_id` to `services/websockets.py` and adopt at `_bulk_permanent_delete` and `empty_trash` — replaces the per-id sequential `await emit_user_event(...)` loops with single gather waves so a 100-item chunk does one publish wave instead of 100 sequential awaits. (`delete_assets` itself was already on the bulk SDK endpoints; only the WS emit loop needed updating.)
- Document the helpers, when to catch errors inside the per-item coroutine vs. let them propagate, the concurrency-counter test pattern, and the cancellation-leak gotcha for future helper authors in `docs/references/code-practices.md`.

## Linear

https://linear.app/gumnut-ai/issue/GUM-668

## Test plan

- [x] \`uv run pytest\` (874 tests pass, including new helper-level tests and per-endpoint concurrency-counter tests)
- [x] \`uv run ruff format\` / \`uv run ruff check\` clean
- [x] \`uv run pyright\` clean
- [x] Regression test for the cancellation-leak fix: \`tests/unit/utils/test_concurrency.py::test_cancellation_does_not_warn_unawaited_coroutines\`
- [x] Existing per-item-error-mapping and SDK-error-propagation tests for the affected endpoints still pass unchanged

### DIFF
--- a/routers/api/albums.py
+++ b/routers/api/albums.py
@@ -3,19 +3,14 @@ from typing import Annotated, List
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Response
-from gumnut import (
-    APIStatusError,
-    AsyncGumnut,
-    ConflictError,
-    GumnutError,
-)
+from gumnut import AsyncGumnut
 
-from routers.utils.bulk import BulkChunkError, chunked_per_item_bulk
-from routers.utils.concurrency import gather_with_concurrency
-from routers.utils.error_mapping import (
-    classify_bulk_item_error,
-    log_bulk_transport_error,
+from routers.utils.bulk import (
+    BulkChunkError,
+    chunked_per_item_bulk,
+    classify_bulk_item_call,
 )
+from routers.utils.concurrency import gather_with_concurrency
 from routers.utils.gumnut_client import get_authenticated_gumnut_client
 from routers.utils.current_user import get_current_user
 from routers.immich_models import (
@@ -368,24 +363,14 @@ async def _add_assets_to_one_album(
     gumnut_asset_ids: list[str],
 ) -> BulkIdErrorReason | None:
     """Add assets to one album; return None on success or the mapped error."""
-    try:
-        await client.albums.assets_associations.add(
+    return await classify_bulk_item_call(
+        client.albums.assets_associations.add(
             uuid_to_gumnut_album_id(album_uuid), asset_ids=gumnut_asset_ids
-        )
-        return None
-    except ConflictError:
-        # Dead branch: upstream returns 200 with duplicate_assets, never 409.
-        return BulkIdErrorReason.duplicate
-    except APIStatusError as album_error:
-        return classify_bulk_item_error(album_error, BulkIdErrorReason)
-    except GumnutError as album_error:
-        log_bulk_transport_error(
-            logger,
-            context="add_assets_to_albums",
-            exc=album_error,
-            extra={"album_id": str(album_uuid)},
-        )
-        return BulkIdErrorReason.unknown
+        ),
+        error_enum=BulkIdErrorReason,
+        log_context="add_assets_to_albums",
+        log_extra={"album_id": str(album_uuid)},
+    )
 
 
 @router.delete("/{id}/user/{userId}", status_code=204)
--- a/routers/api/albums.py
+++ b/routers/api/albums.py
@@ -11,6 +11,7 @@ from gumnut import (
 )
 
 from routers.utils.bulk import BulkChunkError, chunked_per_item_bulk
+from routers.utils.concurrency import gather_with_concurrency
 from routers.utils.error_mapping import (
     classify_bulk_item_error,
     log_bulk_transport_error,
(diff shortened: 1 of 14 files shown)


routers/api/albums.py
```diff
@@ -344,38 +345,47 @@ async def add_assets_to_albums(
...
+    """Add assets to one album; return None on success or the mapped error."""
+    try:
+        await client.albums.assets_associations.add(
+            uuid_to_gumnut_album_id(album_uuid), asset_ids=gumnut_asset_ids
+        )
+        return None
+    except ConflictError:
+        # Dead branch: upstream returns 200 with duplicate_assets, never 409.
+        return BulkIdErrorReason.duplicate
+    except APIStatusError as album_error:
```
ternarybits: I'm pretty sure we already have a pattern in this codebase somewhere for making bulk calls and handling exceptions.  It was introduced in a PR within the past week. Can we reuse that pattern?

routers/api/people.py
```diff
@@ -117,103 +120,94 @@ async def update_people(
...
 
-    return results
+        await client.people.update(
+            person_id=gumnut_person_id,
+            **update_kwargs,
+        )
+
+        return BulkIdResponseDto(id=person_item.id, success=True, error=None)
+
+    except HTTPException as he:
```
ternarybits, line 162: Is there a common pattern in how we handle errors across bulk calls?  Think about how we could make the code cleaner.

routers/api/albums.py
ternarybits: Fixed — extracted `classify_bulk_item_call` next to `chunked_per_item_bulk` in `routers/utils/bulk.py` and adopted it here. Also dropped the dead `ConflictError` branch (the comment was already noting it as unreachable since upstream returns 200 with `duplicate_assets`).

routers/api/people.py
ternarybits: Fixed — `_update_one_person` now delegates the `APIStatusError`/`GumnutError` tail to `classify_bulk_item_call` (the same helper now used in `_add_assets_to_one_album`), mirroring the per-chunk policy in `chunked_per_item_bulk`. The pre-call exceptions specific to this endpoint (UUID-parse `ValueError`, `_resolve_thumbnail_face_id` `HTTPException`) stay at the call site. Self-review caught a regression where SDK errors from `_resolve_thumbnail_face_id` would have leaked out of the helper — fixed by wrapping both SDK calls inside a single helper invocation, with a regression test.

routers/api/albums.py
```diff
@@ -344,37 +340,36 @@ async def add_assets_to_albums(
...
+    # Walk in input order so the surfaced error is sticky-first-by-input-order,
+    # not first-to-complete (which would vary with scheduler timing).
+    first_error: BulkIdErrorReason | None = next(
+        (err for err in per_album_errors if err is not None), None
+    )
+    if first_error is None:
         return AlbumsAddAssetsResponseDto(success=True)
-    return AlbumsAddAssetsResponseDto(
-        success=False, error=first_error or BulkIdErrorReason.unknown
+    return AlbumsAddAssetsResponseDto(success=False, error=first_error)
```
ternarybits, line 357: Would it make sense to try to merge all server errors together into the error that we return, rather than return the first error?

routers/api/people.py
```diff
@@ -117,103 +117,102 @@ async def update_people(
...
+    `classify_bulk_item_call`, mirroring the per-chunk policy used by
+    `chunked_per_item_bulk` for chunked bulk endpoints. The pre-call
+    exceptions specific to this endpoint (`ValueError` from UUID parsing,
+    `HTTPException` from `_resolve_thumbnail_face_id`'s "missing face"
+    branch) stay here.
+    """
+    log_extra = {"person_id": person_item.id}
+
+    try:
+        # immich openapi specs switch between str and UUID for people id, so
```
ternarybits, line 145: Double check that this is still true.

routers/api/people.py
```diff
@@ -117,103 +117,102 @@ async def update_people(
...
+        gumnut_person_id = uuid_to_gumnut_person_id(UUID(person_item.id))
+    except ValueError as ve:
+        logger.warning(
+            "Invalid person id in bulk update",
+            extra={**log_extra, "error": str(ve)},
+        )
+        return BulkIdResponseDto(id=person_item.id, success=False, error=Error1.unknown)
+
+    try:
+        sdk_error = await classify_bulk_item_call(
```
ternarybits, line 156: Doesn't classify_bulk_item_call handle the various HTTPExceptions?  Could it?

routers/api/trash.py
```diff
@@ -64,13 +65,18 @@ async def empty_trash(
...
         )
-        for gumnut_id in chunk:
-            asset_uuid = safe_uuid_from_asset_id(gumnut_id)
-            await emit_user_event(
-                WebSocketEvent.ASSET_DELETE,
-                user_id,
-                str(asset_uuid),
+        # `emit_user_event` is fire-and-forget — gather the per-id emits so a
+        # 100-item chunk does one publish wave instead of 100 sequential awaits.
+        await asyncio.gather(
```
ternarybits: I'm seeing this pattern of wrapping `emit_user_event` with `asyncio.gather` in multiple places.  Should we create a method to capture that pattern more explicitly, just passing in a list of `gumnut_id` or `uuid` and the event type?

routers/api/albums.py
ternarybits: Skipping — `AlbumsAddAssetsResponseDto` (Immich contract) has a single `error: BulkIdErrorReason | None` field, so multiple distinct errors can't be returned without breaking the upstream schema. Alternative merge strategies (mode, "worst") would lose info rather than add it; sticking with the deterministic input-order-first pick.

routers/api/people.py
ternarybits: Verified — still true. `PeopleUpdateItem.id` is `Annotated[str, ...]` (immich_models.py:1101), while sibling DTOs (e.g. `AssetFaceUpdateItem.personId` at line 105, line 119) are typed as `Annotated[UUID, ...]`. The OpenAPI spec inconsistency persists; the inline `UUID(...)` parse is still needed and can still raise on malformed input.

routers/api/people.py
ternarybits: Engaging with the spirit of the question (consolidate redundant error handling) without extending `classify_bulk_item_call` itself — that helper specializes in SDK errors (`APIStatusError` / `GumnutError`), and pulling `HTTPException` into it would conflate FastAPI and SDK layers. Instead: `_resolve_thumbnail_face_id` is the only `HTTPException` source on this path and only ever raises HTTP 400 ("no face found"), so the 404/401/403 branches were dead defensive code. Collapsed the catch to a single `Error1.unknown` mapping with an inline comment pinning the single-source rationale (commit 37abfcd).

routers/api/trash.py
ternarybits: Done — extracted `emit_user_event_per_id(event, user_id, payload_ids)` in `services/websockets.py` next to `emit_user_event`, and adopted at both call sites (`empty_trash` here and `_bulk_permanent_delete` in `assets.py`). The helper takes an `Iterable[str]` of pre-stringified ids; both call sites pass generators. Removed the now-unused `asyncio` imports from both routers and documented the helper under "WebSocket Emission" in `docs/references/code-practices.md` (commit 37abfcd).

taggartgorman (approved):
LGTM. Reviewed against correctness, data integrity, security, silent failures, performance, test coverage, clarity, and minimalism.

Spot-checks:
- `gather_with_concurrency` cancellation-leak fix (`coro.close()` on acquire-time cancellation) is correct and pinned with a regression test.
- Per-item error-catching contract is consistent: caught inside the coroutine for `BulkIdResponseDto`-returning endpoints (`update_people`, `add_assets_to_albums`); propagated for the 204-on-all-success contract (`delete_people`).
- `add_assets_to_albums` walks errors in input order so the surfaced first-error is sticky-by-input-order, not first-to-complete.
- Concurrency-counter tests (`peak > 1` and `peak <= BULK_FANOUT_CONCURRENCY_LIMIT`) at the helper and per-endpoint level pin the contract.

Note (not blocking): `delete_people` parallelization makes the partial-state on first failure non-deterministic (up to `limit-1` in-flight deletes may complete after the error lands). Sequential had the same partial-state issue but with a deterministic break point — worth keeping in mind for any retry/idempotency reasoning at the caller.

PR #157, opened 2026-05-07
Migrate uvicorn WebSocket transport to websockets-sansio (GUM-608)

ternarybits:
## Summary

- Switch uvicorn's WebSocket transport from the deprecated legacy `websockets` impl (selected by `--ws auto`) to `--ws websockets-sansio` in the Dockerfile, `.vscode/launch.json`, the README's Production Mode recipe, and the mobile-dev guide. The legacy impl wraps its receive loop in `asyncio.shield(self.transfer_data_task)`; the shielded task raises `ConnectionClosedError` / `ConnectionClosedOK` on peer close and the surrounding lifecycle doesn't always observe it before GC, producing ~70 Sentry asyncio ERROR records per day for benign close codes 1000 / 1005 / 1011.
- Bump `uvicorn[standard]` from 0.40.0 → 0.46.0 (declared as a direct dep with floor `>=0.44.0`). The bump is required, not cosmetic: 0.44.0 added keepalive pings to the sansio impl. Without it, flipping the `--ws` flag would silence the leak but also stop probing dead peers, regressing the close-code-1011 detection the legacy impl was doing for us. 0.42.0 also fixed several sansio bugs.
- Add a `tests/unit/config/test_uvicorn_ws_config.py` static guard that asserts `uvicorn.Config(..., ws="websockets-sansio").load()` resolves to `WebSocketsSansIOProtocol` and that the sansio impl module doesn't import from `websockets.legacy`. No application code changes — the ASGI WebSocket scope/receive/send contract is identical between impls, so `python-socketio` / `python-engineio` / `services/websockets.py` all stay as-is.
- Document in `docs/references/uvicorn-settings.md` (new "ws" section) and cross-link from `docs/architecture/websocket-implementation.md` so the next person editing Socket.IO code knows the transport is configured at the uvicorn layer.

## Linear

https://linear.app/gumnut-ai/issue/GUM-608/migrate-immich-adapter-from-legacy-websockets-api-to-websocketsasyncio

## Test plan

- [x] `uv run ruff check && uv run ruff format --check && uv run pyright` — clean
- [x] `uv run pytest` — 882 tests pass, including the new static guard
- [ ] Post-deploy: watch Sentry for asyncio `"exception in shielded future"` ERROR records (close codes 1000 / 1005 / 1011) over a 48h window — they should stop appearing. The acceptance criterion in the Linear task is fundamentally a runtime signal that can't be asserted in CI without booting a full uvicorn server and racing the 75s keepalive timeout, so production monitoring is the validation surface.
- [ ] Post-deploy: smoke-test the Immich web and mobile sync streams (handshake, live updates, reconnection after a network drop) to confirm the sansio transport keeps client behavior identical.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,9 @@
                 "--port",
                 "8001",
                 "--host",
-                "0.0.0.0"
+                "0.0.0.0",
+                "--ws",
+                "websockets-sansio"
             ],
             "jinja": true,
             "envFile": "${workspaceFolder}/.env",
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,7 +106,11 @@ ENV LOG_LEVEL=info
 # Run the application with optimized uvicorn settings
 # Render sets PORT environment variable automatically (typically 10000)
 # Using exec form with shell for proper signal handling and variable substitution
+# `--ws websockets-sansio` selects the modern websockets impl; default `auto`
+# routes through the deprecated legacy module which leaks shielded-future
+# exceptions on peer close. See docs/references/uvicorn-settings.md.
 CMD ["sh", "-c", "uvicorn main:app --host 0.0.0.0 --port ${PORT:-8080} --log-level ${LOG_LEVEL} \
+  --ws websockets-sansio \
   --timeout-graceful-shutdown 60 \
   --timeout-keep-alive ${TIMEOUT_KEEP_ALIVE:-75} \
   --limit-concurrency ${LIMIT_CONCURRENCY:-200} \
--- a/README.md
+++ b/README.md
@@ -76,13 +76,14 @@ docker build --build-arg IMMICH_VERSION=v2.2.3 -t immich-adapter .
 For production deployments or when testing with mobile clients (iOS/Android), use these optimized settings:
 
 ```bash
-uv run fastapi run --port 3001 \
+uv run uvicorn main:app --port 3001 \
+  --ws websockets-sansio \
   --timeout-keep-alive 75 \
   --limit-concurrency 200 \
   --backlog 2048
 ```
 
-See [docs/references/uvicorn-settings.md](docs/references/uvicorn-settings.md) for details on these settings and why they matter for mobile clients.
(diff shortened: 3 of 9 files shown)


taggartgorman (approved):
Auto-approved by /gumnut:review-pr.

## Pre-approve checklist

- [x] **Findings**: 0 Must Fix / 0 Should Fix / 0 Consider worth surfacing.
- [x] **Codebase value**: Eliminates ~70 Sentry events/day of asyncio "exception in shielded future" noise on benign WebSocket close. Alternatives considered and rejected: (a) suppress at the logging layer — masks the symptom rather than fixing it; (b) flip `--ws` without the uvicorn floor bump — silences the leak but stops probing dead peers, regressing close-code-1011 detection that the legacy impl was doing for us. The bundled scope (flag + `uvicorn[standard]>=0.44.0` + docs + static guard) is the minimum that doesn't introduce a different regression.
- [x] **Security/auth**: Does not touch authentication, authorization, or permission logic. WebSocket transport library swap only; ASGI scope/receive/send contract is preserved between impls, so `python-engineio` / `python-socketio` / `services/websockets.py` consumers stay as-is.
- [x] **Data deletion**: Does not touch destructive operations or data removal. No trash/delete/restore paths in the diff.
- [x] **Confidence**: High. Config-only diff (Dockerfile, .vscode/launch.json, README, two guides, one reference, pyproject, lockfile, one new test); zero app code changes; all four CI checks green (API compatibility, lint/type, test, version sync); the new static guard codifies that `--ws websockets-sansio` resolves to `WebSocketsSansIOProtocol` and will fail loudly if a future uvicorn version renames or removes the key.
- [x] **Purpose clarity**: PR description clearly explains both the flag flip and why the `>=0.44.0` floor is required (keepalive pings shipped in 0.44.0), not cosmetic.

Notes for the post-deploy validation in the test plan:
- Sentry asyncio "exception in shielded future" ERROR records (close codes 1000 / 1005 / 1011) over a 48h window should drop to zero — that is the runtime acceptance criterion that can't be asserted in CI without booting a full uvicorn server and racing the 75s keepalive timer.
- Smoke-testing the Immich web + mobile sync streams (handshake, live updates, reconnect after a network drop) is the right surface to confirm the sansio transport keeps client behavior identical, since the ASGI contract guarantee is what makes this PR safe in the first place.

PR #166, opened 2026-05-12
GUM-671: Wire up /api/map/markers and enable map feature flag

taggartgorman:
## Summary

- Replace the `/api/map/markers` stub with a real implementation that paginates `client.assets.list(limit=200)`, filters in-process for assets whose `metadata.latitude` and `longitude` are both set, and applies **two bounded caps**: at most `MAP_MARKERS_CAP = 2000` markers returned, and at most `MAX_ASSETS_SCANNED = 30 × PHOTOS_API_MAX_PAGE_SIZE = 6000` assets walked. Whichever cap fires first ends iteration. Both are intentional latency bounds, not error conditions.
- Forward `fileCreatedAfter` / `fileCreatedBefore` as the SDK's native `local_datetime_after` / `local_datetime_before` kwargs.
- **Short-circuit `isFavorite=True` / `isArchived=True` to `[]`** because Gumnut tracks neither state — returning unfiltered markers under a restrictive filter would be a wrong answer. Mirrors `timeline.py`'s `isFavorite` handling. `withPartners` / `withSharedAlbums` are accept-and-drop (additive filters; the unfiltered superset is still a correct answer).
- Emit one INFO log per request with `assets_scanned`, `markers_returned`, `marker_cap_hit`, `scan_cap_hit` so the caps can be tuned from real traffic.
- Flip `server_features["map"]` from `False` to `True` so Immich web/mobile show the map view; leave `reverseGeocoding` at `False` (separate work — the flag also gates the change-location pin-drop UI which expects server-side re-geocode on asset updates).

## Why two caps

Per-page payload is the full asset object read just for three GPS fields, so each page is ~280 ms on prod. Two ceilings keep map-view latency bounded:

- `MAP_MARKERS_CAP = 2000`. Benchmarked against a real library at ~70% GPS-tagged density: 2000 markers ≈ 15 pages ≈ 4 s; 5000 ≈ ~11 s. 11 s is too slow for a map-view load. The SDK orders newest-first, so the cap drops the oldest GPS-tagged assets when it fires.
- `MAX_ASSETS_SCANNED = 6000` (30 pages, ~8 s ceiling). Bounds the worst case where a *low* GPS-density library would otherwise walk hundreds of pages chasing a marker cap it'll never fill. Without this, e.g. 5% density × 50K assets ≈ 200 pages × ~280 ms ≈ ~56 s — past any reasonable client/edge timeout.

If real usage shows 2000 markers is insufficient, the next step is either a slim-projection mode on `/api/assets` or a dedicated backend `/map/markers` endpoint.

## Notes for reviewers

- The closest precedent is `routers/api/memories.py::_fetch_assets_for_day` — same paginated `client.assets.list()` + explicit `break` after a result cap, same `Annotated[T | SkipJsonSchema[None], Query()] = None` parameter style.
- The SDK accepts `local_datetime_after`/`before` natively (`Union[str, datetime, None]`), so the date-range forwarding uses typed kwargs rather than the `extra_query` dict that `timeline.py` still uses. Both produce the same wire request.
- `docs/references/code-practices.md` picks up two checklist addenda: (1) the cross-doc updates a stub→real promotion implies (`server_features` flip + the two evergreen status docs + grep for paragraphs summarizing the prior state); (2) restrictive filters the backend can't honor must short-circuit, not silently drop.

## Test plan

- [x] `uv run ruff check && uv run ruff format && uv run pyright` clean
- [x] `uv run pytest` — full suite passes
- [x] `tests/unit/api/test_map.py` covers: GPS happy path, `metadata is None` skip, `latitude`/`longitude` null skip, null location names, date-range forwarding, no-date-range case, partner/shared-album drop, the `isFavorite`/`isArchived` short-circuit (parametrized over True/False/None), the 2000-marker cap, and the 6000-asset scan cap. Both cap tests use a paginating mock so dropping the explicit `break` walks extra pages and fails.
- [x] Updated integration test moves `"map"` to the enabled feature-flag set.
- [ ] Manual: point an adapter at production photos-api, hit `GET /api/map/markers` from Immich web and mobile; confirm markers render and date-range filters narrow the result set.

### DIFF
--- a/routers/api/map.py
+++ b/routers/api/map.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import datetime
 from typing import Annotated, Any, List
 
@@ -10,6 +11,8 @@ from routers.immich_models import MapMarkerResponseDto, MapReverseGeocodeRespons
 from routers.utils.gumnut_client import get_authenticated_gumnut_client
 from routers.utils.gumnut_id_conversion import safe_uuid_from_asset_id
 
+logger = logging.getLogger(__name__)
+
 
 router = APIRouter(
     prefix="/api/map",
@@ -20,9 +23,12 @@ router = APIRouter(
 
 # Hard cap on returned markers. The per-page payload is the full asset object
 # (faces, people, urls, exif) read just for three GPS fields, so each page is
-# ~280 ms on prod. 2000 markers ≈ 15 pages ≈ 4 s; 5000 ≈ 11 s — too slow for a
-# map-view load. Revisit (slim-projection asset list, or a dedicated backend
-# `/map/markers` endpoint) if real usage shows 2000 is insufficient.
+# ~280 ms on prod. Benchmarked against a real library at ~70% GPS-tagged
+# density: 2000 markers ≈ 15 pages ≈ 4 s; 5000 ≈ 31 pages ≈ 11 s — too slow
+# for a map-view load. Revisit (slim-projection asset list, or a dedicated
+# backend `/map/markers` endpoint) if real usage shows 2000 is insufficient.
+# The SDK orders by capture time descending, so when the cap fires the
+# oldest GPS-tagged assets are the ones dropped.
 MAP_MARKERS_CAP = 2000
 
 
@@ -66,6 +72,21 @@ async def get_map_markers(
             )
         )
         if len(markers) >= MAP_MARKERS_CAP:
+            # Log so the team has signal when real usage hits the cap — the
+            # response itself has no truncation marker, so this is the only
+            # observable. If this fires regularly, revisit the cap.
+            logger.warning(
+                "Map markers response truncated at cap",
+                extra={
+                    "cap": MAP_MARKERS_CAP,
+                    "file_created_after": fileCreatedAfter.isoformat()
+                    if fileCreatedAfter is not None
+                    else None,
+                    "file_created_before": fileCreatedBefore.isoformat()
+                    if fileCreatedBefore is not None
+                    else None,
+                },
+            )
             break
     return markers
 
--- a/routers/api/map.py
+++ b/routers/api/map.py
@@ -1,8 +1,14 @@
 from datetime import datetime
-from typing import List
-from fastapi import APIRouter, Query
+from typing import Annotated, Any, List
 
+from fastapi import APIRouter, Depends, Query
+from gumnut import AsyncGumnut
+from pydantic.json_schema import SkipJsonSchema
+
+from routers.api.constants import PHOTOS_API_MAX_PAGE_SIZE
 from routers.immich_models import MapMarkerResponseDto, MapReverseGeocodeResponseDto
+from routers.utils.gumnut_client import get_authenticated_gumnut_client
+from routers.utils.gumnut_id_conversion import safe_uuid_from_asset_id
 
 
 router = APIRouter(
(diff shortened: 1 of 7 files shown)


routers/api/map.py
```diff
@@ -12,21 +21,74 @@
...
+            )
+        )
+        if len(markers) >= MAP_MARKERS_CAP:
+            # Log so the team has signal when real usage hits the cap — the
+            # response itself has no truncation marker, so this is the only
+            # observable. If this fires regularly, revisit the cap.
+            logger.warning(
+                "Map markers response truncated at cap",
+                extra={
+                    "cap": MAP_MARKERS_CAP,
```
taggartgorman: **[Consider]** The 2000-marker cap bounds *markers returned* but not *assets walked*. The loop only `break`s once `len(markers) >= MAP_MARKERS_CAP`, so for libraries with low GPS density it can fan out across many pages before filling the cap (or never filling it).

Concrete worst case: a user with 50K assets at ~5% GPS density walks ~40K assets ≈ 200 pages × ~280 ms ≈ 56 s before filling the cap — past any reasonable client/edge timeout. The existing `WARNING` fires *only* on cap hit, so this worst case is observability-silent: we'd see request timeouts in Sentry/Render logs but no signal pointing at this endpoint.

Cheap mitigation: add a sibling `MAX_PAGES_WALKED` cap (e.g., 30 pages ≈ 6K assets ≈ ~8 s ceiling) that also logs `WARNING` when hit. Same observability principle as the existing cap, just bounding the other axis.

```python
MAX_PAGES_WALKED = 30  # ~6K assets at PHOTOS_API_MAX_PAGE_SIZE=200; ~8s ceiling

markers: list[MapMarkerResponseDto] = []
pages_walked = 0
async for asset in client.assets.list(**list_kwargs):
    # ... existing per-asset logic ...
    # at page boundary (e.g., every PHOTOS_API_MAX_PAGE_SIZE iterations) check and break
```

Non-blocking — the PR's own benchmark assumes 70% GPS density, which is realistic for most users with phone-sourced photos. Mostly worth surfacing because the failure mode is silent (request timeout) rather than visibly truncated.

taggartgorman (commented):
## Auto-approve checklist (fast path)

This PR triages as auto-approvable per `/gumnut:review-pr` (0 Must Fix / 0 Should Fix). Posting as COMMENT instead of APPROVE only because GitHub blocks self-approval.

- [x] **Findings**: 0 Must Fix / 0 Should Fix (1 Consider — see line comment on `routers/api/map.py`)
- [x] **Codebase value**: Ships a real feature replacing a stub, flips the gating flag, and codifies the cross-doc update pattern in `code-practices.md` so the next stub→real promotion is less likely to drift. "Do nothing / leave stub" would surface a broken map; "ship a backend `/map/markers` endpoint" is the larger change correctly deferred via the cap + WARN.
- [x] **Security/auth**: Endpoint stays behind `Depends(get_authenticated_gumnut_client)`; no auth/permission logic touched. Diff in `server.py` is a single boolean flag flip.
- [x] **Data deletion**: Read-only endpoint; no writes, no `delete`, no `trash` calls.
- [x] **Confidence**: SDK default `state="live"` excludes trashed assets (verified against SDK docstring). Follows the `memories.py::_fetch_assets_for_day` pattern. 8 unit tests cover the GPS happy path, all None-skip variants, date forwarding, drop-and-ignore filters, and the cap with a paginating mock. CI green across lint / type / test / compatibility.
- [x] **Purpose clarity**: Clearly stated in PR body and the inline comment above `MAP_MARKERS_CAP`.

## Summary

The implementation, the param-typing migration (`bool = Query(default=None)` → `Annotated[bool | SkipJsonSchema[None], Query()] = None`), the doc-consistency updates, and the new `code-practices.md` checklist item all land cleanly together. Nice touch flagging the `reverseGeocoding` flag stays `False` with a scoped comment explaining why.

routers/api/map.py
taggartgorman: Done in bb41ede — added `MAX_ASSETS_SCANNED = 30 * PHOTOS_API_MAX_PAGE_SIZE` (6000 assets ≈ ~8 s ceiling), so the worst-case low-density library now bounds at ~8 s instead of ~56 s.

Implemented as a per-asset counter rather than a page-boundary check since the SDK paginator yields one asset at a time — same effect, simpler loop.

Dropped the `WARNING` on the marker cap in the same commit (both caps are intentional bounds to keep map-view latency reasonable, not failures that warrant an observable — if we ever do hit the timeouts the scan cap is preventing, those would surface as request errors regardless).

routers/api/map.py
```diff
@@ -12,21 +18,72 @@
...
+                MapMarkerResponseDto(
+                    id=str(safe_uuid_from_asset_id(asset.id)),
+                    lat=metadata.latitude,
+                    lon=metadata.longitude,
+                    city=metadata.city,
+                    state=metadata.state,
+                    country=metadata.country,
+                )
+            )
+            if len(markers) >= MAP_MARKERS_CAP:
```
ternarybits, line 97: I think this method could use some logging to explain how may assets it fetched, how many it returned with lat/long, if it hit a cap, etc.

routers/api/map.py
```diff
@@ -12,21 +18,72 @@
...
+    """Return up to `MAP_MARKERS_CAP` markers for the caller's GPS-tagged assets.
+
+    `isArchived` / `isFavorite` / `withPartners` / `withSharedAlbums` are
+    accepted for client-compatibility (Immich clients send them) but have no
+    Gumnut analog, so they're dropped at the adapter.
     """
-    Return a list of map markers.
-    Gumnut currently does not support mapping, so this is a stub implementation that returns an empty list.
-    """
+    _ = isArchived, isFavorite, withPartners, withSharedAlbums  # accepted, dropped
```
ternarybits: **[Should Fix] Favorites/archive map filters return unfiltered markers**

`isFavorite` and `isArchived` are accepted here but then dropped, so a request like `/api/map/markers?isFavorite=true` or `?isArchived=true` returns all GPS markers instead of the filtered result. Since the adapter currently emits assets as non-favorite timeline assets elsewhere, the adjacent timeline endpoints short-circuit these unsupported filters to `[]`.

Please mirror that behavior here by returning `[]` when `isFavorite is True` or `isArchived is True`, and add unit coverage for those query combinations.

ternarybits (commented):
Thanks for wiring this up. The marker implementation and caps are well covered, and CI is green. I found two issues worth fixing before merge.

**[Should Fix] Stale server-features doc still says map is disabled**

`docs/design-docs/immich-adapter-gap-analysis.md:307` still says GUM-552 flipped `map: true` to `false` because map was unsupported. This PR now re-enables `map` and updates the map gap section, so the later paragraph contradicts both `routers/api/server.py` and the updated gap entry.

Please remove `map` from that disabled-feature list, or add a follow-up sentence noting that map was later re-enabled once marker support shipped while `reverseGeocoding` remains disabled.

routers/api/map.py
taggartgorman: There was logging the the first pass, but I removed it as I thought it looked noisy. I'll add it back in.

ternarybits (approved):

routers/api/map.py
taggartgorman: Added an INFO log at the end of `get_map_markers` with structured `extra` keys (`assets_scanned`, `markers_returned`, `marker_cap_hit`, `scan_cap_hit`). `scan_cap_hit` is computed defensively so it isn't reported when both caps coincide.

routers/api/map.py
taggartgorman: Fixed — `isFavorite=True` / `isArchived=True` now short-circuit to `[]` before any SDK call (mirrors `timeline.py`). Used `is True` so `False`/`None` (no-restriction) still return normal results. Added a parametrized unit test covering both branches; renamed the prior `test_unsupported_filters_are_dropped` to `test_partner_and_shared_album_filters_are_dropped` since favorites/archived are now handled separately.

PR #168, opened 2026-05-14
GUM-767: Remove adapter-side orientation swap; surface raw dims from photos-api

taggartgorman:
## Summary

- Remove the adapter's compensating orientation helpers (`display_dimensions`, `wire_orientation`) — photos-api now stores `asset.width`/`asset.height` in display space at ingest, so the swap is no longer needed and would double-correct against the new ingest semantics.
- Rewire `exifInfo.exifImageWidth`/`exifImageHeight` to read pre-rotation raw dims from `metadata.raw_width`/`raw_height` (new fields exposed by the photos-api wire serializer), with NULL fallback to `asset.width`/`asset.height` for drift-cohort rows whose original raw dims were never captured.
- Emit the EXIF `orientation` tag unchanged on the wire. Immich mobile pairs it with the raw dims to re-derive display dims locally; Immich web reads `asset.width`/`height` directly and ignores it.
- Bump `gumnut-sdk` floor to `>=0.90.0` (the release that exposes `raw_width`/`raw_height` on `MetadataResponse`); resolves to `0.91.0`.
- Refresh `docs/references/code-practices.md` "Asset dimensions and orientation" section to describe the new contract and point future contributors at the new `exif_dims_and_orientation(gumnut_asset)` helper instead of the deleted helpers. The helper returns `(exifImageWidth, exifImageHeight, wire_orientation)` and nulls the orientation tag on the drift-cohort fallback path so mobile can't double-rotate.

## :warning: Deploy ordering — do not merge until partner repo is live

This PR depends on the photos-api side of the same change landing first, and on its dim-swap backfill draining in production. If this PR ships before that:

- For every portrait-orientation asset, photos-api still returns raw landscape `asset.width`/`asset.height`. The adapter would emit those un-swapped, and Immich web's layout box would size landscape while the served pixels are portrait — every portrait stretched.

Confirm the partner photos-api PR is deployed and its backfill has fully drained before merging this one.

## Linear

https://linear.app/gumnut-ai/issue/GUM-767/immich-adapter-remove-display-dimensions-wire-orientation-swap

## Test plan

- [x] `uv run ruff check` — clean
- [x] `uv run ruff format` — clean
- [x] `uv run pyright` — 0 errors / 0 warnings
- [x] `uv run pytest` — 890 passed (existing orientation-swap tests rewritten to assert passthrough + raw-dim semantics; portrait-asset regression test added that verifies a known orientation-6 asset emits raw dims on `exifImageWidth/Height`, display dims on `asset.width/height`, and the orientation tag unchanged)
- [ ] After partner deploy: manual verification — a known portrait asset displays correctly in Immich web (layout box matches served pixels), and face thumbnail crops on the same asset land on the actual face.


routers/utils/asset_conversion.py
```diff
@@ -35,49 +35,25 @@ def normalize_rating(rating: float | int | None) -> int | None:
...
-        return None
-    if width and height and orientation in _FLIPPED_ORIENTATIONS:
-        return None
-    return str(orientation)
+    metadata = asset.metadata
+    raw_w = getattr(metadata, "raw_width", None) if metadata else None
+    raw_h = getattr(metadata, "raw_height", None) if metadata else None
+    return (
+        raw_w if raw_w is not None else asset.width,
+        raw_h if raw_h is not None else asset.height,
```
taggartgorman: **[Should Fix — verification request]** This fallback may regress drift-cohort portrait shots on mobile.

For a drift-cohort row (`raw_width`/`raw_height` NULL, asset dims display-space post-backfill) tagged `orientation=6`:
- Wire emits `exifImageWidth=2268, exifImageHeight=4032, orientation="6"` (display-space dims + unchanged tag).
- Per the PR description, mobile pairs these and applies the 5–8 swap → derives display `(4032, 2268)`.
- But served pixels and `asset.width`/`height` are `(2268, 4032)` — the layout box mismatches the served pixels.

This is the same double-rotation class of bug the deleted `wire_orientation` was guarding against ("_the orientation tag must be nulled out — otherwise [the client] would re-apply the rotation on top of swapped dims_"). On the fallback path, the new code drops that protection.

Two ways to resolve:
1. Have the helper signal the fallback path and null `orientation` at the callsites for those rows (degrades drift rows to the old wire contract).
2. Verify that the partner photos-api backfill *also* populates `raw_width`/`raw_height` for drift rows, in which case the fallback is empirically dead — and the docstring's claim that the fallback yields "the same value mobile would compute from raw dims + orientation anyway" should be reworded (it's still wrong for rotated portraits if hit).

The existing `test_sync_exif_v1_falls_back_to_asset_dims_when_raw_dims_null` asserts `orientation == "6"` on the fallback path — that assertion is what would need to flip if option 1 is chosen.

routers/utils/asset_conversion.py
```diff
@@ -35,49 +35,25 @@ def normalize_rating(rating: float | int | None) -> int | None:
...
+    is already display-space for that cohort and is the same value mobile
+    would compute from raw dims + orientation anyway.
     """
-    if orientation is None:
-        return None
-    if width and height and orientation in _FLIPPED_ORIENTATIONS:
-        return None
-    return str(orientation)
+    metadata = asset.metadata
+    raw_w = getattr(metadata, "raw_width", None) if metadata else None
```
taggartgorman: **[Consider]** With the SDK floor bumped to `>=0.90.0`, `raw_width`/`raw_height` are guaranteed fields on `MetadataResponse` (both typed as `Optional[int] = None`). `getattr` with a default defends against an `AttributeError` that can't happen — direct attribute access is clearer:

```python
metadata = asset.metadata
if metadata is None:
    return asset.width, asset.height
return (
    metadata.raw_width if metadata.raw_width is not None else asset.width,
    metadata.raw_height if metadata.raw_height is not None else asset.height,
)
```

Not blocking — `extract_sync_exif` in the same module uses the `getattr` pattern defensively, so this matches that local style. But `extract_exif_info` does direct access after an early-return for `metadata is None`, so both already coexist.

taggartgorman (commented):
Self-review caught the bulk of this cleanly — extracting `raw_dimensions_with_fallback`, doc refresh, and comprehensive test rewrites. One verification request on the drift-cohort fallback and one Consider-level minimalism note.

The new wire contract is well-described and the deploy-ordering ⚠️ in the PR description is good practice. Tests cover the main rotation case and the drift-cohort fallback explicitly — the line comment below is asking whether the test expectation is actually correct for mobile clients in that fallback case.

routers/utils/asset_conversion.py
taggartgorman: Fixed — verified your math (the doc's outcome assertion doesn't survive a concrete trace) and went with option 1. Helper is now `exif_dims_and_orientation(asset)` and returns `(width, height, wire_orientation_str | None)`; the orientation tag is nulled on the fallback path so mobile doesn't re-apply the 5–8 swap to display-space dims. The fallback test now asserts `orientation is None`, and `code-practices.md` documents the new contract.

Option 2 was empirically dead per photos-api's own `exif-orientation-handling.md`: NULL means drift-cohort or tag-less file even post-backfill, so the fallback path is live.

routers/utils/asset_conversion.py
taggartgorman: Fixed — helper now uses direct attribute access with an explicit `metadata is None` early-return, no `getattr` defaults. Folded into the drift-cohort rewrite above.

taggartgorman:
## Round 2: feedback addressed

Both inline threads auto-fixed (two-way doors).

**Auto-fixed**
- [Thread 1 — Should Fix] **Drift-cohort double-rotation on fallback path** (`routers/utils/asset_conversion.py`). The trace was correct: emitting display-space `asset.width/height` on `exifImageWidth/Height` plus a non-null portrait orientation made mobile re-apply the 5–8 swap and derive landscape dims. Helper renamed to `exif_dims_and_orientation` and now returns `(width, height, wire_orientation_str | None)` — the orientation tag is nulled on the fallback path. Callers emit all three tuple elements verbatim, so the rule can't be forgotten at a future call site. Fallback tests updated to assert `orientation is None`. Doc section in `code-practices.md` rewritten for the new contract.
- [Thread 2 — Consider] **Direct attribute access in the helper**. Helper now does `if metadata is None: return ...` early-return + direct `metadata.raw_width` / `raw_height` access instead of `getattr` defaults. Two dead `orientation` locals in `extract_exif_info` and `extract_sync_exif` removed (the helper extracts orientation from metadata internally).

**Also cleaned up (self-review caught)**
- Scrubbed 7 internal issue-tracker IDs from test docstrings/comments in `tests/unit/api/sync/test_orientation.py`, `tests/unit/utils/test_asset_conversion.py`, and `tests/unit/api/test_timeline.py` — per the repo's public-repo hygiene rule (`docs/references/code-practices.md` line 20).

Checks: `ruff check` clean, `pyright` 0 errors, `pytest` 890 passed.

Both threads resolved.

routers/utils/asset_conversion.py
```diff
@@ -35,49 +35,39 @@ def normalize_rating(rating: float | int | None) -> int | None:
...
-
-    Gumnut stores raw sensor dimensions plus an EXIF orientation tag. Immich's
-    wire contract expects ``asset.width`` / ``asset.height`` to already reflect
-    the post-rotation (display) dimensions — clients (e.g. immich web's
-    ``getAssetRatio``) don't consult orientation when sizing layout boxes.
-
-    For orientations 5–8 (90°/270° rotations), swap the dimensions; otherwise
-    return them unchanged.
+def exif_dims_and_orientation(
+    asset: AssetResponse,
```
ternarybits: Consider: every other `AssetResponse`-accepting helper in this module names the parameter `gumnut_asset` (`extract_exif_info(gumnut_asset)`, `extract_sync_exif(gumnut_asset, ...)`, `build_asset_upload_ready_payload(gumnut_asset, ...)`, `convert_gumnut_asset_to_immich(gumnut_asset, ...)`). The new `exif_dims_and_orientation(asset)` breaks the local convention. All four call sites already hold a `gumnut_asset` variable. Non-blocking — pure naming nit.

ternarybits (approved):
## ✅ Approve

Clean two-way swap: 2 helpers (`display_dimensions` + `wire_orientation`) → 1 helper (`exif_dims_and_orientation`) that returns a 3-tuple so callers can't forget to pair dims with orientation. Net less code, tighter contract, tests rewritten to assert emit-site behavior rather than helper mechanics. Drift-cohort fallback (raw NULL → asset.width/height + null orientation) is well documented and tested; the doc refresh in `code-practices.md` keeps contributor guidance aligned.

Cross-repo deploy ordering is correctly flagged in the PR body — partner photos-api PR (GUM-766) is on a branch and not yet on `origin/main`, and SDK 0.90+ (resolves to 0.91.0) does expose `raw_width/raw_height` on `MetadataResponse`. Per `docs/references/exif-orientation-handling.md`, the drift-cohort treatment (NULL → display-space `asset.width/height`) matches the documented contract.

### Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (2 Consider posted as non-blocking inline / inline-prose comments)
- [x] **Codebase value**: −20 LOC net, single source of truth for the dims/orientation pairing, eliminates "forgot to call the companion helper" as a class of bug. Alternatives considered: keeping `display_dimensions` + `wire_orientation` (rejected — leaves the call-site-pair-or-bug hazard) and shipping a NamedTuple return (overkill at 4 callers that destructure immediately).
- [x] **Security/auth**: PR doesn't touch authentication, authorization, or permission logic. Checked all changed files; pure wire-conversion logic.
- [x] **Data deletion**: No destructive operations. Pure read-side metadata reshaping.
- [x] **Confidence**: Correctness is well-understood. Verified that immich web's `getDimensions(exifInfo)` (swaps on orientation 5–8) and `getAssetRatio(asset)` (reads `asset.width/height` directly) both produce correct display dims under the new wire contract; mobile re-derives display dims from raw + orientation locally. The drift-cohort `(asset.width/height, None)` path leaves both web (no swap when orientation is None) and mobile (no orientation → no rotation) at the right answer.
- [x] **Purpose clarity**: PR description, code-practices doc section, and helper docstring all describe the new contract and deploy-ordering hazard consistently.

### Consider-level (non-blocking)

- **PR body still references `raw_dimensions_with_fallback(asset)`** in the fifth Summary bullet. The committed helper is `exif_dims_and_orientation` (rename confirmed in Round 2). Worth a quick edit so future readers chasing the helper name from the PR don't hit a dead end.

(One additional Consider on parameter naming is posted as an inline comment.)

routers/utils/asset_conversion.py
taggartgorman: Fixed — renamed to `gumnut_asset` in `routers/utils/asset_conversion.py` and updated the signature reference in `docs/references/code-practices.md`. All callers pass positionally, so no call-site changes were needed.

PR #171, opened 2026-05-15
Re-enable video playback streaming via CDN with iOS-compatible Accept-Ranges

taggartgorman:
## Summary

- Restore the `/api/assets/{id}/video/playback` endpoint that was stubbed to `200 OK` in `3499d1b`. The endpoint now streams the asset's `original` variant from CDN via the existing `_retrieve_and_stream_variant` helper and forwards the client's `Range` header for seeking. The end-to-end Range path through the Cloudflare Worker → R2 was already in place.
- Advertise `Accept-Ranges: bytes` on every successful CDN response in `stream_from_cdn`, not just on `206`. iOS AVPlayer probes this header on the initial 200 to decide whether the source is seekable — without it, MP4s whose `moov` atom isn't at the front are unplayable and AVPlayer can fail abruptly, which matches the iOS crash that motivated the original revert. R2 via the Worker supports byte ranges unconditionally, so this is safe on all asset variants.
- Add `TestPlayAssetVideo` (success / Range forwarding / missing-variant 404) and a `test_accept_ranges_on_200` case in `test_cdn_client.py`; update `adapter-architecture.md` (Assets row) and `immich-adapter-gap-analysis.md` (gap §33 flipped to Closed) to reflect the change.
- Drive-by `code-practices.md` clarification: the no-internal-IDs rule now covers every committed surface (code, docs, PR bodies, commit messages), not just code/docstrings.

## Test plan

- [x] `uv run ruff check` — passes
- [x] `uv run ruff format` — clean
- [x] `uv run pytest tests/unit` — 865 passed
- [ ] **Manual iOS device test** — required pre-merge. The `Accept-Ranges`-on-200 hypothesis is the highest-likelihood fix from the original investigation notes, but I can't iOS-test from the dev environment. If iOS still crashes, candidates to try next are upstream `Content-Type` precision and AVPlayer's behavior across redirects.
- [ ] Manual web playback regression check (was already working before; should be unaffected since `Accept-Ranges` was already emitted on `206`).

### DIFF
--- a/docs/architecture/adapter-architecture.md
+++ b/docs/architecture/adapter-architecture.md
@@ -361,7 +361,7 @@ The adapter implements a subset of Immich's API surface. Unimplemented endpoints
 
 | Area | Endpoints | Notes |
 |------|-----------|-------|
-| Assets | Upload, download (original + thumbnail), video playback, delete, bulk delete, existence check, statistics | Streaming downloads via `StreamingResponse`; video playback streams the `original` variant from CDN with Range/seek support and advertises `Accept-Ranges: bytes` on the initial 200 so iOS AVPlayer treats the source as seekable; `DELETE /api/assets` soft-deletes by default and permanently deletes when `force=true` |
+| Assets | Upload, download (original + thumbnail), video playback, delete, bulk delete, existence check, statistics | Streaming downloads via `StreamingResponse`; video playback streams the `original` variant from CDN with Range/seek support; `DELETE /api/assets` soft-deletes by default and permanently deletes when `force=true` |
 | Trash | Restore-by-ids, restore-all, empty-trash | `trashDays` comes from `TRASH_RETENTION_DAYS`; web and mobile clients see real trash state |
 | Albums | CRUD, add/remove assets, statistics | User sharing not supported (returns 501) |
 | People | CRUD, list with pagination/sort/filter, thumbnails, statistics, merge | |
--- a/docs/architecture/adapter-architecture.md
+++ b/docs/architecture/adapter-architecture.md
@@ -1,6 +1,6 @@
 ---
 title: "Immich Adapter Architecture"
-last-updated: 2026-05-14
+last-updated: 2026-05-15
 ---
 
 # Immich Adapter Architecture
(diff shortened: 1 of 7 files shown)


docs/architecture/adapter-architecture.md
```diff
@@ -361,7 +361,7 @@ The adapter implements a subset of Immich's API surface. Unimplemented endpoints
 
 | Area | Endpoints | Notes |
 |------|-----------|-------|
-| Assets | Upload, download (original + thumbnail), delete, bulk delete, existence check, statistics | Streaming downloads via `StreamingResponse`; `DELETE /api/assets` soft-deletes by default and permanently deletes when `force=true` |
+| Assets | Upload, download (original + thumbnail), video playback, delete, bulk delete, existence check, statistics | Streaming downloads via `StreamingResponse`; video playback streams the `original` variant from CDN with Range/seek support and advertises `Accept-Ranges: bytes` on the initial 200 so iOS AVPlayer treats the source as seekable; `DELETE /api/assets` soft-deletes by default and permanently deletes when `force=true` |
```
taggartgorman: Consider: this Notes cell now packs three independent topics (StreamingResponse for downloads, video-playback specifics with iOS AVPlayer / moov-atom rationale, DELETE soft-delete + force semantics) and is materially longer than any other Notes cell in the table. The iOS-AVPlayer detail is already captured in gap-analysis §33 and in the cdn_client.py inline comment. A tighter clause keeps the architecture table scannable — e.g.:

`Streaming downloads via `StreamingResponse`; video playback streams the `original` variant from CDN with Range/seek support; `DELETE /api/assets` soft-deletes by default and permanently deletes when `force=true`.`

Non-blocking.

taggartgorman (commented):
Self-review (GitHub blocks self-APPROVE, so posting as COMMENT).

## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (3 Consider items below — non-blocking)
- [x] **Codebase value**: Stub → real implementation that reuses `_retrieve_and_stream_variant`; one-line move of `Accept-Ranges` out of the 206 branch makes the header consistent across 2xx responses. Alternatives ("leave stub", "implement without the Accept-Ranges fix", "duplicate per-endpoint logic") are clearly worse.
- [x] **Security/auth**: Touched only `play_asset_video` (uses the standard `get_authenticated_gumnut_client` Depends, like sibling endpoints) and a response-header tweak in `stream_from_cdn`. No auth/permission logic changed.
- [x] **Data deletion**: Read/streaming path only.
- [x] **Confidence**: Diff is small and self-contained; new branches (no-Range, with-Range, missing-variant 404) all have direct unit-test coverage. iOS-AVPlayer runtime behavior is an external gate the author explicitly flagged as pre-merge in the test plan — that's the load-bearing manual verification, and code review can't substitute for it.
- [x] **Purpose clarity**: Description and code agree on intent.

## Non-blocking observations

1. **Dangling SHA in this PR body.** "...stubbed to `200 OK` in `3499d1b`" — `git log --all --oneline | grep 3499d1b` returns nothing in this repo. Worth editing the description to either drop the SHA or point at the actual stubbing commit (search suggests it traces back to the original CDN-streaming change, not `3499d1b`). The corresponding first-commit message body has the same dangling SHA, but per the third commit's own framing, historical references are out of scope for retroactive cleanup; the PR body is still editable.

2. **PR title carries `GUM-467:`.** The new rule in `docs/references/code-practices.md` names PR bodies and commit messages explicitly but not PR titles. PR titles are the most visible piece of public PR metadata and are mutable — worth either dropping the prefix from this title or carving an explicit exception in the rule. Lower-priority than (1).

3. See inline comment on `docs/architecture/adapter-architecture.md` (Assets row).

docs/architecture/adapter-architecture.md
taggartgorman: Fixed in 7f20d28 — dropped the iOS-AVPlayer / Accept-Ranges / moov-atom clause from the Notes cell. Rationale is preserved in gap-analysis §33 and the cdn_client.py inline comment, so the table cell can be a one-clause summary again.

ternarybits (approved):

PR #172, opened 2026-05-15
GUM-821: cascade through file_created_at before falling to created_at

ternarybits:
## What
- Mirror Gumnut's REST `Asset.local_datetime` cascade in both Immich-adapter converters in `routers/utils/asset_conversion.py` (`build_asset_upload_ready_payload` and `convert_gumnut_asset_to_immich`). New cascade for `fileCreatedAt` and `localDateTime`: `metadata.original_datetime -> asset.file_created_at -> asset.created_at`. Symmetric chain for `fileModifiedAt`: `metadata.modified_datetime -> asset.file_modified_at -> asset.updated_at`.
- New `TestDateCascade` class in `tests/unit/utils/test_asset_conversion.py` covering metadata-wins, file-fallback, and full-fallback paths against both converters.
- Pre-existing mocks in `test_assets.py`, `test_memories.py`, and `test_search.py` now set `file_created_at`/`file_modified_at` to `None` so they reach the converters' real fallback path rather than operating on auto-generated `Mock` attribute stubs.

## Why
- When `metadata.original_datetime` is null (asset bytes have no usable EXIF `DateTimeOriginal`), the adapter was falling all the way back to `Asset.created_at` - DB row insertion time, i.e. upload time - skipping `Asset.file_created_at`, the uploader-asserted capture date. Result: any asset uploaded with a correct `timestamp` but no EXIF capture-date showed "today" in the Immich timeline.
- This is the root cause of GUM-770. A temporary seed-script workaround PATCHes `original_datetime` after asset creation; with this fix in place that workaround can be reverted.
- Fixes the timeline for **all** existing assets where `file_created_at` was set but EXIF is missing, not just newly-seeded ones. Pure internal precedence change - no schema, API shape, or contract changes.

## Linear
https://linear.app/gumnut-ai/issue/GUM-821/immich-adapter-fall-back-to-file-created-at-before-created-at-in-date

## Test plan
- [x] `uv run pytest` - 893 passed
- [x] `uv run ruff format && uv run ruff check && uv run pyright` - clean
- [ ] After merge, file follow-up to revert the temporary seed-data `original_datetime` PATCH workaround.

### DIFF
--- a/routers/utils/asset_conversion.py
+++ b/routers/utils/asset_conversion.py
@@ -282,13 +282,29 @@ def build_asset_upload_ready_payload(
         gumnut_asset.metadata.modified_datetime if gumnut_asset.metadata else None
     )
 
-    # fileCreatedAt: capture time in actual UTC, fallback to upload time
-    file_created_at = to_actual_utc(metadata_original_dt) or gumnut_asset.created_at
-    # fileModifiedAt: modify time in actual UTC, fallback to upload time
-    file_modified_at = to_actual_utc(metadata_modified_dt) or gumnut_asset.updated_at
-    # localDateTime: capture time in keepLocalTime format, fallback to upload time
+    # fileCreatedAt: capture time in actual UTC.
+    # Cascade mirrors Gumnut's REST Asset.local_datetime resolution:
+    # metadata.original_datetime → asset.file_created_at → asset.created_at.
+    # Falling all the way to created_at (upload time) makes assets without
+    # EXIF capture-date show "today" in the Immich timeline.
+    file_created_at = (
+        to_actual_utc(metadata_original_dt)
+        or to_actual_utc(gumnut_asset.file_created_at)
+        or gumnut_asset.created_at
+    )
+    # fileModifiedAt: modify time in actual UTC; same cascade against the
+    # file-modified chain.
+    file_modified_at = (
+        to_actual_utc(metadata_modified_dt)
+        or to_actual_utc(gumnut_asset.file_modified_at)
+        or gumnut_asset.updated_at
+    )
+    # localDateTime: capture time in keepLocalTime format; cascades through
+    # the same capture-time fields as fileCreatedAt.
     local_date_time = (
-        to_immich_local_datetime(metadata_original_dt) or gumnut_asset.created_at
+        to_immich_local_datetime(metadata_original_dt)
+        or to_immich_local_datetime(gumnut_asset.file_created_at)
+        or gumnut_asset.created_at
     )
 
     width = gumnut_asset.width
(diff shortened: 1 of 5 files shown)


taggartgorman (approved):
LGTM — clean bug fix with explicit test coverage.

## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (2 Consider items below)
- [x] **Codebase value**: Change leaves the codebase better than it found.
      Rationale: Fixes a real, user-visible bug (assets without EXIF capture-date no longer show as "today" in the timeline) and the two converters here are now consistent with each other. Smaller-scope "do nothing" was rejected because the bug is reproducible from any non-EXIF upload path; using `asset.local_datetime` directly was a viable alternative (see Consider #2) but the explicit cascade pays off in symmetry with `fileModifiedAt`, which has no pre-computed equivalent.
- [x] **Security/auth**: PR does not touch authentication, authorization, or permission logic.
      Evidence: Diff is limited to `routers/utils/asset_conversion.py` (pure data conversion) and unit-test files; no changes to auth middleware, token handling, or permission checks.
- [x] **Data deletion**: PR does not touch destructive operations or data removal.
      Evidence: Reviewed full diff; only datetime precedence logic and test-fixture None-defaults are modified.
- [x] **Confidence**: Correctness is well understood.
      Rationale: `or`-cascade short-circuits on falsy values; `to_actual_utc(None)` returns `None`, so the chain ordering is correct. Both converters now have identical resolution order. New `TestDateCascade` class covers metadata-wins, file-fallback, and full-fallback paths against both converters; CI green across lint / type / tests / compat.
- [x] **Purpose clarity**: PR purpose and approach are clear from description and code.

## Non-blocking suggestions

**[Consider] Could `fileCreatedAt`/`localDateTime` read `asset.local_datetime` directly?**

`asset.local_datetime` already encapsulates the same `metadata.original_datetime → file_created_at → created_at` cascade on the photos-api side (the existing sync-stream converter at `routers/api/sync/converters.py:141` uses it that way: `fileCreatedAt = to_actual_utc(asset.local_datetime)`). The two converters touched in this PR could collapse to that one call and skip mirroring the cascade locally.

The trade-off is real, though: `fileModifiedAt` has no `asset.local_datetime`-equivalent server-side field and still needs an explicit cascade. Keeping `fileCreatedAt`/`localDateTime` explicit preserves symmetry with the `fileModifiedAt` chain and makes the resolution order visible at the call site instead of relying on the server-side semantics being identical. Defensible to keep as-is; raising in case a follow-up wants to unify across the three converters.

**[Consider] PR description references private-repo paths**

`AGENTS.md` calls out: "This repo is public. Do not reference private sibling repos … from committed files or PR bodies — links resolve to dead ends for external readers and surface the existence of internal docs."

The body references `photos-api/scripts/seed_demo_data.py` and `gumnut-ai/photos#672`. Inlining the behavior ("a seed-script workaround that PATCHes original_datetime post-create can be reverted once this lands") would keep the rationale self-contained for external readers. Past PRs (#164) have merged with similar refs, so this appears to be loosely enforced — flagging because the rule is explicit.

PR #173, opened 2026-05-15
Stop sync v1 converter leaking 0 width/height to mobile

taggartgorman:
## Summary

- `gumnut_asset_to_sync_asset_v1` now coerces 0→None on `SyncAssetV1.width/height`, mirroring the `if x else None` convention already used in `build_asset_upload_ready_payload` and `convert_gumnut_asset_to_immich`. This is the load-bearing field for the Immich mobile asset viewer's `BoxConstraints` math — `RemoteAssetEntity.width/height` is sourced from this top-level row, not the EXIF subobject.
- Hardens `exif_dims_and_orientation` with the same rule for the EXIF wire fields: truthy checks on both `metadata.raw_width/raw_height` and the `asset.width/height` fallback, plus a final `(None, None, None)` so the helper can no longer surface 0 on either REST or sync. Defense in depth.
- Tests cover the videos-without-EXIF cohort end-to-end: `SyncAssetV1.width/height` emits None, and `ExifResponseDto`/`SyncAssetExifV1` emit None.
- Documents the "0 means unknown" rule in `docs/references/code-practices.md` so new converters don't re-introduce the leak.

Why this hits mobile but not web: photos-api stores `0` (not `NULL`) for assets it couldn't probe — mostly videos without EXIF width/height tags. The mobile asset viewer divides width by height to size its viewport and only guards against `null`. `0 / 0 → NaN → BoxConstraints crash` on tap. The web viewer reads the same field but does not perform that division.

## Linear

https://linear.app/gumnut-ai/issue/GUM-830/immich-adapter-sync-v1-converter-leaks-0-widthheight-to-mobile

## Test plan

- [x] `uv run ruff check` clean
- [x] `uv run pyright` clean
- [x] `uv run pytest` — 899 passing (40 in the touched files)
- [x] After deploy, find a video that was returning `width: 0`/`height: 0` on the sync wire; re-sync from mobile; confirm the viewer opens (it may still fail to play, but must not crash)
- [x] Spot-check a known-good photo to confirm valid dims still pass through


taggartgorman (commented):
## Self-review (auto-approvable)

Reviewing my own branch — GitHub won't accept a self-approval, so logging the review-pr outcome here as a comment.

### Pre-approve checklist
      Rationale: Aligns the third-of-three converters with the existing 0→None convention (`build_asset_upload_ready_payload`, `convert_gumnut_asset_to_immich`), so `SyncAssetV1` no longer ships the only inconsistent shape. The `exif_dims_and_orientation` hardening also closes a latent secondary path where `raw_w=0, raw_h=0, orientation=6` would have emitted `(0, 0, "6")` on the EXIF wire — the same double-rotation hazard the helper docstring warns against. Considered alternatives: (1) fix only the sync converter — rejected because the EXIF wire could still leak 0 on the REST flow when metadata is None and asset.width/height are 0; (2) extract a shared `coerce_zero_to_none` helper — rejected as premature for three one-liner sites that already share a clear convention; (3) do nothing and patch mobile-side — rejected because the adapter is the right place to enforce "0 means unknown" semantics for every Immich client.
      Evidence: Scanned all 5 changed files — `converters.py` (pure conversion), `asset_conversion.py` (pure conversion), `code-practices.md` (docs), two test files. No auth/permission code paths.
      Evidence: No deletion paths, no DB writes, no trash/restore handlers touched. `trashed_at` plumbing in the same file is unmodified.
      Rationale: Change is three falsy guards plus a control-flow rearrangement in one helper, all covered by new tests on both the sync and REST wires (`test_sync_asset_v1_emits_none_when_dims_zero`, `test_zero_raw_dims_fall_back_to_asset_dims`, `test_zero_dims_everywhere_emit_none`).
- [x] **Purpose clarity**: PR description names the cohort (videos without EXIF dims), the consumer that crashes (mobile asset viewer's `BoxConstraints` math), and why the web viewer is unaffected (it doesn't divide). Matches the diff exactly.

Outcome: auto-approvable. Will merge once a second pair of eyes signs off, or proceed on the test-plan validation steps.

ternarybits (approved):

PR #174, opened 2026-05-15
Align Immich asset date resolution

ternarybits:
## What
- Added shared date helpers in `routers/utils/asset_conversion.py` so REST, WebSocket, sync, and timeline surfaces resolve Immich asset dates consistently.
- Capture-time helpers (`resolve_capture_datetime`, `resolve_file_created_at`, `resolve_local_date_time`) trust Photos API's resolved `asset.local_datetime` as the source of truth.
- The modify-time helper (`resolve_file_modified_at`) still applies a `metadata.modified_datetime → asset.file_modified_at` cascade locally, because `file_modified_at` is the raw file mtime — Photos API does not resolve a single modify-time field for us.
- Updated focused tests and documented the helper convention (with the capture-vs-modify asymmetry called out) in `docs/references/code-practices.md`.

## Why
- Different adapter endpoints previously mixed Photos API's resolved `local_datetime` with an adapter-side metadata/file/upload-time cascade, which made future date-resolution changes easy to apply inconsistently.
- Treating Photos API's resolved `local_datetime` as the source of truth for capture time keeps timeline, asset detail, search, albums, memories, sync, and WebSocket payloads aligned without duplicating the cascade in the adapter.
- The modify-time helper keeps the EXIF preference so `fileModifiedAt` doesn't regress to raw file mtime on the wire.

## Linear
https://linear.app/gumnut-ai/issue/GUM-825/immich-adapter-audit-and-align-date-field-resolution-across-all

## Test plan
- [x] `uv run ruff check`
- [x] `uv run pyright`
- [x] `uv run pytest`


routers/api/timeline.py
```diff
@@ -184,12 +187,12 @@ async def get_time_bucket(
 
     for asset in filtered_assets:
         asset_id = asset.id
-        created_at = asset.local_datetime
+        created_at = resolve_capture_datetime(asset) or asset.created_at
```
ternarybits: Should we fold the fallback to asset.created_at into resolve_capture_datetime?

routers/utils/asset_conversion.py
```diff
@@ -355,35 +356,19 @@ def convert_gumnut_asset_to_immich(
...
-        or updated_at_fallback
-    )
-    local_date_time = (
-        to_immich_local_datetime(metadata_original_dt)
-        or to_immich_local_datetime(gumnut_asset.file_created_at)
-        or created_at_fallback
-    )
+    file_created_at = resolve_file_created_at(gumnut_asset)
+    if file_created_at is None:
+        file_created_at = created_at_fallback
```
ternarybits: Should we fold these fallbacks into resolve_file_created_at, resolve_file_modified_at, etc.?

routers/api/timeline.py
ternarybits: Fixed - timeline now passes `asset.created_at` as the explicit fallback to `resolve_capture_datetime(asset, asset.created_at)`.

routers/utils/asset_conversion.py
ternarybits: Fixed - REST conversion now passes required fallbacks into `resolve_file_created_at`, `resolve_file_modified_at`, and `resolve_local_date_time`; overloads keep the non-null contract explicit for pyright.

routers/utils/asset_conversion.py
```diff
@@ -35,6 +35,49 @@ def normalize_rating(rating: float | int | None) -> int | None:
...
+    so adapter callers treat it as the single source of truth and must not
+    re-add a fallback chain here.
+    """
+    return gumnut_asset.local_datetime
+
+
+def resolve_file_created_at(gumnut_asset: AssetResponse) -> datetime:
+    """Return capture time formatted for Immich ``fileCreatedAt`` fields."""
+    file_created_at = to_actual_utc(resolve_capture_datetime(gumnut_asset))
+    assert file_created_at is not None
```
ternarybits: In production code, we usually use exceptions instead of assertions.

ternarybits: Also, in what case could this be None?

ternarybits:
Addressed this round in `6dab990`.

Auto-fixed (two-way door):
- **[Should Fix] `assert ... is not None` in the date helpers** (`routers/utils/asset_conversion.py:52,59,77`, ternarybits) — Verified the value cannot be None given the SDK's typed contract: `asset.local_datetime` and `asset.file_modified_at` are both non-Optional, and `to_actual_utc(non-None) → datetime`. The asserts existed only as type-narrowing for pyright. Replaced with `@overload` decorators on `to_actual_utc` / `to_immich_local_datetime` in `routers/utils/datetime_utils.py` so non-None in narrows to non-None out at the type level, then dropped all three asserts. Behavior unchanged. Same fix applied to all three helpers since the comment's reasoning applies to each.

Side fix:
- `docs/references/code-practices.md` — added a Python Style bullet on overloads-vs-asserts for type narrowing, pointing to the two helpers as the canonical pattern.

Skipped:
- None.

Verification:
- `uv run ruff check`
- `uv run pyright` (0 errors)
- `uv run pytest` (`895 passed`)

routers/utils/asset_conversion.py
ternarybits: Fixed in `6dab990`. The value can never actually be None — `asset.local_datetime` is non-Optional in the SDK, so `to_actual_utc(...)` always returns a datetime here. The assert was only narrowing the helper's `datetime | None` return for pyright. Added `@overload` decorators to `to_actual_utc` and `to_immich_local_datetime` so non-None in → non-None out at the type level, then dropped all three asserts (same pattern at L59 and L77). Documented the overload-vs-assert convention in `docs/references/code-practices.md`.

routers/utils/asset_conversion.py
```diff
@@ -35,6 +35,43 @@ def normalize_rating(rating: float | int | None) -> int | None:
...
+    field for us — ``file_modified_at`` is the raw file mtime. The adapter
+    applies the ``metadata.modified_datetime → file_modified_at`` cascade
+    here so the EXIF modify time isn't lost on the wire.
+    """
+    metadata_modified = (
+        gumnut_asset.metadata.modified_datetime if gumnut_asset.metadata else None
+    )
+    return to_actual_utc(metadata_modified) or to_actual_utc(
+        gumnut_asset.file_modified_at
+    )
```
taggartgorman, line 72: Non-blocking, optional: this helper has three observable input shapes — (a) `metadata=None`, (b) `metadata` set but `modified_datetime=None`, (c) `metadata.modified_datetime` set. `TestDateResolution` covers (a) and (c); the (b) case is the realistic production shape when an asset has EXIF for camera/GPS but no `ModifyDate` tag. The two branches are functionally identical at the `to_actual_utc(None) or to_actual_utc(file_modified_at)` line, so risk is low — but a one-line addition (populate `metadata` with `original_datetime` set and `modified_datetime=None`, assert `fileModifiedAt == FILE_MODIFIED_DT`) would lock in the partial-metadata behavior explicitly. Feel free to skip if you'd rather not grow the test surface.

taggartgorman (approved):
LGTM — auto-approve.

This centralizes Immich asset date resolution behind four helpers (`resolve_capture_datetime`, `resolve_file_created_at`, `resolve_local_date_time`, `resolve_file_modified_at`) and routes all four outbound converters (REST asset, upload-ready WebSocket payload, sync v1, timeline buckets) through them. The capture/modify asymmetry — trust Photos API's resolved `local_datetime` for capture time, keep the local `metadata.modified_datetime → file_modified_at` cascade for modify time — is justified and explained both in the helper docstrings and in `docs/references/code-practices.md`. Net -29 LoC; previously-divergent cascades in three converters collapse into one source of truth.

Removing the `or datetime.now(timezone.utc)` fallbacks for `createdAt`/`updatedAt` is type-correct (the SDK declares those fields non-Optional). The two deleted unit tests were exercising SDK-impossible inputs, so removing them is consistent.

### Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (1 non-blocking Consider — see line comment)
- [x] **Codebase value**: After-state has one helper layer instead of three divergent inline cascades; named after the wire fields they emit.
- [x] **Security/auth**: Not touched. Diff is pure outbound conversion plus datetime utility overloads plus tests plus one docs page.
- [x] **Data deletion**: Not touched. No write paths, no trash/restore logic, no DB code.
- [x] **Confidence**: All four outbound converter call sites migrated; SDK type contract verified; CI green on lint, type check, tests, version sync.
- [x] **Purpose clarity**: PR description, helper docstrings, and docs page all explain the asymmetry consistently.

PR #176, opened 2026-05-18
Wire single-asset metadata edit endpoint to photos-api

taggartgorman:
## Summary

- Replaces the no-op `PUT /api/assets/{id}` stub with a real call to the Photos API `update_asset` PATCH; forwards `description`, paired `latitude` + `longitude`, and `dateTimeOriginal`. Out-of-scope DTO fields (`isFavorite`, `rating`, `visibility`, `livePhotoVideoId`) are silently ignored — request still succeeds, the adapter just doesn't act on parts the Photos API doesn't model. Bulk `PUT /api/assets` remains stubbed pending a backend bulk contract.
- Adds `WebSocketEvent.ASSET_UPDATE = "on_asset_update"` and emits it after a successful PATCH carrying the full converted `AssetResponseDto` — matches upstream Immich's emission shape and the web client signature `(asset: AssetResponseDto) => void`, so the web timeline manager can upsert on receipt.
- Adapter-side 422 validation: paired lat/lon (half-set or half-cleared) and invalid `dateTimeOriginal` short-circuit before the network call. Uses `model_fields_set` (Pydantic v2) to distinguish "field omitted" from "field explicitly null" — necessary because the DTO defaults every field to `None`.
- SDK floor bumped from `>=0.90.0` to `>=0.92.0` (resolves to 0.95.0). The backend PATCH returns the freshly-merged asset, so single-asset edits drop from 2 round-trips to 1.
- Docs synced: `adapter-architecture.md` Assets row, `websocket-implementation.md` (event moved from "Not Applicable" → Phase 1 supported), `websocket-events-reference.md` (trigger now disambiguates upstream sidecar processing vs. adapter direct edits). `code-practices.md` gains a new "Omit vs explicit-null" section documenting the `model_fields_set` pattern, and the "Implementing New Endpoints" checklist now includes the two websocket docs when a new event is emitted.

## Known gap (out of scope — tracked separately)

The Immich **mobile** client sends date, location, favorite, and visibility edits through the **bulk** endpoint `PUT /api/assets` (with `ids: [singleId]` even for single-asset edits) — confirmed in `immich/mobile/lib/repositories/asset_api.repository.dart`. This PR wires the single-asset endpoint only. The bulk endpoint stays a no-op stub until the backend bulk contract is settled.

Concretely, with this PR:

| Field | Web | Mobile |
|------|-----|--------|
| Description | ✅ single endpoint | ✅ single endpoint |
| Date / time | ✅ single asset; ❌ multi-select (bulk) | ❌ bulk |
| Location (lat/lon) | ✅ single asset; ❌ multi-select (bulk) | ❌ bulk |
| Rating | (out of scope) | ✅ single endpoint, ignored adapter-side |
| Favorite / visibility | (out of scope) | ❌ bulk |

Web users editing a single asset are fully covered. Mobile users editing description are covered; mobile users editing date/location see the UI update optimistically but the change does not persist. This matches what the issue body carved out explicitly. The user-visible mobile impact is broader than "multi-select flows" — flagged on the separate bulk ticket.

## Test plan

- [x] `uv run ruff format && uv run ruff check && uv run pyright && uv run pytest` — all green (915 passed)
- [x] CI passing
- New unit tests cover: each in-scope field round-trip; description clear via null; description="" forwarded (backend rejects); paired lat/lon validation in both directions; both-null clears coords; datetime parsing (with offset / Z / naive / invalid → 422 / null → clear); empty payload skips the SDK; out-of-scope-only payload skips the SDK; mixed in/out-of-scope payload sends only in-scope kwargs; `ASSET_UPDATE` websocket emit carries the full DTO.
- [x] Immich web single-asset "Edit date and time" modal changes the date and the photo moves in the timeline. (verified)
- [ ] Immich web description field on the asset detail panel persists across reloads.
- [ ] Immich web map drag / location-edit flow (single asset) updates lat/lon.
- [ ] Immich mobile "edit description" — works (single endpoint).
- [ ] Immich mobile "edit date" / "edit location" — known no-op until bulk endpoint lands; not gated by this PR.

### DIFF
--- a/routers/api/assets.py
+++ b/routers/api/assets.py
@@ -1,5 +1,5 @@
 from itertools import batched
-from typing import List, Literal, NamedTuple, cast
+from typing import Any, List, Literal, NamedTuple, cast
 from uuid import UUID, uuid4
 import base64
 import logging
@@ -752,6 +752,68 @@ async def copy_asset(
     return Response(status_code=status.HTTP_204_NO_CONTENT)
 
 
+def _parse_update_original_datetime(value: str | None) -> datetime | None:
+    """Parse the `dateTimeOriginal` field on an `UpdateAssetDto`.
+
+    Distinct from `_parse_datetime` (which substitutes a fallback on parse
+    failure for upload paths) — here an invalid input must surface as 422.
+    A `None` value means "clear" and is passed through.
+    """
+    if value is None:
+        return None
+    try:
+        return datetime.fromisoformat(value.replace("Z", "+00:00"))
+    except (ValueError, AttributeError) as exc:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail=f"Invalid dateTimeOriginal: {value!r}",
+        ) from exc
+
+
+def _build_metadata_patch(request: UpdateAssetDto) -> dict[str, Any] | None:
+    """Build the SDK `update_asset` kwargs from an `UpdateAssetDto`.
+
+    Uses `model_fields_set` to distinguish "field omitted" from "field
+    explicitly null" — both look like `None` on the model because the DTO
+    defaults every field to `None`. Out-of-scope DTO fields
+    (`isFavorite`, `rating`, `visibility`, `livePhotoVideoId`) are silently
+    ignored; the request still succeeds, we just don't act on parts the
+    Photos API doesn't model. Returns `None` when no in-scope fields were
+    set, signalling the caller to skip the PATCH entirely.
+
+    Validates paired lat/lon adapter-side so the request 422s before the
+    network call when the client sends half-set or half-cleared coords.
+    """
+    provided = request.model_fields_set
+    patch: dict[str, Any] = {}
+
+    if "description" in provided:
+        patch["description"] = request.description
+
+    lat_set = "latitude" in provided
+    lon_set = "longitude" in provided
+    if lat_set != lon_set:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail="latitude and longitude must be provided together",
+        )
+    if lat_set and lon_set:
+        if (request.latitude is None) != (request.longitude is None):
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="latitude and longitude must be cleared together",
+            )
+        patch["latitude"] = request.latitude
+        patch["longitude"] = request.longitude
+
+    if "dateTimeOriginal" in provided:
+        patch["original_datetime"] = _parse_update_original_datetime(
+            request.dateTimeOriginal
+        )
+
+    return patch or None
+
+
 @router.put("/{id}")
 async def update_asset(
     id: UUID,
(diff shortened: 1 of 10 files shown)


routers/api/assets.py
```diff
@@ -752,19 +752,96 @@ async def copy_asset(
...
+    payload = _build_metadata_patch(request)
+    if payload is None:
+        return await get_asset_info(id, client=client, current_user=current_user)
+
+    gumnut_asset = await client.assets.update_asset(
+        uuid_to_gumnut_asset_id(id), **payload
+    )
+    immich_asset = convert_gumnut_asset_to_immich(gumnut_asset, current_user)
+    await emit_user_event(
+        WebSocketEvent.ASSET_UPDATE, str(current_user.id), immich_asset
```
taggartgorman: `str(current_user.id)` is redundant — `UserResponseDto.id` is typed `str` (`routers/immich_models.py:2902`). The other `emit_user_event` call sites in this same file pass `current_user.id` directly (lines 279, 284). The places that *do* stringify are working from a `UUID` (`current_user_id: UUID` in trash/delete) — that doesn't apply here. Suggest dropping the `str(...)` to match the local convention.

taggartgorman: Fixed in 905e4e7 — dropped the `str(...)` and updated the test assertion to match.

ternarybits (approved):
## Pre-approve checklist
      Rationale: Converts a no-op stub into a real PATCH (web users editing description / date / location now persist). The `model_fields_set` decomposition gets generalized into a reusable `code-practices.md` section. Alternatives considered: (a) reject out-of-scope fields with 400 — rejected because Immich clients legitimately send them and the bulk endpoint is the right place to handle favorite/visibility; (b) emit `on_asset_update` for the out-of-scope-only path — rejected because nothing actually changed and busting the client cache would be wrong.
      Evidence: `routers/api/assets.py` update_asset still uses the existing `get_authenticated_gumnut_client` / `get_current_user` deps; no changes to `routers/utils/current_user.py`, auth middleware, or session handling.
      Evidence: `_build_metadata_patch` only writes (description / latitude / longitude / original_datetime); no DELETE paths altered.
      Rationale: 18 new unit tests cover every in-scope field round-trip, paired-lat/lon XOR in both directions, half-cleared coords, three datetime serializations (offset / Z / naive / invalid → 422 / null → clear), empty-payload skip, out-of-scope-only skip, mixed in/out-of-scope, and the WS emit payload shape. SDK `update_asset` signature confirmed to accept the kwargs we pass.
      Rationale: PR body lays out scope (in-scope vs. out-of-scope fields), known mobile gap (bulk endpoint not wired), and a per-field web/mobile coverage table.

## Summary

Replaces the no-op `PUT /api/assets/{id}` stub with a real photos-api PATCH, adds `on_asset_update` WebSocket emission with the full `AssetResponseDto` payload, and adds adapter-side 422 validation for paired lat/lon and invalid `dateTimeOriginal`. The `model_fields_set` pattern for distinguishing omit vs explicit-null is cleanly factored into `_build_metadata_patch` and documented in `code-practices.md` for reuse by future update endpoints. Docs synced (adapter-architecture, both websocket docs). All 4 CI checks pass.

PR #179, opened 2026-05-20
Resolve _image asset_urls keys for video thumbnail variants

taggartgorman:
## Summary

- The thumbnail/preview/fullsize streaming path in `routers/api/assets.py` (`_retrieve_and_stream_variant`, used by `GET /api/assets/{id}/thumbnail?size=…`) now branches on `asset.mime_type` when resolving the `asset_urls` key. For video assets, the three still-image sizes resolve to the `_image`-suffixed keys (`thumbnail_image`, `preview_image`, `fullsize_image`) that the upstream Photos API emits once a video has an extracted still. Image assets and the hard-coded `"original"` paths used by `/api/assets/{id}/original` and `/api/assets/{id}/video/playback` are unaffected.
- The mapping is owned by a small helper `_resolve_variant_key(mime_type, variant)`. The resolved key is used for the existence check, the `asset_urls` lookup, the warning log, and the 404 detail — so the existing `404 "Asset variant '<key>' not available"` path stays intact for the new variant names when extraction is pending, terminally failed, or pre-backfill.
- `AssetResponseDto` has no `thumbnailPath` field in this adapter, and Immich clients drive thumbnail UX off the dedicated `/{id}/thumbnail` endpoint, so the asset-conversion path (`convert_gumnut_asset_to_immich`) is unchanged.
- `gumnut-sdk` floor bumped from `>=0.92.0` to `>=0.97.0` (lockfile resolves to `0.97.0`, the first SDK release published after the upstream Photos API change). Hygiene only — `AssetResponse.asset_urls` is typed `Optional[Dict[str, AssetVariant]]`, so the new keys already flow through at runtime regardless of the floor.
- Round-2 docs: extended this repo's public-repo hygiene rule (`AGENTS.md`) and the Mock-attribute audit bullet in `docs/references/code-practices.md` to cover new branching predicates on previously-unbranched SDK attributes — the failure mode this PR's mime-branching surfaced.

## Why

The upstream Photos API now emits `thumbnail_image` / `preview_image` / `fullsize_image` for videos with an extracted still; the un-suffixed names stay reserved for future use, so the adapter cannot just point the existing keys at the video stills. Without this change the adapter looks up the un-suffixed keys on video assets and returns 404 for every video thumbnail, even when a still is available.

## Deploy ordering

The upstream Photos API change must ship and be deployed before this adapter starts requesting `_image` variants. Existing videos uploaded before the upstream pipeline shipped have no `_image` variants yet — a separate backfill populates them over time; until it runs, video thumbnail requests will continue to 404 as they do today, which matches current behaviour.

## Test plan

- [x] `uv run ruff check && uv run ruff format --check && uv run pyright && uv run pytest` — all green (919 passed)
- Unit tests added in `tests/unit/api/test_assets.py`:
  - Parametrized: video + `size=thumbnail|preview|fullsize` resolves to `thumbnail_image|preview_image|fullsize_image` and streams from the matching CDN URL.
  - Pre-extraction video (only `original` in `asset_urls`) requesting `size=thumbnail` returns 404 with `thumbnail_image` in the detail.
- Round 2: `test_play_asset_video_success` and `test_play_asset_video_forwards_range_header` now pass `mime_type="video/mp4"` to the mock so the realistic call shape is exercised — protects against a future addition of `"original"` to the video-only variant set silently breaking playback.
- [ ] Smoke against staging once the upstream Photos API change is deployed: open the Immich web/mobile timeline and confirm video posters render for clips with an extracted still; clips without one continue to fall back as today.


taggartgorman (commented):
Code change itself is clean and well-scoped:

- The mime/variant resolution is centralized in one helper, so all three callers (`view_asset`, `download_asset`, `play_asset_video`) flow through the same point — no risk of one path drifting.
- `AssetResponse.mime_type` is a non-nullable `str` in the SDK schema, so `.startswith("video/")` is safe.
- `"original"` correctly bypasses the suffix for both image and video assets, which matches the contract for `/{id}/original` and `/{id}/video/playback`.
- Tests cover the three new positive cases and the pre-extraction 404 path.

Three non-code items below — one Should Fix on PR-body hygiene, one Consider on test coverage, one Consider as a line comment.

---

**[Should Fix] PR body cross-links to a private repo**

The "Deploy ordering" section ends with `Cross-link: gumnut-ai/photos#723`, but that repo is private. Per `AGENTS.md`:

> This repo is public. Do not reference private sibling repos … from committed files or PR bodies — links resolve to dead ends for external readers and surface the existence of internal docs. Inline the relevant rationale or context instead.

The surrounding prose ("photos-api now emits `thumbnail_image`/…", "must ship and be deployed before this adapter starts requesting `_image` variants") already inlines the necessary context, so dropping that single line is the whole fix.

---

**[Consider] No explicit `video/*` mime + `original`-variant test coverage**

`_make_mock_asset_with_urls` defaults `mime_type="image/jpeg"`, and `test_play_asset_video_success` / `test_play_asset_video_forwards_range_header` / `test_download_asset_success` all rely on that default even though their `asset_urls` payloads describe `video/mp4`. The current behavior is correct (`"original" not in _VIDEO_IMAGE_VARIANTS` short-circuits regardless of mime), so this is regression protection only — if someone later added `"original"` to `_VIDEO_IMAGE_VARIANTS`, video playback/download would silently break and these tests wouldn't catch it. Passing `mime_type="video/mp4"` to the mock in at least one of those tests would lock down the realistic call path.

routers/api/assets.py
```diff
@@ -89,6 +89,24 @@
     AssetMediaSize.thumbnail: "thumbnail",
 }
 
+# Variants that get an `_image` suffix for video assets. The un-suffixed
+# names are reserved for potential future animated previews (e.g. ABR HLS).
```
taggartgorman: [Consider] The second sentence — "The un-suffixed names are reserved for potential future animated previews (e.g. ABR HLS)." — documents a hypothetical future use rather than current behavior. Per prescriptive-doc hygiene, "reserved for future X" notes age poorly: if ABR HLS never lands or lands under a different name, the comment is misleading. The first sentence is sufficient — readers can already see from `_resolve_variant_key` that un-suffixed video keys aren't produced today. Optional: drop the second sentence, or rephrase as a statement of current behavior (e.g. "Un-suffixed video variant keys are not currently produced by photos-api.").

taggartgorman:
### Round 2 changes

**Auto-fixed (two-way door)**

- **[Should Fix] PR body referenced a private sibling repo.** Removed the `Cross-link: gumnut-ai/photos#723` line and rephrased the "Why" / "Deploy ordering" sections to describe the upstream Photos API change without naming the private repo. Surrounding prose already inlined the rationale, so no information was lost. Also extended `AGENTS.md` with two concrete violation examples (cross-link lines, private file paths) plus guidance not to copy Linear/design-doc cross-link framing verbatim, since that's how the violation slipped in.
- **[Consider] Mock didn't pin the realistic video MIME for the `play_asset_video` tests.** Passed `mime_type="video/mp4"` to `_make_mock_asset_with_urls` in `test_play_asset_video_success` and `test_play_asset_video_forwards_range_header`. The current behaviour is unchanged (`"original" not in _VIDEO_IMAGE_VARIANTS` short-circuits regardless of MIME), but the test now fails closed if someone later adds `"original"` to the video-only variant set. Also extended `docs/references/code-practices.md`: the existing "audit Mock fixtures when adding an SDK field" rule now explicitly covers new branching predicates on previously-unbranched attributes (the failure mode this round surfaced).
- **[Consider] Forward-looking ABR HLS comment on `_VIDEO_IMAGE_VARIANTS`.** Dropped the second sentence. The first sentence (and the code right next to it) is sufficient; the design-doc-style rationale doesn't belong in the source comment.

Full self-review run on the updated branch (3 dimensions × 3 subagents) returned zero new findings. Lint, format, type check, and the full 919-test suite all pass.

routers/api/assets.py
taggartgorman: Fixed in b079ff2 — dropped the second sentence. Agree on prescriptive-doc hygiene; the first sentence plus the adjacent `_resolve_variant_key` are enough.

ternarybits (approved):

PR #181, opened 2026-05-21
Wire bulk asset-metadata edit endpoint via SDK bulk_update_assets

taggartgorman:
## What
- Replace the `PUT /api/assets` 204 stub with a real bulk metadata edit via the typed `client.assets.bulk_update_assets` call (gumnut-sdk 0.99.0). In-scope fields: `description`, paired `latitude` + `longitude`, and the capture time. Requests are chunked over `BULK_CHUNK_SIZE` (100).
- Capture time is set via one of **three mutually exclusive datetime modes**:
  - **absolute `dateTimeOriginal`** (optionally localized by a paired `timeZone`) — identical for every id, replicated as one homogeneous `change`; no read.
  - **`dateTimeRelative`** — a per-asset second-shift.
  - **standalone `timeZone`** — reinterpret each asset's existing wall-clock in the given zone (preserve the clock digits, swap the tzinfo).
- The two per-asset modes need each asset's *current* `original_datetime`, so the handler reads the chunk first (`client.assets.list(state="all", ids=...)`, including trashed assets so they aren't silently skipped — symmetric with the homogeneous path, which forwards every id to the bulk write regardless of trash state) and writes a heterogeneous per-item `change`. This is a **bulk GET + bulk PATCH** per chunk — one read and one write — not a per-asset GET fan-out. Assets with no existing capture time (and ids not returned by the read) are skipped for the datetime rewrite; if they carry no other in-scope field they drop out of the write entirely. There is a small read-then-write window in which an asset's datetime could change between the two calls.
- Combining datetime modes (`dateTimeRelative` + `dateTimeOriginal`, or `dateTimeRelative` + `timeZone`) is rejected with 422 — the intended result would be ambiguous. Silently ignore `isFavorite` / `rating` / `visibility` / `duplicateId` to match the single-asset endpoint's pattern.
- Explicit `null` for `dateTimeRelative` and `timeZone` is treated as "field not set" (the per-asset rewrites only trigger on non-null values), mirroring the wire shape some Immich client SDKs emit for untouched fields.
- Map malformed IANA timezone keys (empty string, absolute paths, non-normalized paths) to 422 — `ZoneInfo()` raises both `ZoneInfoNotFoundError` (unknown but well-formed zones) and plain `ValueError` (malformed keys), so the handler catches both. Resolution is eager, so a bad standalone-`timeZone` name 422s before any read.
- The two per-asset datetime modes are modeled as a tagged union (`_RelativeShift` / `_ReinterpretTimezone`) so each mode's field is non-optional and the "exactly one mode" invariant lives in the type rather than a runtime `assert`. `_build_bulk_metadata_change` returns a named `_BulkMetadataChange(base, transform)`.
- Document the cross-chunk atomicity caveat: the SDK's `bulk_update_assets` is per-call atomic, but the adapter's sequential chunked loop does not preserve that guarantee across chunks. A failure on chunk N (N ≥ 2) leaves chunks 1..N-1 already committed and propagates as one 5xx. Captured in both the handler docstring and `docs/references/code-practices.md` so all future bulk paths share the same caveat shape.
- Bump `gumnut-sdk` from `>=0.97.0` to `>=0.99.0`; refresh `uv.lock`.
- Document the bulk-write WebSocket emission default, and the "reads feeding a bulk write use `state="all"`" convention, in `docs/references/code-practices.md`.

## Why
- Immich's web "Edit date and time" / location / description bulk-edit modals, and the mobile `updateDateTime` / `updateLocation` / `updateFavorite` / `updateVisibility` flows (which use this endpoint even for single-asset edits), have been silently no-oping against the 204 stub. Wiring the endpoint closes a broad mobile/web feature gap.
- `dateTimeRelative` and standalone `timeZone` were initially scoped out (rejected with 422) on the assumption that translating them required a per-asset GET. They don't: `list_assets(ids=...)` returns each asset's current `original_datetime` in one bulk read, and `bulk_update_assets` already accepts heterogeneous per-item changes — so both modes are one read + one write per chunk. Implementing them covers the full Immich bulk-edit surface rather than surfacing a 422 for relative shifts and timezone-only edits.
- The per-asset read uses `state="all"` so trashed assets aren't silently dropped: the homogeneous path forwards every id to `bulk_update_assets` regardless of trash state, so the read-driven per-asset modes must too — otherwise the same selection would update trashed assets via absolute-date/description/lat-lon but skip them via `dateTimeRelative` / standalone `timeZone`. Mirrors sync hydration's id-keyed read.
- No WebSocket emission from the bulk path: `bulk_update_assets` returns no per-asset payload, so mirroring the single-asset `on_asset_update` shape would require a *post-update* re-fetch per chunk (the per-asset read above is pre-update state, for the datetime modes only). Mobile triggers a sync refresh on its own and web has optimistic UI for these flows; the round-trip isn't justified.

## Test plan
- [x] `uv run ruff format && uv run ruff check && uv run pyright && uv run pytest` all green
- [x] Unit tests for `tests/unit/api/test_assets.py` cover: homogeneous description / lat-lon / absolute-datetime round-trips, chunking boundaries (100 / 101 / 205), per-asset `dateTimeRelative` shift and standalone-`timeZone` reinterpret, skip-on-null capture time (with and without an accompanying homogeneous field), the three datetime-mode 422 conflicts, malformed-tz 422 covering both `ZoneInfoNotFoundError` and `ValueError`, eager standalone-tz 422 (no read on bad name), null-as-omitted for `dateTimeRelative` / `timeZone`, `state="all"` on the per-asset read (trashed assets still rewritten), the zero-writable-chunk skip, multi-chunk read+write ordering, SDK error propagation, and cross-chunk no-rollback partial-commit
- [ ] Verify Immich web multi-select "Edit date and time" / location / description flows in a manual session
- [ ] Verify Immich mobile single-asset date and location edits (which hit this bulk endpoint)


taggartgorman (commented):
Self-review of the bulk metadata edit wiring. Three findings — one real bug (timeZone ValueError → 500), and two non-blocking notes (a docstring caveat about multi-chunk atomicity, and a small test-coverage gap pinning the documented null-handling for `dateTimeRelative` / `timeZone`).

routers/api/assets.py
```diff
@@ -598,16 +600,144 @@ async def _upload_streaming(
...
+    The Immich bulk DTO carries `timeZone` as an IANA name (e.g.
+    `America/Los_Angeles`). The Photos API encodes the offset directly into
+    `original_datetime`, so we localize wall-clock here. Aware inputs are
+    re-anchored: the wall-clock components are preserved and re-tagged with
+    the new tz, matching Immich's "interpret these clock digits in this zone"
+    UX (the modal sends a naive `dateTimeOriginal` when `timeZone` is set,
+    but we handle aware too for safety).
+    """
+    try:
+        tz = ZoneInfo(tz_name)
```
taggartgorman: `_combine_datetime_with_timezone` only catches `ZoneInfoNotFoundError`, but `ZoneInfo()` also raises **`ValueError`** for malformed keys — e.g. `""`, leading-slash paths, or up-level references like `"../etc/passwd"`. The DTO declares `timeZone: str | None` with no constraint, so these reach `ZoneInfo()` unchanged and bubble out as an uncaught 500 instead of the 422 the rest of this branch uses.

The existing `test_update_assets_invalid_timezone_is_422` uses `"Mars/Olympus_Mons"`, which is a valid-format path that triggers `ZoneInfoNotFoundError` — the malformed-key branch isn't exercised.

Suggested fix: broaden the except to `(ZoneInfoNotFoundError, ValueError)` and add a parametrized regression test covering `""` and a non-normalized path so we don't silently slip back to 500.

routers/api/assets.py
```diff
@@ -598,16 +600,144 @@ async def _upload_streaming(
...
 async def update_assets(
     request: AssetBulkUpdateDto,
     client: AsyncGumnut = Depends(get_authenticated_gumnut_client),
-):
-    """
-    Update asset metadata.
-    This is a stub implementation as Gumnut does not support asset metadata updates.
-    Returns HTTP 204 (No Content) as specified by the Immich API.
+) -> Response:
+    """Bulk-update asset metadata.
```
taggartgorman, line 843: The SDK markets `bulk_update_assets` as atomic per call, but the adapter chunks at `BULK_CHUNK_SIZE` and calls sequentially — so a failure on chunk N≥2 leaves chunks 1..N-1 already committed with no compensating action and no per-chunk error report (the exception propagates as a single 5xx). The docstring mentions chunking but not the atomicity downgrade.

Consistent with `_bulk_permanent_delete` / `_bulk_trash`, but the SDK's explicit atomicity claim makes the gap more confusing here — easy for a future reader to assume the guarantee transitively holds. One sentence in the docstring would close that, and the same caveat could optionally land in `code-practices.md` so all bulk paths share it uniformly.

tests/unit/api/test_assets.py
```diff
@@ -985,21 +987,440 @@ class _State:
...
+        request = AssetBulkUpdateDto(ids=[uuid4()], timeZone="America/Los_Angeles")
+
+        with pytest.raises(HTTPException) as exc_info:
+            await update_assets(request, client=mock_client)
+
+        assert exc_info.value.status_code == 422
+        mock_client.assets.bulk_update_assets.assert_not_awaited()
+
+    @pytest.mark.anyio
+    async def test_update_assets_relative_datetime_is_422(self):
```
taggartgorman: The `_build_bulk_metadata_change` docstring explicitly says explicit `null` for `dateTimeRelative` or standalone `timeZone` is treated as "field not set" rather than 422 (the rejection only triggers when non-null). Both non-null-raises branches are pinned by `test_update_assets_relative_datetime_is_422` and `test_update_assets_timezone_alone_is_422`, but neither null-is-silently-ignored branch is.

A future refactor that changed `is not None` to a truthy check (or dropped the `model_fields_set` membership check) would silently start 422-ing clients that send `dateTimeRelative: null` / `timeZone: null` — a wire shape some Immich client SDKs emit. Two short tests would pin the contract:

```python
# model_validate({"ids": [...], "dateTimeRelative": None, "description": "x"}) → 204 with {"description": "x"}
# model_validate({"ids": [...], "timeZone": None, "description": "x"})         → 204 with {"description": "x"}
```

taggartgorman:
**Round 1 — self-review feedback addressed.** Pushed bb6d310 (rebased onto current `main`).

### Auto-fixed (two-way)

- **`ValueError` → 500 in `_combine_datetime_with_timezone`** (`routers/api/assets.py:595`). `ZoneInfo()` raises bare `ValueError` for malformed keys (empty string, `/Etc/UTC`, `../etc/passwd`), not the `ZoneInfoNotFoundError` the handler caught. Broadened the except to `(ZoneInfoNotFoundError, ValueError)`; parametrized `test_update_assets_invalid_timezone_is_422` to cover all four input classes.
- **Cross-chunk atomicity caveat** (`routers/api/assets.py:705`, `docs/references/code-practices.md`). The SDK's `bulk_update_assets` is per-call atomic, but the adapter's sequential chunked loop is not. Added a one-sentence note to the `update_assets` docstring and a shared guidance block to `code-practices.md` so future bulk-write paths note the limitation uniformly. Also pinned the no-rollback contract with `test_update_assets_multi_chunk_failure_leaves_prior_chunks_committed` (chunk 1 succeeds, chunk 2 raises, `await_count == 2` asserts no retry / no rollback re-call) — surfaced during self-review as a parallel to the existing no-swallow-via-`test_*_propagates_sdk_error` pattern from `code-practices.md`.
- **Null-as-omitted contract pinned** (`tests/unit/api/test_assets.py`). Two tests: `dateTimeRelative: None` + `description: "x"` → 204 forwarding only description; `timeZone: None` + `description: "x"` → same. A future refactor swapping `is not None` for a truthy check would silently start 422-ing some Immich client SDKs.

### Self-review findings dismissed (rationale)

- *Non-null `timeZone` + explicit-null `dateTimeOriginal` corner.* Real edge but unreachable from real clients (the modal sends `dateTimeOriginal` + `timeZone` together, SDK-shape clients don't construct this combo). Adding validation or a test commits us to one of two defensible-but-arbitrary contracts for a scenario that hasn't occurred — premature.
- *Lat/lon paired validation duplicated between `_build_metadata_patch` and `_build_bulk_metadata_change`.* ~7 lines, intentional ("Mirrors `_build_metadata_patch`..." already in the bulk docstring). The two paths are diverging (bulk grew `dateTimeRelative` / standalone-`timeZone` validation, atomicity caveat). Extracting a helper now would constrain that divergence for marginal benefit.

routers/api/assets.py
taggartgorman: Fixed in bb6d310. Confirmed `ZoneInfo()` raises bare `ValueError` for `""`, `/Etc/UTC`, and `../etc/passwd`; only `Mars/Olympus_Mons` (well-formed but unknown) raises `ZoneInfoNotFoundError`. Broadened the except to `(ZoneInfoNotFoundError, ValueError)` and added an inline comment enumerating the `ValueError` subcases. Parametrized `test_update_assets_invalid_timezone_is_422` over all four input classes so both branches stay pinned.

routers/api/assets.py
taggartgorman: Fixed in bb6d310. Added a paragraph to the `update_assets` docstring spelling out that the SDK guarantees per-call atomicity but the chunked loop does not, and that a chunk-N (N ≥ 2) failure leaves chunks 1..N-1 committed with no compensating rollback. Also landed the same caveat as a shared guidance block in `docs/references/code-practices.md` ("Cross-chunk atomicity is not guaranteed") so future bulk-write paths share it uniformly. Pinned the no-rollback half of the contract with `test_update_assets_multi_chunk_failure_leaves_prior_chunks_committed` — chunk 1 succeeds, chunk 2 raises, `await_count == 2` asserts no retry/rollback re-call — parallel to the existing no-swallow `test_*_propagates_sdk_error` pattern that `code-practices.md` already prescribes.

tests/unit/api/test_assets.py
taggartgorman: Fixed in bb6d310. Added the two suggested tests: `test_update_assets_relative_datetime_null_is_ignored` and `test_update_assets_timezone_null_is_ignored`. Both use `AssetBulkUpdateDto.model_validate({...})` to force the field into `model_fields_set` with an explicit `None`, send a `description="x"` alongside, and assert 204 + bulk_update_assets called with only `{"description": "x"}`. A future refactor that flips `is not None` to a truthy check (or drops the `in provided` membership check) will now break these tests.

routers/api/assets.py
```diff
@@ -598,16 +600,152 @@ async def _upload_streaming(
...
+    Out-of-scope DTO fields (`isFavorite`, `rating`, `visibility`,
+    `duplicateId`) are silently ignored — the request still succeeds, the
+    adapter just doesn't act on parts the Photos API doesn't model.
+
+    Two Immich fields are rejected with 422 when set to a non-null value
+    rather than silently dropped (explicit `null` is treated as "field not
+    set", mirroring how `timeZone: null` is handled — clients sometimes
+    send null for fields they don't intend to change):
+
+    - `dateTimeRelative` — a per-asset second-shift that the Photos API
```
ternarybits: Wouldn't this just take on bulk GET and one bulk PATCH?

taggartgorman: Yes, you're correct. I had forgotten about being able to retrieve multiple assets by id with GET /api/assets.

Updated in b92ac4f to handle both these cases.

routers/api/assets.py
```diff
@@ -598,16 +600,235 @@ async def _upload_streaming(
...
+    def apply(self, current: datetime) -> datetime:
+        if self.relative_seconds is not None:
+            return current + timedelta(seconds=self.relative_seconds)
+        assert self.reinterpret_tz is not None
+        return current.replace(tzinfo=self.reinterpret_tz)
+
+
+def _build_bulk_metadata_change(
+    request: AssetBulkUpdateDto,
+) -> tuple[dict[str, Any], _PerAssetDatetime | None]:
```
ternarybits: Extract out `tuple[dict[str, Any]` or just `dict[str, Any]` into a named type for readability?

ternarybits (commented):
Solid work wiring the bulk endpoint — the three-mode datetime model, chunking, atomicity caveats, and test coverage are all thorough, and it mirrors the single-asset path's conventions well. Two `Should Fix` items and one non-blocking test suggestion below. CI is green and the timezone-key validation (incl. path-traversal inputs) is correctly guarded.

routers/api/assets.py
```diff
@@ -598,16 +600,235 @@ async def _upload_streaming(
...
+    for chunk in batched(request.ids, BULK_CHUNK_SIZE):
+        gumnut_ids = [uuid_to_gumnut_asset_id(uid) for uid in chunk]
+        updates: list[Update]
+
+        if transform is None:
+            sdk_change = cast(UpdateChange, base_change)
+            updates = [{"id": gid, "change": sdk_change} for gid in gumnut_ids]
+        else:
+            # Per-asset datetime rewrite: read current values, then write.
+            page = await client.assets.list(ids=gumnut_ids, limit=len(gumnut_ids))
```
ternarybits: **[Should Fix / verify]** This per-asset read uses the SDK's default live-only state filter, so any selected asset that isn't live (e.g. trashed) is absent from `page.data`, gets `None` from `current_by_id.get(gid)`, and is silently dropped from the datetime rewrite. This is asymmetric with the homogeneous path in this same handler, which issues no read and forwards every id straight to `bulk_update_assets` — so the same selection updates trashed assets via absolute-date/description/lat-lon but silently skips them via `dateTimeRelative` / standalone `timeZone`.

The codebase established exactly this fix for the same `assets.list(ids=...)` shape: `routers/api/sync/entity_fetch.py:56-63` passes `state="all"` precisely because "the default live-only filter would silently drop them from page.data," and `docs/design-docs/trash-soft-delete-adapter.md` documents it as the id-keyed read convention.

Can you confirm whether trashed assets can reach this path and whether the backend `bulk_update_assets` accepts trashed ids? If both, consider `client.assets.list(state="all", ids=gumnut_ids, limit=len(gumnut_ids))` plus a test with a non-live asset to pin that the per-asset modes still rewrite it.

routers/api/assets.py
```diff
@@ -598,16 +600,235 @@ async def _upload_streaming(
...
+      absolute path.
+    """
+
+    relative_seconds: float | None
+    reinterpret_tz: ZoneInfo | None
+
+    def apply(self, current: datetime) -> datetime:
+        if self.relative_seconds is not None:
+            return current + timedelta(seconds=self.relative_seconds)
+        assert self.reinterpret_tz is not None
```
ternarybits: **[Should Fix]** `assert self.reinterpret_tz is not None` is an `assert` in production code, which `docs/references/code-practices.md` calls out directly ("Type narrowing — overloads, not asserts ... Asserts are stripped under `python -O` ... For genuine runtime defense, use exceptions, not `assert`"). The rest of `routers/`/`services/`/`config/` has zero asserts. It can't fire given the current call sites, so this is a convention fix rather than a live bug — but under `-O` the stripped assert would let a `None` tz silently fall through to `.replace(tzinfo=None)`. Suggest an explicit `if self.reinterpret_tz is None: raise ...` or restructuring `_PerAssetDatetime` into a tagged shape so both fields aren't `X | None`.

routers/api/assets.py
```diff
@@ -598,16 +600,235 @@ async def _upload_streaming(
...
+            }
+            updates = []
+            for gid in gumnut_ids:
+                change = dict(base_change)
+                current = current_by_id.get(gid)
+                if current is not None:
+                    change["original_datetime"] = transform.apply(current)
+                if change:
+                    updates.append({"id": gid, "change": cast(UpdateChange, change)})
+            if not updates:
```
ternarybits, line 924: **[Consider]** The `if not updates: continue` branch (an entire chunk where every asset is non-writable in a per-asset datetime mode, with no homogeneous field) isn't directly covered — `test_update_assets_per_asset_skips_null_datetime` keeps one writable asset, so `updates` stays non-empty. Reachable via e.g. `AssetBulkUpdateDto(ids=[x], dateTimeRelative=60.0)` with the read returning `{x: None}`. Low impact, but a small test asserting `bulk_update_assets.assert_not_awaited()` for a zero-writable chunk would pin it.

routers/api/assets.py
taggartgorman: Fixed in bf34054. Confirmed the asymmetry: the homogeneous path forwards every id straight to `bulk_update_assets` regardless of trash state, so the adapter contract already assumes the backend accepts trashed ids there — the per-asset read just needed to match. Changed the read to `client.assets.list(state="all", ids=...)` (mirroring `entity_fetch.py`), and added `test_update_assets_per_asset_reads_all_states_incl_trashed` asserting the read kwargs include `state="all"` and the trashed asset is still rewritten.

routers/api/assets.py
taggartgorman: Fixed in bf34054. Replaced the two-`Optional` NamedTuple + `assert` with a tagged union of single-field NamedTuples (`_RelativeShift` / `_ReinterpretTimezone`). Each mode's field is now non-optional, so the "exactly one mode" invariant lives in the type — no `assert` to strip under `-O`, and no dead runtime guard.

routers/api/assets.py
taggartgorman: Fixed in bf34054. `_build_bulk_metadata_change` now returns a named `_BulkMetadataChange(base, transform)` NamedTuple instead of the inline `tuple[dict[str, Any], _PerAssetDatetime | None]`.

routers/api/assets.py
taggartgorman: Fixed in bf34054. Added `test_update_assets_per_asset_zero_writable_chunk_skips_write`: both assets have a null `original_datetime`, relative mode, no homogeneous field → empty `updates` → `bulk_update_assets.assert_not_awaited()`, pinning the `if not updates: continue` branch.

ternarybits (approved):

PR #185, opened 2026-05-26
Emit base64 SHA-1 checksum to Immich clients, never SHA-256

taggartgorman:
## What
- Add a `resolve_immich_checksum()` helper in `routers/utils/asset_conversion.py` that returns Gumnut's `checksum_sha1` (base64 SHA-1, the Immich-facing format); when it is null, the helper logs a WARNING with the asset id and returns `""`.
- Route all three outbound converters through it — REST (`convert_gumnut_asset_to_immich`), WebSocket (`build_asset_upload_ready_payload`), and sync stream (`gumnut_asset_to_sync_asset_v1`) — so they agree. Removes the sync converter's bespoke null-warning + SHA-256 fallback (and its now-unused module logger).
- Add tests asserting the SHA-1 is emitted at every site and that the SHA-256 / `"placeholder-checksum"` literal are never produced, plus that the null path logs a warning carrying the asset id. Backfill `checksum_sha1` on the test mocks that flow through the REST/WS converters.
- Document the convention in `docs/references/code-practices.md`.

## Why
Immich's `checksum` field is a base64-encoded **SHA-1** (28 chars): clients compute the SHA-1 of a local file and compare it to the server value for pre-upload dedup and for local↔remote asset linking ("merged" state) in the mobile client. The adapter was sending Gumnut's `checksum` field instead, which is a base64 **SHA-256** (44 chars). That is a format-level mismatch — a 44-char value can never equal the 28-char SHA-1 a client computes — so dedup silently never matched and a photo present both on-device and in Gumnut showed up as **two separate timeline entries**, never recognized as backed up. The REST and WebSocket converters emitted the SHA-256 outright (and a `"placeholder-checksum"` literal when empty); the sync converter preferred `checksum_sha1` but silently fell back to the SHA-256 when it was null.

Emitting `""` for the rare null-`checksum_sha1` case is deliberate: an empty value yields a clean "no match" (a dedup false-negative — the documented Immich behavior), whereas a wrong-format value looks valid but never matches. The inbound dedup path already keys on `checksum_sha1`, so this makes the two directions symmetric.

Scope is the adapter fix only. Backfilling the small cohort of historical rows with a null `checksum_sha1` is tracked separately as a server-side data operation.

## Test plan
- [x] `uv run pytest` — 926 passed
- [x] New `TestChecksumEmission` tests cover all three converters, the null→`""` path, and the null-path warning log.


routers/utils/asset_conversion.py
```diff
@@ -26,6 +27,37 @@
...
+    the client-computed SHA-1, so it silently breaks dedup and makes a
+    backed-up photo appear as two separate timeline entries.
+
+    When ``checksum_sha1`` is null (rare legacy rows), return ``""`` rather
+    than substituting the SHA-256 or a placeholder. An empty checksum yields a
+    clean "no match" (a dedup false-negative — the documented Immich behavior)
+    instead of a value that looks valid but never matches.
+    """
+    if gumnut_asset.checksum_sha1 is None:
+        logger.warning(
```
taggartgorman, line 53: **Consider (verify):** this WARNING fires once per asset with a null `checksum_sha1`, and the helper runs inside per-asset loops (sync stream, search/memories/albums list conversions). If the legacy null cohort is small this is exactly the observability signal you want; if it's large, a full sync could emit thousands of lines per pass. Worth confirming the null cohort is genuinely small in production — otherwise aggregating to a single per-request summary would keep it useful without the spam. (Note: this isn't a regression on the sync path, which already warned per-asset; it does newly warn on the REST/WS paths that were previously silent.)

taggartgorman (commented):
## Review: Approve ✅ (recommendation)

_Posted as a comment — GitHub blocks approving your own PR. No blocking findings; safe to merge._

Clean, well-scoped fix. The adapter was emitting Gumnut's base64 SHA-256 `checksum` on Immich's outbound `checksum` field, which clients expect to be base64 SHA-1 — a format mismatch that silently broke dedup and duplicated backed-up photos in the timeline. Routing all three outbound converters (REST / WebSocket / sync) through a single `resolve_immich_checksum()` helper is the right shape: it centralizes three divergent implementations, removes the sync converter's bespoke SHA-256 fallback, and makes the `null → ""` no-match decision explicit and tested.

Verified during review:
- The three converters are the **only** active outbound `checksum=` emit sites (`models.py`'s is commented-out dead code).
- Inbound `bulk_upload_check` keys on `checksum_sha1` and excludes falsy rows — symmetric with emitting `""` for null.
- The new `code-practices.md` section is accurate against the code and free of private-repo references.

No blocking findings. A few non-blocking Consider notes inline.

## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (3 Consider, inline + below)
- [x] **Codebase value**: After-state replaces three divergent checksum implementations with one helper, deletes a wrong-format fallback and an orphaned logger, and makes the null path observable. Beats "do nothing" (leaves dedup broken) and a smaller per-converter patch (would leave the three call sites parallel and able to drift again).
- [x] **Security/auth**: No auth/authz/permission logic touched. Checked changed files (`asset_conversion.py`, `converters.py`, tests, docs); no token/permission/session code.
- [x] **Data deletion**: No destructive operations. Change only affects the value emitted on an outbound read field.
- [x] **Confidence**: Correctness well understood — change limited to checksum field selection across three converters, covered by the new `TestChecksumEmission` suite (926 tests pass).
- [x] **Purpose clarity**: Purpose and approach clear from description and code.

### Consider (non-blocking)
- Some rationale duplication across the helper docstring, the doc section, the test-class docstring, and the repeated conftest comment. The helper docstring is the natural canonical owner; trimming the test-class docstring / dropping the duplicated conftest comment would reduce drift risk. Optional polish.

tests/unit/utils/test_asset_conversion.py
```diff
@@ -388,3 +390,87 @@ def test_zero_dims_everywhere_emit_none(self, sample_gumnut_asset):
...
+    SHA-1 (28 chars) a client computes, so emitting it silently breaks dedup
+    and duplicates backed-up photos in the timeline.
+    """
+
+    # 28-char base64 SHA-1 (the correct Immich wire value) vs. the SHA-256
+    # placeholder the fixture carries on ``.checksum``.
+    SHA1_B64 = "PaDX6+c+Lhjpm5/ciXUROL1ryaU="
+    OWNER_ID = "22222222-2222-2222-2222-222222222222"
+
+    def test_rest_converter_emits_sha1(self, sample_gumnut_asset, mock_current_user):
```
taggartgorman, line 407: **Consider:** the three `*_emits_sha1` happy-path tests assert the returned value but not that *no* WARNING is emitted. A one-line negative assertion here would lock in "no log spam on the common path" as a contract, so a future refactor that accidentally moved `logger.warning` out of the `is None` branch would fail a test:
```python
with caplog.at_level(logging.WARNING, logger="routers.utils.asset_conversion"):
    result = convert_gumnut_asset_to_immich(sample_gumnut_asset, mock_current_user)
assert not [r for r in caplog.records if r.levelno == logging.WARNING]
```

taggartgorman: Fixed — added `test_non_null_sha1_emits_no_warning`, which wraps the common (populated `checksum_sha1`) path in `caplog.at_level(logging.WARNING, ...)` and asserts no WARNING records, locking in the no-spam contract.

routers/utils/asset_conversion.py
taggartgorman: Skipping the aggregation change — confirmed the null cohort is genuinely small: production has 373 / 21,276 assets (~1.75%) with null checksum_sha1, and that is the entire cohort, shrinking toward 0 once the separately-tracked backfill runs. A full sync emits at most a few hundred bounded WARNING lines, not thousands, so the per-asset signal stays useful without per-request aggregation complexity.

routers/utils/asset_conversion.py
```diff
@@ -26,6 +27,37 @@
...
+    value for pre-upload dedup and for local↔remote asset linking ("merged"
+    state) in the mobile client. Gumnut exposes that value as
+    ``AssetResponse.checksum_sha1``.
+
+    Gumnut's other ``checksum`` field is a base64-encoded SHA-256 (44 chars).
+    It must never be sent on this field: a wrong-format value can never equal
+    the client-computed SHA-1, so it silently breaks dedup and makes a
+    backed-up photo appear as two separate timeline entries.
+
+    When ``checksum_sha1`` is null (rare legacy rows), return ``""`` rather
```
ternarybits, line 47: Are there any rows that are actually empty?  Did we not backfill them?

ternarybits (approved):

routers/utils/asset_conversion.py
taggartgorman: Yes, a bit less than 2% of rows in production are missing SHAs. There is a separate backend task for the backfill script.

PR #188, opened 2026-05-27
Serve 1440px preview for wide-landscape thumbnails

taggartgorman:
## What
- Upgrade a `thumbnail`-size request to the 1440px `preview` variant for **wide-landscape** assets — `width > height` AND aspect ratio > 1.5 (`_LANDSCAPE_PREVIEW_ASPECT_THRESHOLD`) — in `routers/api/assets.py`. The decision happens in a new `_upgrade_variant_for_aspect()` helper called inside `_retrieve_and_stream_variant()`, before variant-key resolution.
- Missing/zero dimensions and `preview`/`fullsize`/`original` requests pass through unchanged; video upgrades resolve to `preview_image` via the existing `_image`-suffix logic.
- Document the behavior, the landscape-only rationale, and the backend preview/thumbnail co-presence assumption in `docs/references/code-practices.md`.

## Why
- Landscape assets (most visibly wide videos) look blurry in the Immich web timeline. The grid is a justified-rows layout that renders every row at a fixed target height (~235px). A thumbnail is generated at 250px on its longest edge, so for a landscape asset that 250px is the *width* and the height is only `250/aspect` (~140px at 16:9). The grid upscales it to fill the row → blurry.
- Portrait assets are unaffected (250px lands on their height, which already meets the row height), so the fix is gated on landscape orientation — this targets the real root cause (aspect/orientation, not media type) while avoiding the ~10–20× bandwidth cost of the larger variant on assets that are already crisp.
- The asset dimensions are already fetched in `_retrieve_and_stream_variant`, so there is no extra backend round-trip.

## Test plan
- [x] `uv run ruff check` / `uv run ruff format --check` — clean
- [x] `uv run pytest` — 974 passed
- New tests in `tests/unit/api/test_assets.py`: wide-landscape image → `preview`; wide-landscape video → `preview_image`; portrait / square / near-square / 1.5-boundary / unknown / `0,0` dims → still `thumbnail`; `preview`-size request not re-upgraded.


taggartgorman (commented):
**Reviewed — no blocking issues.** Clean, well-scoped, and thoroughly tested. (Posting as a comment since GitHub blocks self-approval.)

Upgrades only `thumbnail` requests for wide-landscape assets (`width > height` and aspect > 1.5) to the 1440px `preview`, reusing dims already fetched in `_retrieve_and_stream_variant` (no extra round-trip). Verified:

- `AssetResponse.width`/`height` are `Optional[int]` — matches the helper signature; pyright passes.
- Only `view_asset` (`/thumbnail`) reaches the upgrade with a `thumbnail` variant; `download_asset` and `play_asset_video` pass `"original"` and short-circuit. No auth/deletion/schema surface touched.
- None/0/portrait/square handled before the division; the 1.5 boundary is exclusive and tested; video resolves to `preview_image` downstream.
- New `width/height=None` defaults on the test mock keep existing tests on the thumbnail path; CI green (974 passing).

The backend preview/thumbnail co-presence assumption (upgrade happens before the `asset_urls` existence check, so a missing `preview` would 404 rather than degrade) is documented in both the docstring and `code-practices.md`, with mitigation guidance if the contract ever changes. Reasonable, deliberate trade-off.

**Findings: 0 Must Fix / 0 Should Fix / 0 Consider.** Considered "do nothing" (leaves blur) and "upgrade all thumbnails" (~10–20× bandwidth on already-crisp assets); the aspect+orientation gate beats both.

ternarybits (approved):

routers/api/assets.py
```diff
@@ -104,6 +104,47 @@
...
+# request is upgraded to the larger `preview` variant. The Immich web timeline
+# is a justified-rows grid where every row renders at a fixed target height
+# (~235px). A thumbnail is generated at 250px on its longest edge, so for a
+# landscape asset that 250px is the *width* and the height is only 250/aspect
+# (~140px at 16:9). The grid then upscales it to fill the row, which looks
+# blurry. Serving the 1440px `preview` keeps wide-landscape cells crisp.
+# 1.5 catches 16:9 and wider while leaving 4:3/3:2 on the cheap thumbnail.
+# Portrait assets are unaffected: 250px lands on their height, which already
+# meets the row height, so they stay sharp without the bandwidth cost. Tunable.
+_LANDSCAPE_PREVIEW_ASPECT_THRESHOLD = 1.5
```
ternarybits, line 117: 3:2 is a pretty common aspect ratio.  Make sure there's a test that for images exactly 3:2, the smaller thumbnail is served.

routers/api/assets.py
```diff
@@ -104,6 +104,47 @@
     {"thumbnail", "preview", "fullsize"}
 )
 
+# Aspect ratio (width / height) above which a landscape asset's `thumbnail`
+# request is upgraded to the larger `preview` variant. The Immich web timeline
+# is a justified-rows grid where every row renders at a fixed target height
+# (~235px). A thumbnail is generated at 250px on its longest edge, so for a
```
ternarybits, line 110: Honestly, this looks like a poor decision for Immich.  If the fixed target height is ~235px, then even 4:3 and 3:2 photos will look blurry, right?

16:9 will create a thumbnail that's 250x140
4:3 will create a thumbnail that's 250x187
3:2 will create a thumbnail that's 250x166

What am I missing here?

PR #189, opened 2026-05-27
Surface video duration through the Immich adapter

taggartgorman:
## What
- Forward the upstream nullable float-seconds `duration` (asset response, gumnut-sdk 0.101.0) through all four sites that emit an Immich duration: the single-asset `AssetResponseDto`, the upload-ready WebSocket payload, the sync v1 stream converter, and the timeline-bucket array.
- Add `format_duration(seconds) -> "HH:MM:SS.ffffff" | None` (decomposed from rounded integer microseconds so a value just under a minute/hour boundary carries up rather than rendering an out-of-range `:60`).
- Bump `gumnut-sdk` to `>=0.101.0`.

## Why
- Immich clients render `0:00` for every video because the adapter hardcodes a zero/empty duration at each emit site. Now that the asset response carries a real `duration`, the adapter forwards it so video tiles and the single-asset view show the correct length.
- When upstream `duration` is NULL (an image, or a video whose duration hasn't been extracted yet) each site preserves exactly what it emitted before — video zero / image empty-string for the non-nullable single-asset DTO, `None` everywhere else — rather than fabricating a value. This mirrors how the adapter already handles unknown width/height: forward what upstream provides, never guess. Videos stay `0:00` until extraction and backfill populate real values.

## Notes
- Fixed asset test mocks that set a nonexistent `duration_in_seconds`: a bare `unittest.mock.Mock` auto-vivifies a truthy child for any unset attribute, so once converters read `.duration` the NULL path would break. Shared fixtures and inline mocks now set `duration` explicitly.

## Test plan
- [x] `uv run pytest` (980 passing, incl. new formatter/converter/timeline-bucket tests covering populated durations, NULL fallback per site, and minute/hour rounding boundaries)
- [x] `uv run ruff format` / `uv run ruff check`
- [x] `uv run pyright` (0 errors)

### DIFF
--- a/routers/api/timeline.py
+++ b/routers/api/timeline.py
@@ -217,8 +217,9 @@ async def get_time_bucket(
 
         is_trashed_list.append(bool(asset.trashed_at))
 
-        # None preserves this bucket's prior all-None duration behavior;
-        # format_duration owns the NULL handling and string formatting.
+        # Forward each asset's duration: format_duration returns the
+        # HH:MM:SS.ffffff string, or None on NULL upstream — preserving this
+        # bucket's prior all-None output for images / not-yet-extracted videos.
         duration_list.append(format_duration(asset.duration))
 
     # Return as dict to bypass Pydantic validation issues with None in List[str]
--- a/routers/api/timeline.py
+++ b/routers/api/timeline.py
@@ -217,8 +217,8 @@ async def get_time_bucket(
 
         is_trashed_list.append(bool(asset.trashed_at))
 
-        # Upstream NULL (image, or video not yet extracted) stays None, matching
-        # the prior all-None behavior; a real float renders HH:MM:SS.ffffff.
+        # None preserves this bucket's prior all-None duration behavior;
+        # format_duration owns the NULL handling and string formatting.
         duration_list.append(format_duration(asset.duration))
 
     # Return as dict to bypass Pydantic validation issues with None in List[str]
--- a/routers/api/timeline.py
+++ b/routers/api/timeline.py
@@ -14,6 +14,7 @@ from routers.immich_models import (
     TimeBucketsResponseDto,
 )
 from routers.utils.asset_conversion import (
+    format_duration,
     mime_type_to_asset_type,
     resolve_capture_datetime,
 )
@@ -184,6 +185,7 @@ async def get_time_bucket(
     visibility_list = []
     local_offset_hours_list = []
     is_trashed_list = []
+    duration_list: list[str | None] = []
 
     for asset in filtered_assets:
         asset_id = asset.id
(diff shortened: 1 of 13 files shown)


routers/api/timeline.py
```diff
@@ -215,12 +217,16 @@ async def get_time_bucket(
 
         is_trashed_list.append(bool(asset.trashed_at))
 
+        # None preserves this bucket's prior all-None duration behavior;
+        # format_duration owns the NULL handling and string formatting.
+        duration_list.append(format_duration(asset.duration))
```
taggartgorman, line 223: **[Consider]** The comment leads with "None preserves this bucket's prior all-None duration behavior," but this line now appends `format_duration(asset.duration)`, which returns a real `HH:MM:SS.ffffff` string for populated videos and only `None` on NULL upstream. Leading with "None preserves..." can read as if the bucket still emits all-None. Reframing to lead with the new behavior (and note the NULL fallback) reads more clearly:

```suggestion
        # Forward each asset's duration: format_duration returns the
        # HH:MM:SS string, or None on NULL upstream — preserving this
        # bucket's prior all-None output for images / not-yet-extracted videos.
        duration_list.append(format_duration(asset.duration))
```

taggartgorman (commented):
**Review verdict: approve-equivalent** (GitHub blocks formally approving your own PR, so this is posted as a COMMENT). No blocking issues — fixes the long-standing `0:00` video duration by forwarding the new upstream `duration` through all four emit sites, following the established width/height "forward, don't fabricate" pattern. One non-blocking comment below.

## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (1 Consider — comment wording, inline below)
- [x] **Codebase value**: Fixes a real user-visible bug (every video rendered `0:00`) by forwarding an upstream value with a small pure helper. Versus "do nothing" it removes the wrong hardcoded value; a smaller scope (one site) would leave the other three still wrong, so all four emit sites is the right scope. Mirrors the existing width/height handling rather than inventing a new convention.
- [x] **Security/auth**: No auth/authz/permission logic touched. Files: asset_conversion.py, sync/converters.py, timeline.py, pyproject/uv.lock, tests — pure data formatting.
- [x] **Data deletion**: No destructive operations. Searched the diff for delete/trash/remove — none present.
- [x] **Confidence**: Correctness well understood. `format_duration` rounds to whole microseconds before decomposing so `:60` is unreachable; type/nullability verified against AssetResponse (`float | None`), AssetResponseDto (`str`), SyncAssetV1 (`str | None`); new tests cover populated/NULL per site and minute/hour boundaries; CI green (lint, types, 980 tests).
- [x] **Purpose clarity**: PR description and code make the purpose and the NULL-preservation rationale explicit.

routers/api/timeline.py
taggartgorman: Fixed in e77cde2 — reframed the comment to lead with the forward behavior (`format_duration` returns the `HH:MM:SS.ffffff` string, or `None` on NULL upstream), then note the all-None fallback for images / not-yet-extracted videos.

ternarybits (approved):

PR #195, opened 2026-06-01
Pass photos-api thumbhash through to Immich clients

taggartgorman:
## What
- Emit each asset's upstream `thumbhash` directly at all four conversion sites that previously hardcoded a placeholder: the REST `AssetResponseDto`, the upload-ready WebSocket `SyncAssetV1`, the sync-stream `SyncAssetV1`, and the timeline bucket builder. It is a plain nullable passthrough (no field remap, no warning — unlike `resolve_immich_checksum`), so it is emitted inline rather than through a resolver helper; the rationale for emitting `None` lives as a comment at the REST site.
- Bump the `gumnut-sdk` pin `0.101.0 → 0.102.0` — the first SDK release whose `AssetResponse` carries the new nullable `thumbhash` field, so the adapter can finally see the upstream value. (The relock also materializes the existing 14-day supply-chain cooldown into an absolute `exclude-newer` timestamp.)
- Add a `TestThumbhashEmission` class (paralleling `TestChecksumEmission`), a per-asset timeline thumbhash test, and `thumbhash` defaults to the asset mocks/fixtures across the suite that flow through the converters.

## Why
- The upstream Photos API now generates a real per-asset base64 ThumbHash, but the adapter emitted hardcoded placeholders at every conversion site, so Immich mobile/web clients got no usable blur placeholder while thumbnails load.
- The timeline-bucket endpoint was worse than a no-op: it shipped the **same** constant thumbhash for every tile in a bucket, so clients rendered one identical blur across an entire month of photos. Wiring real per-asset values fixes that.
- The Immich DTO fields (`AssetResponseDto.thumbhash`, `SyncAssetV1.thumbhash`) already exist and are correctly typed `str | None`, so this is a pure passthrough with no Immich-side schema work.

## Test plan
- [x] `uv run pytest` — 999 passing (new `TestThumbhashEmission` + per-asset timeline test; updated asset mocks/fixtures suite-wide)


taggartgorman (commented):
## ✅ Approved — automated review (`/gumnut:review-pr`)

Clean, well-scoped passthrough. Wires each asset's real upstream `thumbhash` through the four active conversion sites (REST `AssetResponseDto`, upload-ready + sync-stream `SyncAssetV1`, timeline bucket), replacing hardcoded placeholders.

**Findings: 0 Must Fix / 0 Should Fix / 0 Consider.**

Verified:
- `gumnut-sdk` 0.102.0 `AssetResponse.thumbhash` is `str | None`; both target DTO fields and the timeline `list[str | None]` field accept it — type-correct passthrough, no Pydantic risk.
- Timeline `thumbhash_list` is appended in the same per-asset loop as the other lists, so ordering matches `asset_count`. The change fixes a real bug: the bucket previously shipped one identical blur for every tile.
- "Four conversion sites" is accurate — the only other `thumbhash` holder (`ImmichAsset`) is dead/commented code.
- Lockfile `exclude-newer-package: gumnut-sdk = false` makes the 0.102.0 pin consistent with the materialized cooldown timestamp.
- Strong test coverage: `TestThumbhashEmission` (all converters + null path) and a per-asset timeline test asserting distinct values + `None` (would catch a regression to the shared constant). Suite-wide `thumbhash = None` fixture defaults correctly avoid the Mock-truthiness trap.

### Pre-approve checklist
- [x] **Codebase value**: Replaces a per-bucket shared-constant blur bug with real per-asset values; "do nothing" and "resolver helper" alternatives are both worse
- [x] **Security/auth**: Untouched — opaque base64 passthrough, no auth/token code in diff
- [x] **Data deletion**: Untouched — read-path conversion only
- [x] **Confidence**: `str | None` → `str | None`; field types verified in SDK + all DTOs; ordering checked; covered by new + existing tests; CI green
- [x] **Purpose clarity**: Clear from description and code

ternarybits (approved):

PR #198, opened 2026-06-02
GUM-958: expose real storage quota to Immich + remap over-quota 507→400

ternarybits:
## What

Surface the Gumnut API's new per-user storage caps to Immich clients — the quota numbers (on both quota surfaces) and the over-quota upload error.

**Quota fields** — replace the hardcoded Immich `quotaSizeInBytes` / `quotaUsageInBytes` with the real per-user numbers from the Gumnut `UserResponse` (`storage_limit_bytes` cap, `storage_used_bytes` usage), via a shared `map_user_quota` helper, on **both** places the adapter reports quota:
- `GET`/`PUT /api/users/me` (`UserAdminResponseDto`).
- The sync stream's `SyncAuthUserV1` payload (was hardcoded `None`/`0`; `quotaUsageInBytes` is a required int so a missing upstream value coerces to 0).
- Bump `gumnut-sdk` `0.104.0 → 0.105.0` (first release exposing the user storage fields).

**Over-quota upload (507 → 400)** — the Gumnut API returns `507 Insufficient Storage` when a cap is hit on upload. Immich's own server returns **400** ("Quota has been exceeded!"), and the Immich **mobile** app string-matches that exact message to abort the rest of a batch upload. So translate 507 → 400 with Immich's canonical message on every upload path:
- Buffered path (`map_gumnut_error`) + the global error handler: remap 507 → 400 (previously a raw 507 passed through).
- Streaming path: remap 507 → 400 before the generic `5xx → 502` mapping that otherwise masked it.
- Log the 507 at WARNING (an expected, user-attributable condition), not ERROR.

## Why

The adapter previously reported a fixed 100 GiB cap / 1 GiB usage on `/users/me`, `None`/`0` on the sync stream, and surfaced an over-quota upload as a generic gateway error — none of which a real Immich server does. This makes the adapter behave like a real Immich server for storage limits, with the two quota surfaces consistent.

Rollout-safe: both quota fields are `int | None`, and the SDK yields `None` for an omitted field (Immich treats it as unlimited/unknown). The detail string is kept byte-for-byte because the Immich mobile client depends on it for batch-abort. The Gumnut-side 507 enforcement isn't deployed yet, so the remap is forward-compatible.

## Stacking

Stacked on the GUM-983 asset-`include` branch, which owns the `0.102 → 0.104` SDK bump and the asset-field null-guards. Review against that branch; retarget to `main` once GUM-983 merges. Tracking: GUM-958.

## Test plan

- [x] `uv run pytest` — 1010 pass (quota mapping/rollout on both `/me` endpoints, the sync auth-user payload, and the helper; 507 → 400 on buffered, streaming, and global-handler paths; 507 → WARNING log level; e2e + sync mocks updated)
- [x] `uv run ruff format` / `uv run ruff check` / `uv run pyright` clean


routers/utils/current_user.py
```diff
@@ -22,6 +23,22 @@
 from routers.utils.gumnut_id_conversion import safe_uuid_from_user_id
 
 
+def map_user_quota(user: UserResponse) -> tuple[int | None, int | None]:
```
ternarybits: Let's give the return type a named type so that it's more clear what we're returning, and what "none" means.

ternarybits: Fixed in 359896c — the return is now an `ImmichUserQuota` NamedTuple (`size_bytes` / `usage_bytes`) with a docstring documenting the field-to-Immich mapping and what `None` means (unknown/unlimited during rollout). All three call sites now read by attribute.

routers/utils/current_user.py
```diff
@@ -65,8 +85,8 @@ async def get_current_user_admin(
         shouldChangePassword=False,
         status=UserStatus.active if user.is_active else UserStatus.deleted,
         storageLabel="admin",
-        quotaSizeInBytes=1024 * 1024 * 1024 * 100,
-        quotaUsageInBytes=1024 * 1024 * 1024,
+        quotaSizeInBytes=quota_size,
+        quotaUsageInBytes=quota_usage,
```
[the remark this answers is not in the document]
ternarybits: Already handled later in this PR: `gumnut_user_to_sync_auth_user_v1` now maps the real quota via `map_user_quota` as of 8f97edc, so sync-bootstrap clients get the same quota as `/api/users/me` (usage coerces a rollout `None` to 0, since the field is a required int here). This review ran against the first commit (64c6fc3), before that fix landed. Resolving.

routers/utils/error_mapping.py
```diff
@@ -192,6 +220,11 @@ def map_gumnut_error(
             extra=log_extra,
             exc_info=exc_info,
         )
+        # Surface an over-quota upload as Immich's native 400, not the raw 507.
+        if e.status_code == status.HTTP_507_INSUFFICIENT_STORAGE:
```
taggartgorman, line 224: **[Consider]** 507→400 remap isn't tested through the buffered upload endpoint

The remap is covered at the helper level (`test_error_mapping.py`), the global handler (`test_exceptions.py`), and the streaming pipeline (`test_streaming_upload.py`) — but this buffered path, `upload_asset()`'s `except` clause calling `map_gumnut_error` (`routers/api/assets.py:606-607`), isn't exercised end-to-end. That's the most likely real-world trigger for an over-quota upload, and the repo already has the convention for endpoint-level upload error tests: `test_upload_asset_api_error` (`tests/unit/api/test_assets.py:482`) drives `upload_asset` with a mocked SDK 401 and asserts the mapped status. A future refactor of `upload_asset`'s error handling could bypass the remap with no failing test.

Non-blocking — mirror `test_upload_asset_api_error`: mock the SDK create to raise a 507, call `upload_asset(...)`, and assert `status_code == 400` and detail `"Quota has been exceeded!"`.

taggartgorman (approved):
Reviewed via `/gumnut:review-pr` (multi-perspective: correctness/data-integrity/security/silent-failures, performance/test-coverage, clarity/minimalism, + orchestrator value check). Verified the core claims independently rather than trusting the description: the SDK's non-validating response construction materializes an omitted `storage_*` field as `None` at runtime (so the rollout-safe `int | None` handling is correct even though the SDK declares the fields as required `int`); the Immich model field types match the coercion logic (`SyncAuthUserV1.quotaUsageInBytes` is a required `int`, so `None`→0; the other quota fields are `int | None`); and the 507→400 remap is correctly ordered on all three upload paths (buffered helper, streaming pipeline before the 5xx→502 fallback, and the global handler).

Clean change that makes the adapter behave like a real Immich server for storage limits: fabricated quota constants are replaced with real per-user values via a single `map_user_quota`/`ImmichUserQuota` helper (which also removes the two-`int|None` transposition risk), and the over-quota error reaches clients as the exact string the Immich mobile app string-matches to abort a batch upload. Well-tested (mapping + rollout-`None` on both `/me` endpoints and the sync auth user; 507→400 on buffered-helper, streaming, and global-handler paths; 507→WARNING log level).

## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (1 Consider, inline below)
- [x] **Codebase value**: Better than before. Replaces fabricated quota (fixed 100 GiB / 1 GiB) and a generic gateway error with real per-user numbers and Immich's native over-quota response. vs. *do nothing* — the adapter keeps reporting false quota and masking over-quota uploads as 502; vs. *smaller scope* — the quota fields and the 507 remap are two halves of the same "expose real storage caps" goal. `map_user_quota` genuinely simplifies 3 call sites and centralizes the `None` semantics; the NamedTuple is proportionate (two same-typed `int | None` values that were transposable).
- [x] **Security/auth**: Does not touch authentication, authorization, or permission logic. Changes are quota *reporting* (informational) and HTTP error *translation* (507→400); enforcement stays upstream in photos-api. Files checked: `config/exceptions.py`, `routers/utils/error_mapping.py`, `services/streaming_upload.py`, `routers/utils/current_user.py`, `routers/api/users.py`, `routers/api/sync/converters.py`. Searched: auth/token/permission/401.
- [x] **Data deletion**: No destructive operations. Searched: delete/trash/remove/drop — none in scope.
- [x] **Confidence**: High. Runtime `None`-on-omitted-field, Immich model field types, and remap ordering/blast-radius all independently verified. CI green (1010 tests; ruff + pyright clean).
- [x] **Purpose clarity**: Clear from the description and the code.

_One non-blocking nuance, not a finding: during the rollout window `/me` reports `quotaUsageInBytes: null` while the sync stream reports `0` for the same user — forced by the Immich schema (the sync field is a required `int`) and self-resolving once photos-api ships the fields._

PR #210, opened 2026-06-10
Isolate refreshed JWT per request to prevent cross-user leak

ternarybits:
## What
- Replace process-global refreshed-token storage (a module-level `ContextVar` plus `threading.local`) with a per-request mutable holder installed on a `ContextVar` before `call_next`; the Gumnut response hook mutates that shared holder.
- Install the holder at the top of `AuthMiddleware.dispatch`; drop the now-redundant global clear-on-entry/exit calls.
- Add regression tests for per-context isolation, interleaved requests where only one refreshes, concurrent response hooks, and the lazy-install / clear-without-holder fallbacks — plus an end-to-end test driving two overlapping requests with distinct session tokens through the real `BaseHTTPMiddleware` (`httpx.AsyncClient` + `ASGITransport`), asserting `update_stored_jwt` is called exactly once with the refreshing session's token. Event-coordinated concurrency tests are bounded with `asyncio.timeout(5)` (no pytest-timeout is configured).
- Document the per-request-state hazard in `docs/references/code-practices.md`, and document in the middleware that tokens refreshed during StreamingResponse body generation are intentionally dropped (`call_next` returns at response start; the safe direction — self-heals via sliding refresh).
- Remove the vacuous `test_token_refresh_cookie_has_security_flags` integration test: its mock patched a namespace the middleware doesn't read, it targeted an unauthenticated endpoint, and a JWT refresh never touches cookies by design.

## Why
The refreshed access token (`x-new-access-token`) was captured into shared process-global state. Under Starlette's `BaseHTTPMiddleware`, a `ContextVar` set inside the downstream handler does not propagate back to `dispatch`, so the only live path was the shared `threading.local` — which every concurrent request on the event-loop thread shares. Under concurrent multi-user load, one user's refreshed JWT could be persisted into another user's session, granting cross-account access in both directions. Reproduced against the repo's own Starlette version. Found during an internal auth scan (HIGH severity).

## Test plan
- [x] `uv run pytest` (full suite, 1039 pass)
- [x] `uv run ruff format` / `ruff check` / `pyright` clean


docs/references/code-practices.md
```diff
@@ -67,6 +67,8 @@ This format is enforced by the global exception handler in `config/exceptions.py
 
 **Note:** In middleware (e.g., `auth_middleware.py`), you must return `JSONResponse` directly with this format, as `HTTPException` raised in `BaseHTTPMiddleware.dispatch()` is not caught by FastAPI's exception handlers due to Starlette's middleware architecture.
 
+**Passing per-request state from a handler back up to the middleware:** a `ContextVar.set()` inside the downstream handler does **not** propagate back to `dispatch()` after `call_next` (same Starlette `BaseHTTPMiddleware` boundary). Never use process-global / module-level mutable state (a bare `ContextVar`, `threading.local`, etc.) to carry per-request values — under concurrent load one request can read another's value, which for credentials means cross-user contamination. Install a per-request mutable holder on a `ContextVar` in `dispatch()` *before* `call_next`, and have the handler mutate that object (see `gumnut_client.py` refreshed-token holder).
```
taggartgorman, line 70: **[Should Fix]** The frontmatter still reads `last-updated: 2026-06-03` — this edit should bump it to `2026-06-10`, per this doc's own convention (the checklists here say "bump `last-updated`" on meaningful doc edits). Anchored here because line 3 isn't in the diff.

routers/middleware/auth_middleware.py
```diff
@@ -173,7 +178,4 @@ async def dispatch(self, request: Request, call_next):
             if self.REFRESH_HEADER in response.headers:
```
taggartgorman, line 186: **[Consider]** Worth a comment at the `get_refreshed_token()` read above (line 162): `call_next` returns at `http.response.start`, so tokens refreshed *while a StreamingResponse body is generated* (e.g. sync stream backend calls) mutate the holder after this read and are silently dropped. That's strictly safer than the old behavior (which could persist them into the wrong session) and self-heals via sliding refresh on the next non-streaming request — but the intentional drop deserves documenting so it isn't "rediscovered" as a bug later.

tests/unit/utils/test_token_refresh.py
```diff
@@ -220,6 +224,120 @@ async def mock_get_session_store():
         assert "x-new-access-token" not in response.headers
 
 
+class TestPerRequestIsolation:
```
taggartgorman, line 232: **[Consider]** These unit tests guard the storage mechanism well, but the original bug passed the whole suite because nothing drove *concurrent* requests through the real middleware. Consider one end-to-end regression test: `httpx.AsyncClient` + `ASGITransport`, two overlapping requests with distinct session tokens where only one endpoint triggers `_response_hook`, asserting `update_stored_jwt` is called exactly once and only with the refreshing session's token. (Mitigating: if holder propagation ever broke, the failure direction is a dropped refresh, not a leak — hence Consider.)

taggartgorman (approved):
Reviewed with focus on the concurrency/auth semantics, since this is a credential-handling fix.

**The fix is correct.** I verified the load-bearing mechanism against the installed Starlette 1.0.1: `call_next` spawns the downstream app task from within `dispatch()`'s context, so the holder installed before `call_next` propagates in, and the hook's *mutation* of that object (unlike a `ContextVar.set`) is visible back in `dispatch()`. With the `threading.local` fallback removed, the cross-user contamination path is gone, and the lazy-holder fallback is context-local so it can't reintroduce leakage. The pre-existing `TestTokenRefresh` middleware tests now only pass if the holder mechanism works across the real `BaseHTTPMiddleware` boundary, so CI doubles as boundary proof. Side paths check out: the streaming-upload `set_refreshed_token` caller runs in the request context, and websockets bypass the middleware but never call the backend.

Three non-blocking comments inline (1 Should Fix on doc frontmatter, 2 Consider).

docs/references/code-practices.md
ternarybits: Fixed — bumped `last-updated` to 2026-06-10.

routers/middleware/auth_middleware.py
ternarybits: Done — added a comment at the read documenting the intentional drop (call_next returns at response start; safe direction; self-heals via sliding refresh).

tests/unit/utils/test_token_refresh.py
ternarybits: Added — test_overlapping_requests_through_real_middleware drives two overlapping requests with distinct session tokens through the real BaseHTTPMiddleware (httpx.AsyncClient + ASGITransport), only one triggering the response hook, and asserts update_stored_jwt is called exactly once with the refreshing session's token. Bounded with asyncio.timeout(5) so a coordination regression fails rather than hanging CI.

PR #222, opened 2026-06-16
Implement POST /api/faces via Gumnut faces.create

taggartgorman:
## What
- Replace the silent no-op `POST /api/faces` stub with a real handler that draws a user-specified face box on an asset and links it to a person, via `client.faces.create(asset_id, bounding_box={x,y,w,h}, person_id)`.
- Rescale the box from Immich's **preview** pixel space (the downscaled image the user drew on, reported as `imageWidth`/`imageHeight`) into the asset's **full-resolution** frame, which is the space the Gumnut API stores and `GET /faces` reads boxes in — so the box round-trips identically instead of landing shrunk in the top-left corner.
- Bump `gumnut-sdk` to `>=0.116.0`, which exposes the auto-generated `faces.create()`.
- Add `TestCreateFace` unit tests; update the architecture and gap-analysis docs to mark the faces module fully implemented, and document the "bumping the Gumnut SDK" testing gotcha.

## Why
- Immich v2.7's "create a face on-the-fly" flow in the face tag editor calls `POST /people` (which already works) then `POST /faces` (previously a silent no-op). The dropped second call left orphaned person records with no linked face or thumbnail. This wires the second call through to the Gumnut API so the person gets a face and thumbnail.

## Box scaling
- The box is scaled by its **endpoints** (`x2 = min(round((x+w)·scale_x), asset.width)`, similarly for y) with a far-edge clamp, **not** by rounding width/height independently. The Gumnut API rejects (422) a box whose `x+w` exceeds `asset.width` (or `y+h` exceeds `asset.height`); independent per-component rounding can push a box drawn flush to the asset's edge 1px past the bound. Endpoint scaling lands an edge-flush box on the asset bound exactly.
- When the asset has no stored dimensions, the box passes through unscaled and the response falls back to the request's preview dimensions.

## Source type
- `sourceType` is derived from the Gumnut `FaceResponse.source` (`manual` → `manual`, else `machine-learning`) via a shared `_to_immich_source_type()` helper used by **both** `create_face` and `get_faces`, so a manually created face reports `manual` consistently on create and on re-read (`get_faces` previously hardcoded `machine-learning`).
- The person is fetched to populate the response, degrading to `person=null` on a fetch failure (mirroring `GET /faces`).

## Notes
- The 0.116 SDK adds a now-required `source` field to the auto-generated `FaceResponse` model. Sync-stream tests that construct `FaceResponse` directly as fixtures were updated to supply it (caught by the full test suite, not the faces tests alone).

## Test plan
- [ ] `uv run pytest` (full suite — 1043 passing)
- [ ] `uv run pyright` (0 errors)
- [ ] `uv run ruff check` / `uv run ruff format`

### DIFF
--- a/docs/references/code-practices.md
+++ b/docs/references/code-practices.md
@@ -1,6 +1,6 @@
 ---
 title: "Code Practices"
-last-updated: 2026-06-15
+last-updated: 2026-06-16
 ---
 
 # Code Practices
@@ -132,6 +132,10 @@ The SDK is auto-generated (Stainless), so a version bump can add **newly-require
    If the new endpoint also emits a WebSocket event (existing one or new), also update **both** websocket docs and bump their `last-updated`:
    - `docs/architecture/websocket-implementation.md` — move the event out of the "Not Applicable" table into the Phase 1 supported table (or add a new row), with the payload, Web/Mobile columns, and a Notes value that names the emit site.
    - `docs/references/websocket-events-reference.md` — update the Summary Table row and the per-event section. Upstream-Immich and adapter triggers may diverge (e.g., upstream emits `on_asset_update` only from sidecar processing; the adapter emits it from `PUT /api/assets/{id}`); make both paths explicit so future readers don't assume the upstream-only note still applies.
+
+### Derive multi-path fields through one helper
+
+A field the adapter surfaces from more than one path must be derived identically in each: the write handler, the paired read handler, **and** the sync-stream converter (`routers/api/sync/converters.py`). Route any value mapped from a Gumnut field through a shared helper instead of hardcoding it per site, or the same entity reads back differently by path — e.g. a face's `sourceType` (mapped from `FaceResponse.source`) returned by `create_face` but hardcoded in `get_faces` / the sync converter, or a face box stored in a different coordinate frame than `get_faces` reads back. When you touch one emit site for such a field, grep for every constructor of the Immich response type (`AssetFaceResponseDto`, `SyncAssetFaceV1` / `SyncAssetFaceV2`, …) before assuming it has one home.
 
 ### Reading Gumnut asset fields — request them via `include`
 
--- a/docs/references/code-practices.md
+++ b/docs/references/code-practices.md
@@ -104,6 +104,10 @@ The two are not auto-synced, but CI enforces that they match (see the `check-imm
 
 Forgetting step 2 causes silent drift — the served web UI stays on the old Immich version while the API models advance.
 
+### Bumping the Gumnut SDK
+
+The SDK is auto-generated (Stainless), so a version bump can add **newly-required** fields to response models (e.g. `FaceResponse.source` arrived in 0.116). Tests construct these models directly as fixtures, so a bump can break suites unrelated to the endpoint you're touching. Run the **full** `uv run pytest` after a bump (not just the changed endpoint's tests), and when a required field is added, `grep` the tests for `<Model>(` to fix every direct construction.
+
 ### Implementing New Endpoints
 
 1. **Generate models**: Use `generate_immich_models.py` to create up-to-date Pydantic models (see [development tools](development-tools.md))
--- a/routers/api/faces.py
+++ b/routers/api/faces.py
@@ -117,9 +117,55 @@ async def get_faces(
 @router.post("", status_code=201)
 async def create_face(
     request: AssetFaceCreateDto,
-):
+    client: AsyncGumnut = Depends(get_authenticated_gumnut_client),
+) -> AssetFaceResponseDto:
+    """Draw a user-specified face box on an asset and assign it to a person.
+
+    Backs Immich's "create a face on-the-fly" flow in the face tag editor: the
+    client creates the person first (POST /people), then calls this to draw the
+    box and link it to that person. Both `assetId` and `personId` are required
+    by the Immich request DTO.
     """
-    Create a new face.
-    This is a stub implementation that returns a empty response.
-    """
-    return
+    gumnut_asset_id = uuid_to_gumnut_asset_id(request.assetId)
+    gumnut_person_id = uuid_to_gumnut_person_id(request.personId)
+
+    face = await client.faces.create(
+        asset_id=gumnut_asset_id,
+        bounding_box={
+            "x": request.x,
+            "y": request.y,
+            "w": request.width,
+            "h": request.height,
+        },
+        person_id=gumnut_person_id,
+    )
+
+    # The create response carries only person_id, so fetch the full person to
+    # populate the Immich response. Degrade to person=None on failure rather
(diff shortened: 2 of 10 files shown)


routers/api/faces.py
```diff
@@ -117,9 +117,75 @@ async def get_faces(
...
+    )
+    scale_y = (
+        asset.height / request.imageHeight
+        if asset.height and request.imageHeight
+        else 1.0
+    )
+
+    face = await client.faces.create(
+        asset_id=gumnut_asset_id,
+        bounding_box={
```
taggartgorman: **[Should Fix]** Scaling each box component with an independent `round()` can push the box 1px past the asset bounds when it's drawn flush to the asset's right/bottom edge. Concretely: `asset.width=4000`, `request.imageWidth=1920`, `x=18`, `width=1902` (flush) → `round(18·2.0833)=38`, `round(1902·2.0833)=3963`, so `x+w = 4001 > 4000`. (Interior boxes never overflow — only edge-flush ones.)

The SDK contract says the box *"must not exceed"* the asset's `width`/`height`, which implies the backend validates. **Can you confirm whether `faces.create` rejects or clamps an out-of-bounds box?**
- If it **rejects** → a user dragging a face box to the image edge gets a failed create in this exact flow. Clamp the far edge before storing, e.g. derive `x2 = min(round((x+width)·scale_x), asset.width)` and `w = x2 - round(x·scale_x)` (same for y/h), and add a regression test for an edge-flush box.
- If it **clamps** → this is cosmetic (≤1px) and can drop to Consider.

routers/api/faces.py
```diff
@@ -117,9 +117,75 @@ async def get_faces(
...
+        person = convert_gumnut_person_to_immich(gumnut_person)
+    except Exception:
+        logger.warning(
+            "Failed to fetch person for created face",
+            extra={"person_id": gumnut_person_id, "asset_id": gumnut_asset_id},
+        )
+
+    bb = face.bounding_box or {}
+    source_type = (
+        SourceType.manual if face.source == "manual" else SourceType.machine_learning
```
taggartgorman: **[Consider]** This correctly derives `sourceType` from `face.source`, but `get_faces` (line 110, unchanged) still hardcodes `SourceType.machine_learning` for every face. So a face created here as `manual` reports `sourceType=manual` on create, then flips to `machine-learning` when re-read via `GET /faces`. Now that `FaceResponse.source` is available, consider deriving it the same way in `get_faces` — ideally via a shared one-line helper so the mapping lives in one place.

routers/api/faces.py
```diff
@@ -117,9 +117,75 @@ async def get_faces(
...
+    # Immich's face editor reports the box in the pixel space of the *preview*
+    # it rendered (request.imageWidth/imageHeight — the downscaled image the
+    # user drew on), but Gumnut stores and returns face boxes in the asset's
+    # full-resolution pixel space (the frame get_faces pairs boxes with via
+    # asset.width/height). Without rescaling, a box drawn at the center of a
+    # 1440-wide preview is stored verbatim and later read back against the
+    # 3024-wide asset, landing shrunk in the top-left corner. Scale the box up
+    # to the asset's real dimensions before storing.
+    asset = await client.assets.retrieve(gumnut_asset_id)
+    scale_x = (
```
taggartgorman, line 153: **[Consider]** The `else 1.0` scale fallback (and the `imageWidth=asset.width or request.imageWidth` response fallback) handle the case where the asset has no stored dimensions, but no test exercises it — `_make_asset` always sets integer width/height. Consider one test with `width=None, height=None` asserting the box passes through unscaled and the response falls back to the request's preview dimensions.

docs/references/code-practices.md
```diff
@@ -104,6 +104,10 @@ The two are not auto-synced, but CI enforces that they match (see the `check-imm
 
 Forgetting step 2 causes silent drift — the served web UI stays on the old Immich version while the API models advance.
 
+### Bumping the Gumnut SDK
```
taggartgorman, line 107: **[Consider]** This adds a substantive "Bumping the Gumnut SDK" subsection, but the file's `last-updated` frontmatter (line 3) is still `2026-06-15`. The other two docs in this PR were bumped to `2026-06-16`; bump this one too to match the repo's docs-as-system-of-record convention.

taggartgorman (commented):
Reviewed the new `POST /api/faces` handler. Solid change — it replaces a silent no-op stub (which left orphaned person records in Immich's "create a face on-the-fly" flow) with a correct implementation, and rescaling the box into the asset's full-resolution space is the right call so it round-trips through `GET /faces`. Tests and docs are thorough; all CI green.

One verification request and three non-blocking suggestions below — none block the core approach.

**Findings: 0 Must Fix · 1 Should Fix (verification request) · 3 Consider**

_Considered but skipped: de-duplicating the box→DTO field mapping shared by `get_faces`/`create_face` — the surrounding fields diverge intentionally, so extraction is low-value._

routers/api/faces.py
taggartgorman: Fixed in da96700. Confirmed the Gumnut API **rejects** (does not clamp) an out-of-bounds box — `faces.create` raises a 422 when `x + w > asset.width` — and the overflow is reachable at realistic dimensions (preview 1440×1920 → asset 2469×1134, edge-flush `x=240,w=1200` stored `x+w=2470 > 2469`). Switched to endpoint scaling with a far-edge clamp (`x2 = min(round((x+w)·scale_x), asset.width)`) so an edge-flush box stores `x+w == asset.width` exactly, and added `test_edge_flush_box_does_not_exceed_asset_bounds` pinning old→2470 / new→2469.

routers/api/faces.py
taggartgorman: Fixed in da96700. Added a shared `_to_immich_source_type()` helper; `get_faces` now derives `sourceType` from `face.source` like `create_face`, so a manual face no longer flips to machine-learning on re-read. Added a `get_faces` assertion for both directions. Follow-up noted in the PR summary comment: the sync-stream converter (`gumnut_face_to_sync_face_v1`) still hardcodes machine-learning — flagged for a separate PR since it's a different (customer-facing) subsystem.

routers/api/faces.py
taggartgorman: Fixed in da96700. Added `test_box_passes_through_when_asset_has_no_dimensions` (`width/height=None`): the box is stored unscaled and the response falls back to the request's preview dimensions. `_make_asset` now types `width/height` as `int | None` to model the SDK's `Optional[int]`.

docs/references/code-practices.md
taggartgorman: Fixed in da96700 — bumped `last-updated` to `2026-06-16` to match the other two docs in this PR.

ternarybits (approved):

PR #230, opened 2026-06-24
Fix 500 on GET /api/albums for albums with naive start/end dates

taggartgorman:
## What
- Route an album's `startDate`/`endDate` through the existing keep-local-time datetime helper (`to_immich_local_datetime`) in `convert_gumnut_album_to_immich`, instead of passing the backend values in raw.
- Add a regression test covering a non-null naive date (now tz-aware), a `None` date (stays `None`), and an already-tz-aware non-UTC date (wall-clock preserved, re-labeled UTC).
- Document the gotcha in the datetime-handling code practices.

## Why
`AlbumResponseDto.startDate`/`endDate` are typed `AwareDatetime`, but the Gumnut API serializes an album's start/end date (the min/max of its assets' local capture datetimes) timezone-naive when the capture timezone is unknown. A non-null naive value raised an unhandled pydantic `ValidationError`, returning HTTP 500 for the entire `GET /api/albums` response — so a single offending album broke the whole albums list in the web client. Each asset's `localDateTime` already flows through this same helper, so albums now display a date range consistent with their assets, and `None` still passes through unchanged.

## Test plan
- [x] `uv run pytest` — full suite passes, including the new regression test
- [x] Verified the regression test fails without the fix (the documented `ValidationError: Input should have timezone info`)


taggartgorman (commented):
## Review: Fix 500 on GET /api/albums for albums with naive start/end dates

**Assessment: auto-approvable** — 0 Must Fix, 0 Should Fix, 1 Consider (non-blocking). Posted as a `COMMENT` review because GitHub doesn't allow approving one's own PR.

The fix correctly mirrors the established `asset_conversion.py` precedent: capture-derived dates (`startDate`/`endDate`, the min/max of assets' `local_datetime`) route through the keep-local-time helper, while server record timestamps (`createdAt`/`updatedAt`) stay raw because they're always tz-aware. `to_immich_local_datetime` preserves the wall-clock and re-labels UTC, keeping the album range consistent with the dates shown on its assets, and `None` passes through unchanged. The regression test pins all three cases at the endpoint level (the layer that actually 500'd).

### Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (1 Consider inline, non-blocking)
- [x] **Codebase value**: Fixes a real 500 that broke the entire albums list for a single offending album; follows the existing converter precedent; adds a regression test and documents the gotcha. Routing through the existing helper is the minimal fix — no smaller-scoped or simpler alternative.
- [x] **Security/auth**: No auth/authz/permission logic touched (converter + test + doc only).
- [x] **Data deletion**: No destructive operations touched.
- [x] **Confidence**: Change is limited to one converter field-pair; behavior verified against the helper and the asset-conversion precedent; covered by the new test; CI lint/typecheck/test all pass.
- [x] **Purpose clarity**: Purpose and approach are clear from the description and code.

docs/references/code-practices.md
```diff
@@ -35,6 +35,7 @@ Style, patterns, and conventions for the immich-adapter codebase.
...
   - `resolve_file_modified_at` — for the `fileModifiedAt` field. Unlike capture time, the Gumnut API does not resolve a single modify-time field — `asset.file_modified_at` is the raw file mtime — so the helper applies a `metadata.modified_datetime → asset.file_modified_at` cascade here. Do not "align" this with the capture-time helpers; the asymmetry is deliberate.
+- **Immich DTO datetime fields are `AwareDatetime`; Gumnut datetimes can be naive.** The Gumnut API serializes a local capture datetime timezone-naive when the capture timezone is unknown, and a non-null naive value fails an `AwareDatetime` field's validator (unhandled `ValidationError` → 500). Never assign a Gumnut datetime straight into an Immich DTO — route local-time fields through `to_immich_local_datetime` (labels naive values UTC keep-local-time, passes `None` through). This applies beyond asset fields: `AlbumResponseDto.startDate`/`endDate` are the min/max of assets' `local_datetime`, so they use the same helper as each asset's `localDateTime`.
```
taggartgorman: **[Consider]** The lead clause “Never assign a Gumnut datetime straight into an Immich DTO” reads as an unqualified absolute, but the same converters assign `created_at`/`updated_at` straight into `AlbumResponseDto`/`AssetResponseDto` — and those Immich fields are `AwareDatetime` too. The load-bearing rule (the em-dash clause) is already correctly scoped to *local-capture-time* fields, which is the only family the Gumnut API serializes naive; record timestamps are tz-aware and pass through raw on purpose. Consider scoping the lead so a future reader doesn't take it literally and flag the existing raw `createdAt`/`updatedAt` assignments as bugs.

```suggestion
- **Immich DTO datetime fields are `AwareDatetime`; Gumnut datetimes can be naive.** The Gumnut API serializes a local capture datetime timezone-naive when the capture timezone is unknown, and a non-null naive value fails an `AwareDatetime` field's validator (unhandled `ValidationError` → 500). Don't assume a Gumnut datetime is tz-aware: route local-capture-time fields through `to_immich_local_datetime` (labels naive values UTC keep-local-time, passes `None` through). Server row timestamps (`created_at`/`updated_at`) are tz-aware and pass through raw. This applies beyond asset fields: `AlbumResponseDto.startDate`/`endDate` are the min/max of assets' `local_datetime`, so they use the same helper as each asset's `localDateTime`.
```

ternarybits (approved):

PR #238, opened 2026-07-02
Fix validateToken to detect expired tokens so Immich clients re-login

taggartgorman:
## What
- `POST /api/auth/validateToken` now probes the backend via `users.me()` (through `Depends(get_authenticated_gumnut_client)`) instead of only checking that a JWT string is present in the session.
- An expired stored token now surfaces as a 401, so the Immich mobile auth guard bounces the user to the login screen on the next launch / guarded navigation; a still-refreshable token is renewed transparently by the auth middleware.
- Rewrote the validateToken unit + integration tests (valid → `authStatus` true, missing → 401, expired → 401 "JWT has expired", transient backend error → 502 not 401) and updated the auth-gating-stubs guidance in `docs/references/code-practices.md`.

## Why
After the app is idle long enough for the stored JWT to expire past its refresh window, the session token is still valid, so every real API call 401s with "JWT has expired" — but `validateToken` (the only endpoint the Immich client trusts to gate login) returned `authStatus=true` because it only checked JWT *presence*. The client stayed silently "logged in" with a dead token and sync failed until a manual logout. Probing the backend makes `validateToken` reflect real token validity. A transient / non-auth backend error maps to 502 (not 401), so users are only bounced on a genuine auth failure, never on a blip.

## Why probe the backend instead of checking `exp` locally?
Instead of calling `users.me()`, we could decode the stored token and check its `exp` in the adapter. We're deliberately using the probe:

- **The adapter treats the token as opaque.** It stores the token encrypted, forwards it, and never interprets it — it doesn't even hold the key to verify it. A local `exp` check would add JWT-payload parsing and make an auth decision from an *unverified* claim, crossing that boundary. The probe instead delegates the validity decision to the Gumnut API, which owns the token's lifecycle.
- **It's the existing pattern.** `users.me()` is the same lightweight pre-flight the sync-stream endpoint and the current-user dependency already use, so this adds no new machinery.
- **It stays correct if the backend's token policy evolves.** A local decode would usually give the same answer today, but it bakes in the assumption that expiry is the only way a token goes invalid. That's the backend's policy to define and change; a local check would silently drift from it, re-introducing the "client trusts a dead token" bug this fixes.

The tradeoff is one extra call per guarded navigation — a cheap, connection-pooled call on a navigation-gated path (not a hot path). If it ever shows up in metrics, the follow-up is a short "last-validated" cache window that keeps the delegation correct while cutting the calls (see Follow-ups).

## Test plan
- [x] `uv run pytest` — 1046 passed
- [x] `uv run ruff check` and `uv run pyright` — clean
- [ ] Manual: with an expired stored JWT for a session, cold-launch the Immich mobile app → lands on the login screen instead of a silently-broken home screen

## Follow-ups (deliberately out of scope)
- A short "last-validated" cache window to avoid a `users.me()` round-trip on every guarded navigation.
- Reconciling the OAuth-callback session TTL / `expires_at` omission so dead sessions are eventually evicted.


taggartgorman (commented):
Reviewed the change end-to-end — the core fix is correct and well-scoped. `validateToken` no longer trusts mere JWT *presence*; it probes the backend via `users.me()`, so an expired token surfaces as a 401 and the Immich auth guard bounces the user to login instead of sitting on a dead session.

Verified the full error→status mapping with concrete inputs: missing JWT → 401 (in the dependency, before any probe), expired → 401, valid → 200, transient `APIConnectionError` → 502, retry-exhausted `RateLimitError` → 502. The `AuthenticationError` vs `RateLimitError` branch ordering in the global handler is not shadowed, so a transient blip or throttle can't masquerade as expiry and eject a valid session. The `users.me()` probe + bubble-to-global-handler approach matches the existing `get_current_user_admin` pattern, so no new machinery is introduced. Docs and public-repo hygiene are clean, and all CI checks pass.

No blocking issues (0 Must Fix / 0 Should Fix). Two optional, non-blocking test-quality suggestions inline — neither blocks merge.

tests/unit/api/test_auth.py
```diff
@@ -219,57 +219,60 @@ async def test_logout_websocket_error_does_not_fail_logout(
...
-        request = Mock()
-        request.state = Mock(spec=["jwt_token"])
-        request.state.jwt_token = None
-
-        with pytest.raises(HTTPException) as exc_info:
-            await validate_access_token(request)
+    async def test_propagates_auth_error_when_jwt_expired(self):
+        """An expired JWT makes users.me() raise AuthenticationError, which must
+        propagate so the global GumnutError handler maps it to a 401."""
+        import httpx
```
taggartgorman: **[Consider]** Reuse the existing conftest helper instead of building `AuthenticationError` inline.

Both expired-JWT tests (this one and `test_expired_jwt_returns_401` at ~356) construct the error with ~12 near-identical lines plus function-local `import httpx` / `from gumnut import AuthenticationError`, while the sibling transient test one class down already reuses a conftest helper (`make_sdk_connection_error`). Since the global handler only reads `exc.body['detail']` and `exc.status_code`, each block collapses to:

```suggestion
        from gumnut import AuthenticationError

        from tests.conftest import make_sdk_status_error
```

...then replace the manual `httpx.Request`/`httpx.Response`/`AuthenticationError(...)` plumbing with `make_sdk_status_error(401, "JWT has expired", body={"detail": "JWT has expired"}, cls=AuthenticationError)`. Drops both local `import httpx` statements and centralizes on the helper the file already uses.

tests/unit/api/test_auth.py
```diff
@@ -314,19 +317,85 @@ async def mock_get_session_store():
...
+        """A transient (non-auth) backend failure maps to 502, not 401.
+
+        validateToken gates when the auth guard logs the user out, so only a
+        genuine auth failure may return 401. A backend blip (connection error)
+        must NOT masquerade as an expired token and eject a still-valid session.
+        """
+        from tests.conftest import make_sdk_connection_error
+
+        mock_client = Mock()
+        mock_client.users.me = AsyncMock(side_effect=make_sdk_connection_error())
```
taggartgorman, line 392: **[Consider]** (optional) `validateToken` is the endpoint that gates Immich logout, so it may be worth locking that a *rate-limited* backend also stays 502 (not 401), mirroring this transient-error test:

```suggestion
        mock_client.users.me = AsyncMock(side_effect=make_sdk_status_error(429, cls=RateLimitError))
```

(with `make_sdk_status_error` / `RateLimitError` imports, asserting `502`). Marginal — the `RateLimitError→502` mapping is already covered globally in `test_exceptions.py` and this transient test already establishes the "non-auth error → not 401" contract, so fine to skip.

tests/unit/api/test_auth.py
taggartgorman: Fixed in e045c71 — both expired-JWT tests now use the conftest `make_sdk_status_error(..., cls=AuthenticationError)` helper and drop the local `import httpx`. Also added a Testing-section bullet in `code-practices.md` documenting the conftest SDK-error helpers.

tests/unit/api/test_auth.py
taggartgorman: Skipping — `RateLimitError`→502 is already covered globally in `test_exceptions.py`, and the transient-error test one class down already establishes the non-auth-error → not-401 contract for this endpoint. Agreed it is marginal, so leaving it out to avoid duplicate coverage.

ternarybits (approved):

PR #240, opened 2026-07-02
Filter map markers to geotagged assets via the Gumnut coordinate filter

ternarybits:
## What
`/map/markers` returns all of a caller's GPS-tagged assets. The adapter previously paged the Gumnut API's asset list and discarded non-geotagged assets **client-side**, bounded by a scan cap so a low-GPS-density library wouldn't walk hundreds of pages chasing markers.

This passes a **world-covering `bbox`** (`-180,-90,180,90`) to the Gumnut API's coordinate filter, so the backend returns **only geotagged assets** (served index-only from its geo covering index).

## Why it's better
- Paging cost now scales with the **marker count**, not the library's GPS density — a sparsely-geotagged library no longer walks page after page of non-geotagged assets.
- The client-side scan cap is removed (every returned asset is a marker); the 2000-marker response cap stays.
- Fewer, cheaper round-trips for the common map-view load.

## Dependencies / deploy ordering
- Bumps `gumnut-sdk` to `>= 0.120.0` (adds the `bbox` filter param).
- Requires the Gumnut API's coordinate-filter support to be **live** before this deploys. If it deploys first, the unknown `bbox` param is ignored and the endpoint falls back to listing all assets (correct results, just without the geotagged pre-filter), so deploy after the API is updated.

## Test plan
- [x] `bbox` forwarded to the SDK; only `limit`/`include`/`bbox` (+ date range) forwarded
- [x] Marker cap still enforced with an explicit break across pages
- [x] Defensive skip for assets missing metadata/coords retained
- [x] Obsolete scan-cap test removed
- [x] format / lint / pyright / full suite (1045) green

### DIFF
--- a/routers/api/map.py
+++ b/routers/api/map.py
@@ -40,6 +40,14 @@ GEOTAGGED_WORLD_BBOX = "-180,-90,180,90"
 # are dropped.
 MAP_MARKERS_CAP = 2000
 
+# Safety net bounding total assets walked. In the normal path the coordinate
+# filter returns only geotagged assets, so the marker cap fires first and this
+# never triggers. It guards the degraded case where the `bbox` filter is *not*
+# applied — an older Gumnut API that ignores the unknown param (e.g. the adapter
+# deploying ahead of the API), or a filter regression — which would otherwise
+# turn a low-GPS-density library's marker request into a full-library scan.
+MAX_ASSETS_SCANNED = 30 * GUMNUT_API_MAX_PAGE_SIZE
+
 
 @router.get("/markers")
 async def get_map_markers(
@@ -83,8 +91,10 @@ async def get_map_markers(
         list_kwargs["local_datetime_before"] = fileCreatedBefore.isoformat()
 
     markers: list[MapMarkerResponseDto] = []
+    assets_scanned = 0
     marker_cap_hit = False
     async for asset in client.assets.list(**list_kwargs):
+        assets_scanned += 1
         metadata = asset.metadata
         # The bbox filter guarantees a coordinate, but guard defensively so an
         # unexpected null can't crash marker construction (lat/lon are required).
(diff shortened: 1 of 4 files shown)


routers/api/map.py
```diff
@@ -73,18 +73,21 @@ async def get_map_markers(
...
         list_kwargs["local_datetime_before"] = fileCreatedBefore.isoformat()
 
     markers: list[MapMarkerResponseDto] = []
-    assets_scanned = 0
     marker_cap_hit = False
     async for asset in client.assets.list(**list_kwargs):
-        assets_scanned += 1
         metadata = asset.metadata
+        # The bbox filter guarantees a coordinate, but guard defensively so an
+        # unexpected null can't crash marker construction (lat/lon are required).
```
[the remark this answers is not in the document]
ternarybits: Good catch — restored MAX_ASSETS_SCANNED as a safety net (60be23e → latest). In the normal path the coordinate filter returns only geotagged assets so the marker cap fires first and the scan bound never triggers; it now guards exactly the degraded case you describe (bbox not honored → non-geotagged assets returned → bound prevents a full-library scan). Added a test covering the unfiltered-response fallback.

routers/api/map.py
```diff
@@ -73,6 +81,9 @@ async def get_map_markers(
     list_kwargs: dict[str, Any] = {
         "limit": GUMNUT_API_MAX_PAGE_SIZE,
         "include": ASSET_INCLUDE_METADATA_ONLY,
+        # Filter to geotagged assets server-side (see GEOTAGGED_WORLD_BBOX) so we
+        # page through markers, not the whole library.
+        "bbox": GEOTAGGED_WORLD_BBOX,
```
ternarybits, line 91: I don't think this filters anything at all.  The world box is the entire world.  It doesn't change the response change to just markers.

ternarybits: Skipping a code change here — the world bbox does filter, because of how the Gumnut API implements the coordinate filter. The bbox compiles to a SQL range predicate over the nullable latitude/longitude columns, and a NULL coordinate never satisfies a range comparison, so any bbox — including the whole-world box — excludes every asset without GPS data. It acts as a *has-coordinates* filter, not a spatial narrowing. The API test suite covers exactly this case: an asset with `latitude=None`/`longitude=None` is excluded from a bbox-filtered `list_assets` response.

So the endpoint now pages through only geotagged assets server-side (every returned asset becomes a marker) instead of scanning the whole library and discarding non-GPS assets adapter-side. The response shape was already markers-only in both versions — the change is in what gets paged, not what gets returned. If a deployed API ever ignored the `bbox` param, the `MAX_ASSETS_SCANNED` safety net bounds that degraded case.

ternarybits: I verified this against the live library (9,628 assets) — the world box does filter:

- Newest 100 **without** bbox → 93 with coordinates + **7 without** (the 5 newest have none).
- Newest 100 **with** `bbox=-180,-90,180,90` → **100 with coordinates, 0 without**; those 7 coordinate-less assets are gone.

Why a world-wide box still filters: the Gumnut API's bbox filter is a coordinate range check (`latitude BETWEEN min AND max`, etc.). An asset with no coordinate has `latitude = NULL`, and `NULL BETWEEN …` is NULL — not true — in SQL, so it's excluded even by a box that spans the whole planet. So the response is reduced to exactly the marker-eligible (geotagged) assets, which is the goal.

That said, I agree `-180,-90,180,90` reads oddly for "only geotagged." If you'd prefer an explicit contract, I can add a dedicated "has coordinates" filter to the API and switch the adapter to it as a small follow-up. This version is correct as-is — happy to go either way.

PR #241, opened 2026-07-02
Apply dateTimeRelative as minutes to match Immich

taggartgorman:
## What
- Fix `PUT /api/assets` bulk update: `AssetBulkUpdateDto.dateTimeRelative` is now applied as a **minutes** offset (`timedelta(minutes=...)`) instead of seconds, matching the Immich server.
- Rename `_RelativeShift.relative_seconds` → `relative_minutes` and update the surrounding docstrings, the inline comment, six unit-test assertions in `tests/unit/api/test_assets.py`, and the architecture-doc phrase.

## Why
- The adapter interpreted `dateTimeRelative` as seconds, but Immich's server applies it as minutes (its `updateDateTimeOriginal` adds `${delta} minute`::interval, in both the current target v2.7.5 and v3.0). Any client sending a relative capture-time shift through the adapter got a result **60× off** — a shift the user intends as N minutes was applied as N seconds.
- The bug is behavioral, not in the DTO. The generated `routers/immich_models.py` field description ("seconds") is left untouched on purpose: it mirrors the pinned upstream v2.7.5 OpenAPI spec (whose own doc-string is wrong — the server SQL has always used minutes) and would revert on the next model regeneration. The `_RelativeShift` docstring documents this mismatch so a reader isn't misled.

## Test plan
- [x] `uv run pytest tests/unit/api/test_assets.py` — 152 passed. The six relative-shift assertions now pin `timedelta(minutes=N)` where N equals the request's `dateTimeRelative=N`; since `timedelta(minutes=N) != timedelta(seconds=N)` for N≠0, they are genuine regression guards against the old behavior.
- [x] `uv run ruff format`, `uv run ruff check`, `uv run pyright` — all clean.


taggartgorman (commented):
## Review: clean ✅ — approve-worthy (posted as COMMENT; GitHub disallows self-approval)

Reviewed the full diff across correctness, data integrity, security, silent failures, performance, test coverage, clarity, minimalism, and value. **0 Must Fix · 0 Should Fix · 0 Consider.**

The core claim was verified against Immich's source: the server applies `dateTimeRelative` as `` "dateTimeOriginal" + `${delta} minute`::interval `` (introduced in immich-app/immich#17717), so the old `timedelta(seconds=…)` was 60× off. Immich itself later corrected the field description from "seconds" to "minutes", corroborating that the upstream DTO text is a known doc-string error — which is why leaving the generated `immich_models.py` description untouched and documenting the mismatch in the `_RelativeShift` docstring is the right call.

The six repinned tests are genuine regression guards: requests send `dateTimeRelative=3600.0`/`60.0` and assert `timedelta(minutes=N)`, and since `timedelta(minutes=N) != timedelta(seconds=N)` for N≠0 they fail under the old behavior.

### Pre-approve checklist
- ✅ **Findings:** 0 Must Fix / 0 Should Fix / 0 Consider
- ✅ **Codebase value:** removes a 60× unit bug; full-sweep (code + tests + doc) beats a code-only fix that would leave stale guards and a wrong architecture doc
- ✅ **Security/auth:** not touched
- ✅ **Data deletion:** not touched (metadata-update path only)
- ✅ **Confidence:** verified against Immich's `${delta} minute`::interval SQL; covered by six regression tests
- ✅ **Purpose clarity:** clear from description and code

ternarybits (approved):

PR #249, opened 2026-07-06
Emit asset/timeline duration as integer milliseconds (Immich v3)

taggartgorman:
## What
- Add a `duration_ms(seconds) -> int | None` converter and emit integer **milliseconds** (null when unknown) for `AssetResponseDto.duration` and `TimeBucketAssetResponseDto.duration`, matching Immich v3's move from the `HH:MM:SS.ffffff` interval string to nullable int ms.
- Remove the `_resolve_single_asset_duration` placeholder helper — the nullable v3 field lets an unknown duration be `null` instead of the old `00:00:00.000000` (video) / `""` (image) placeholders.
- Leave `format_duration` (the interval string) unchanged for the two `SyncAssetV1` sync-payload sites, whose `duration` stays `str | None` in the v3 spec. The int-ms sync variant (`SyncAssetV2`) is a downstream step that will reuse `duration_ms`.
- Correct the v3 change-analysis design doc: drop a reference to a nonexistent `routers/models.py` and describe the as-built converter split.

## Why
- Immich v3 (GA) changed `duration` on every asset/timeline/upload response to nullable integer milliseconds. Without this, clients receive the wrong wire type for those responses.
- **Merges to `migration/immichv3`, not `main`.** This is an early step of the Immich v3 retarget; the long-lived integration branch merges to `main` once all steps land. The branch is intentionally not standalone-green — pyright and pytest carry pre-existing breakage from sibling v3 changes not yet landed (deleted `PersonWithFacesResponseDto`, removed `AssetResponseDto` device fields, `SyncAssetV1.createdAt` now required). This change was verified to add **zero** new pyright errors vs. the branch base.

## Test plan
- [ ] `TestDurationMs` covers the converter: populated values, `None` passthrough, round down/up, and the `0` floor.
- [ ] `TestDurationEmission`: the REST `AssetResponseDto` asserts integer ms (`12.5s -> 12500`), while the `SyncAssetV1` websocket/sync assertions remain the interval string (unchanged in v3).
- [ ] Timeline bucket test asserts `[65250, None]` for `65.25s` / `None`.
- [ ] Note: these unit tests cannot execute on this integration branch yet — pytest collection is blocked by a sibling v3 change that deletes `PersonWithFacesResponseDto`. Verified here by inspection, standalone arithmetic check, and a baseline pyright diff; they run green once the sibling converter issues land.

### DIFF
--- a/docs/design-docs/immich-v3-api-changes.md
+++ b/docs/design-docs/immich-v3-api-changes.md
@@ -2,7 +2,7 @@
 title: "Immich v2.7.5 → v3.0 API Change Analysis"
 status: active
 created: 2026-06-16
-last-updated: 2026-07-02
+last-updated: 2026-07-06
 ---
 
 # Immich v2.7.5 → v3.0 API Change Analysis
--- a/docs/design-docs/immich-v3-api-changes.md
+++ b/docs/design-docs/immich-v3-api-changes.md
@@ -104,7 +104,7 @@ mistake it for behavioral change.
 
 | Change | Detail | Adapter impact |
 |---|---|---|
-| **`duration` string → integer ms** | `AssetResponseDto`, `TimeBucketAssetResponseDto`, `AssetMediaCreateDto`. Was `"HH:MM:SS.ffffff"` interval string, now integer **milliseconds**, nullable on `AssetResponseDto`. | `routers/utils/asset_conversion.py:format_duration` emits the interval string today — must emit int ms. Touches `routers/models.py:69`, and the `duration` fields in `routers/immich_models.py`. Affects every asset/timeline/upload response. |
+| **`duration` string → integer ms** | `AssetResponseDto`, `TimeBucketAssetResponseDto`, `AssetMediaCreateDto`. Was `"HH:MM:SS.ffffff"` interval string, now integer **milliseconds**, nullable on `AssetResponseDto`. | A `duration_ms` helper in `routers/utils/asset_conversion.py` emits int ms (null when unknown) for `AssetResponseDto` / `TimeBucketAssetResponseDto`; the interval-string `format_duration` is retained for `SyncAssetV1`, whose `duration` stays a string in v3 (`SyncAssetV2` is the int-ms sync variant — see §5). The `duration` fields live in `routers/immich_models.py`. Affects every asset/timeline/upload response. |
 | **`AlbumResponseDto` restructured** | Removed `assets`, `owner`, `ownerId`. Owner now derived from **`albumUsers[0]`** (now `minItems:1`; documented ordering: owner first, then auth user if different, rest alphabetical). `shared` now required. | Album responses no longer inline assets or owner. Album conversion must populate `albumUsers[0]` as owner and stop emitting `owner`/`ownerId`/`assets`. |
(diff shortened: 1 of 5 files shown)


docs/design-docs/immich-v3-api-changes.md
```diff
@@ -104,7 +104,7 @@ mistake it for behavioral change.
 
 | Change | Detail | Adapter impact |
 |---|---|---|
-| **`duration` string → integer ms** | `AssetResponseDto`, `TimeBucketAssetResponseDto`, `AssetMediaCreateDto`. Was `"HH:MM:SS.ffffff"` interval string, now integer **milliseconds**, nullable on `AssetResponseDto`. | `routers/utils/asset_conversion.py:format_duration` emits the interval string today — must emit int ms. Touches `routers/models.py:69`, and the `duration` fields in `routers/immich_models.py`. Affects every asset/timeline/upload response. |
+| **`duration` string → integer ms** | `AssetResponseDto`, `TimeBucketAssetResponseDto`, `AssetMediaCreateDto`. Was `"HH:MM:SS.ffffff"` interval string, now integer **milliseconds**, nullable on `AssetResponseDto`. | A `duration_ms` helper in `routers/utils/asset_conversion.py` emits int ms (null when unknown) for `AssetResponseDto` / `TimeBucketAssetResponseDto`; the interval-string `format_duration` is retained for `SyncAssetV1`, whose `duration` stays a string in v3 (`SyncAssetV2` is the int-ms sync variant — see §5). The `duration` fields live in `routers/immich_models.py`. Affects every asset/timeline/upload response. |
```
taggartgorman, line 107: **[Should Fix]** Bump the doc's `last-updated` frontmatter to `2026-07-06`.

This duration row (and the §7 plan entry on line 251) are meaningfully edited by this PR, but the `last-updated:` field in the frontmatter (line 5) still reads `2026-07-02`. Both the code-review convention (stale `last-updated:` on a meaningfully-edited doc is itself a finding) and this repo's own `docs/references/code-practices.md` (repeated "bump `last-updated`") call for advancing it to the edit date.

taggartgorman (commented):
Reviewed the Immich v3 `duration` → integer-milliseconds change. The conversion is correct and minimally scoped:

- `duration_ms` is wired at exactly the two int-ms response sites (`AssetResponseDto`, timeline bucket), while `format_duration` is correctly retained for the two `SyncAssetV1` sites (still `str | None` in v3). Verified against the regenerated model types (`SyncAssetV1.duration: str|None`, `SyncAssetV2.duration: int|None`).
- Dropping the `_resolve_single_asset_duration` placeholder is a net simplification and more correct — a not-yet-extracted video now emits `null` (the nullable v3 field) instead of a fabricated `00:00:00.000000`, while a genuine zero-length video still emits `0`.
- The added `TestDurationMs` cases and updated emission/timeline assertions are arithmetically correct (e.g. `65.25s → 65250`). These can't execute on `migration/immichv3` yet due to the unlanded sibling `PersonWithFacesResponseDto` deletion blocking collection — out of scope for this PR.

One minor, non-blocking convention nit inline (doc frontmatter). Otherwise LGTM.

docs/design-docs/immich-v3-api-changes.md
taggartgorman: Done — bumped `last-updated` to `2026-07-06` in f4476d7.

PR #250, opened 2026-07-06
Restructure AlbumResponseDto to Immich v3 shape (albumUsers[0])

taggartgorman:
## What
- Populate `albumUsers[0]` as the album owner and stop emitting the v3-removed `owner`/`ownerId`/inline `assets`; keep `shared`/`hasSharedLink`. Drop the now-unused `assets` parameter from the album converter (every caller already passes `asset_count`).
- `GET /albums/{id}` no longer fetches + converts every album asset (they are no longer inlined in v3) — a net perf win. The `withoutAssets` query param is retained for client compatibility but is now a no-op.
- Add `tests/unit/utils/test_album_conversion.py` for the v3 shape; fix the `get_album_info` tests that expected an asset fetch.

## Why
Immich v3's `AlbumResponseDto` removed `assets`/`owner`/`ownerId`, made `albumUsers` required (`min_length=1`, owner first), and made `shared` required. The prior conversion still passed the removed fields and `albumUsers=[]`, which violates the new `min_length=1` at runtime. Gumnut has no album sharing, so `albumUsers` is always exactly the single owner entry.

## Base branch
Targets `migration/immichv3`, **not `main`** — step 2 of the Immich v3 retarget. The integration branch merges to `main` once all steps land.

## Test plan
- [ ] New/updated unit tests assert the v3 shape: owner in `albumUsers[0]` with the owner role, `shared` present, no `owner`/`ownerId`/`assets`, `assetCount` forwarded (incl. `None → 0`), and `GET /albums/{id}` performs no asset fetch.
- [ ] `migration/immichv3` is intentionally red — the suite can't run yet (pre-existing removed-`Error1` import + pydantic `pattern`-on-`UUID` instantiation blockers, now documented in the v3 API analysis doc). Verified for this change: ruff clean; scoped pyright adds zero new errors.

### DIFF
--- a/tests/unit/utils/test_album_conversion.py
+++ b/tests/unit/utils/test_album_conversion.py
@@ -1,0 +1,63 @@
+"""Tests for album conversion utilities — focused on the Immich v3 shape.
+
+``convert_gumnut_album_to_immich`` builds ``AlbumResponseDto``, which in Immich
+v3 no longer carries ``owner``/``ownerId`` or inline ``assets``. The owner is
+instead derived from ``albumUsers[0]`` (``minItems: 1``), and ``shared`` is
+required. Gumnut has no album sharing, so ``albumUsers`` is always exactly the
+single owner entry.
+"""
+
+import pytest
+
+from routers.immich_models import AlbumUserRole
+from routers.utils.album_conversion import convert_gumnut_album_to_immich
+
+
+class TestConvertGumnutAlbumToImmich:
+    def test_owner_carried_in_album_users(self, sample_gumnut_album, mock_current_user):
+        """The current user is the sole albumUsers[0] entry, with the owner role."""
+        result = convert_gumnut_album_to_immich(
+            sample_gumnut_album, mock_current_user, asset_count=5
+        )
+
+        assert len(result.albumUsers) == 1
+        assert result.albumUsers[0].role == AlbumUserRole.owner
+        assert result.albumUsers[0].user.id == mock_current_user.id
+        assert result.albumUsers[0].user == mock_current_user
+
+    def test_removed_v3_fields_absent(self, sample_gumnut_album, mock_current_user):
+        """owner/ownerId/assets were dropped from the v3 DTO — none are emitted."""
+        result = convert_gumnut_album_to_immich(
+            sample_gumnut_album, mock_current_user, asset_count=5
+        )
+
+        assert not hasattr(result, "owner")
+        assert not hasattr(result, "ownerId")
+        assert not hasattr(result, "assets")
+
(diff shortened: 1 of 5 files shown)


tests/unit/utils/test_album_conversion.py
```diff
@@ -0,0 +1,63 @@
...
+    def test_owner_carried_in_album_users(self, sample_gumnut_album, mock_current_user):
+        """The current user is the sole albumUsers[0] entry, with the owner role."""
+        result = convert_gumnut_album_to_immich(
+            sample_gumnut_album, mock_current_user, asset_count=5
+        )
+
+        assert len(result.albumUsers) == 1
+        assert result.albumUsers[0].role == AlbumUserRole.owner
+        assert result.albumUsers[0].user.id == mock_current_user.id
+        assert result.albumUsers[0].user == mock_current_user
```
taggartgorman: **[Consider]** Redundant assertion pair — the full-object equality on line 26 subsumes the `.id`-only check on line 25 (equal objects necessarily share `.id`), so the two lines together assert only what line 26 alone does. The full-equality check is the stronger one; you can drop the id-only line.

```suggestion
        assert result.albumUsers[0].user == mock_current_user
```

taggartgorman (commented):
**Review — COMMENT (non-blocking)**

Clean v3 retarget of `AlbumResponseDto`. The owner correctly moves to the single `albumUsers[0]` entry (satisfying the new `min_length=1`), the v3-removed `owner`/`ownerId`/inline `assets` fields are dropped, required `shared`/`hasSharedLink` are kept, and `GET /albums/{id}` no longer fetches + converts every album asset — a real perf win with **no count regression** (the count already came from `gumnut_album.asset_count`, and the fetched assets fed only the now-removed inline array). Scope is tight: every caller was already passing `asset_count` and none passed the dropped `assets` param, so the signature change breaks nothing. Removed imports (`ASSET_INCLUDE`, `convert_gumnut_asset_to_immich`, `AssetResponseDto`) are dead everywhere, and `logger` is still used elsewhere in `albums.py`.

**0 Must Fix / 0 Should Fix / 2 Consider.**

- **[Consider]** The new/updated tests are verified by **inspection only** — they can't run on `migration/immichv3` yet due to the two documented, separately-tracked blockers (the removed `Error1` import breaking router-layer collection, and `pattern`-on-`UUID` raising `TypeError` when any `AlbumResponseDto` is constructed — its `id` is such a field). With no CI on the PR, a runtime-only slip (fixture typo, an assertion that never fires) would go undetected until the branch is unblocked. Not a defect of this PR — just worth confirming these go green once the blockers clear.
- **[Consider]** One trivial redundant assertion, noted inline.

Nothing here blocks the migration step.

PR #251, opened 2026-07-06
Retarget AssetResponseDto to Immich v3 shape

taggartgorman:
## What
- **Asset responses (v3 shape):** the asset converter no longer emits `deviceAssetId`/`deviceId`/`unassignedFaces` and builds `people` as `PersonResponseDto` (no inline face boxes). Removed the now-dead `convert_gumnut_person_to_immich_with_faces` and the deleted `PersonWithFacesResponseDto` import.
- **Search:** `search_random` drops the removed `RandomSearchDto.deviceId` filter.
- **Upload:** the upload path no longer requires/reads device fields (v3 clients don't send them). The Gumnut API's `assets.create` still requires `device_asset_id`/`device_id`, so the adapter synthesizes them — a shared `GUMNUT_UPLOAD_DEVICE_ID` placeholder (`routers/api/constants.py`) plus a **unique per-upload `device_asset_id` (UUID)**.
- **Docs:** design-doc notes — the upload device-field synthesis, and broadening the known pydantic/3.14 pattern-constraint blocker from UUID-only to UUID/datetime fields.

## Why
Immich v3 removed these fields from `AssetResponseDto` and the search/upload DTOs (models already regenerated on this branch), so the adapter code that set/read them mismatched the generated models. The upload synthesis uses a **unique** `device_asset_id` (not the filename) so two distinct files that happen to share a name can never collapse onto one device tuple; true duplicates are still caught by the backend's checksum-based dedup.

This PR merges to **`migration/immichv3`**, not `main` — it's one step of the Immich v3 retarget; the integration branch merges to `main` once all steps land. The branch is intentionally not standalone-green (documented pattern-constraint + import blockers), so this change is verified by scoped checks (ruff, a zero-new-errors pyright diff) plus the runnable subset of tests rather than a fully green suite.

## Test plan
- [x] `uv run pyright` on changed files — no new errors vs. the branch baseline (net −1; only pre-existing branded-ID/dimension mismatches un-masked)
- [x] `uv run pytest tests/unit/services/test_streaming_upload.py tests/unit/services/test_streaming_form_parser.py` — 18 passed, incl. a new test asserting the synthesized device fields (unique UUID + placeholder `device_id`) reach the Gumnut upload
- [x] `uv run pytest tests/unit/api/test_search.py::TestSearchRandom::test_all_dto_fields_have_a_disposition` — passes (confirms the `RandomSearchDto.deviceId` removal matches the regenerated model)
- [x] Converter + buffered-upload assertions authored and verified by inspection (their modules are blocked by the branch's known `Error1` / pattern-constraint blockers; they run once those clear)

### DIFF
--- a/routers/api/assets.py
+++ b/routers/api/assets.py
@@ -538,10 +538,8 @@ async def _upload_buffered(
                 detail=str(ve),
             )
 
-        # Gumnut's asset-create requires device_asset_id/device_id, which Immich
-        # v3 no longer sends; synthesize them (see GUMNUT_UPLOAD_DEVICE_ID). A
-        # unique device_asset_id avoids collapsing distinct assets onto one
-        # device tuple; true duplicates are still caught by checksum dedup.
+        # Synthesize the device fields the Gumnut API requires (see
+        # GUMNUT_UPLOAD_DEVICE_ID).
         device_asset_id = str(uuid4())
 
         try:
--- a/routers/api/assets.py
+++ b/routers/api/assets.py
@@ -27,6 +27,7 @@ from gumnut.types.asset_bulk_update_assets_params import Update, UpdateChange
 from gumnut.types.asset_response import AssetResponse
 
 from config.settings import Settings, get_settings
+from routers.api.constants import GUMNUT_UPLOAD_DEVICE_ID
 from routers.utils.cdn_client import DEFAULT_FORWARDED_HEADERS, stream_from_cdn
 from routers.utils.gumnut_client import (
     BULK_CHUNK_SIZE,
@@ -329,8 +330,6 @@ def _parse_datetime(value: str | None, fallback: datetime) -> datetime:
 
 
 class UploadFields(NamedTuple):
-    device_asset_id: str
-    device_id: str
     file_created_at: datetime
     file_modified_at: datetime
 
@@ -340,20 +339,16 @@ def _extract_upload_fields(fields: dict[str, str]) -> UploadFields:
 
     Raises ValueError if required fields are missing.
     """
-    device_asset_id = fields.get("deviceAssetId", "")
-    device_id = fields.get("deviceId", "")
     file_created_at_str = fields.get("fileCreatedAt", "")
 
-    if not device_asset_id or not device_id or not file_created_at_str:
-        raise ValueError(
-            "Missing required fields: deviceAssetId, deviceId, fileCreatedAt"
-        )
+    if not file_created_at_str:
+        raise ValueError("Missing required field: fileCreatedAt")
 
     file_modified_at_str = fields.get("fileModifiedAt") or None
     file_created_at = _parse_datetime(file_created_at_str, datetime.now(timezone.utc))
     file_modified_at = _parse_datetime(file_modified_at_str, file_created_at)
 
-    return UploadFields(device_asset_id, device_id, file_created_at, file_modified_at)
+    return UploadFields(file_created_at, file_modified_at)
 
 
 # Delay applied before emitting upload-success events for video uploads, to give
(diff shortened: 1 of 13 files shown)


tests/unit/services/test_streaming_upload.py
```diff
@@ -124,6 +120,30 @@ async def test_successful_upload(self):
...
+        with patch(
+            "services.streaming_upload._get_streaming_http_client",
+            return_value=mock_client,
+        ):
+            pipeline = StreamingUploadPipeline(request, "http://localhost:8000", "jwt")
+            await pipeline.execute(_extract_fields)
+
+        sent_data = mock_client.post.call_args.kwargs["data"]
+        # A unique UUID so distinct assets never collapse onto one device tuple.
+        UUID(sent_data["device_asset_id"])  # raises if not a valid UUID
```
taggartgorman, line 144: **[Consider]** The `device_asset_id` **uniqueness** invariant is never asserted — this test (and the buffered-path `test_upload_asset_success`) only check that the value parses as a valid UUID. But uniqueness across uploads is the whole justification for the change (the `constants.py` comment and PR body both lean on it). A regression that hoisted `uuid4()` to a module-level constant, or reused `GUMNUT_UPLOAD_DEVICE_ID` for both fields, would still be a valid UUID and pass this assertion while silently collapsing distinct assets onto one device tuple.

Cheap fix: execute the pipeline twice and assert the two synthesized `device_asset_id`s differ, pinning the per-upload-uniqueness guarantee. (Consider, not Should Fix — the backend's checksum dedup, not the device tuple, is the real collision safety net.)

taggartgorman (commented):
Automated review (multi-agent). **0 Must Fix, 0 Should Fix, 2 Consider** — both non-blocking.

The change is clean: it drops the removed `deviceAssetId`/`deviceId`/`unassignedFaces` fields, retypes `people` to `PersonResponseDto`, removes now-dead code, and shims the upload device fields the Gumnut API still requires. The central data-integrity question — whether a random per-upload `device_asset_id` breaks dedup/idempotency — resolves in the PR's favor: upload idempotency and `/bulk-upload-check` are checksum-based, and the one device-tuple-dependent endpoint (`/exist`) is already gone from the v3 models, so a fresh UUID per upload can't create duplicate rows or regress a live v3 flow. Both upload paths synthesize identically. Value check passes (removes dead code, aligns with the regenerated models, minimal scope).

Two optional suggestions inline.

routers/api/assets.py
```diff
@@ -536,15 +531,19 @@ async def _upload_buffered(
...
-                _extract_upload_fields(fields)
-            )
+            file_created_at, file_modified_at = _extract_upload_fields(fields)
         except ValueError as ve:
             raise HTTPException(
                 status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
                 detail=str(ve),
             )
 
+        # Gumnut's asset-create requires device_asset_id/device_id, which Immich
```
taggartgorman: **[Consider]** The synthesis rationale is now stated in three places: the `GUMNUT_UPLOAD_DEVICE_ID` docstring, this buffered-path comment, and the near-identical streaming-path comment in `services/streaming_upload.py`. The "unique `device_asset_id` avoids collapsing distinct assets onto one device tuple" point is repeated almost verbatim between the two call sites, and the checksum-dedup line restates the constant's own docstring.

Optional: move the uniqueness rationale into the `GUMNUT_UPLOAD_DEVICE_ID` docstring (the one thing it doesn't already cover) and shorten both call-site comments to a pointer like `# Synthesize the device fields the Gumnut API requires (see GUMNUT_UPLOAD_DEVICE_ID).` Marginal — the verbose per-site comments are consistent with this file's convention, and the code duplication is a single `uuid4()` line that is *not* worth extracting into a helper.

tests/unit/services/test_streaming_upload.py
taggartgorman: Fixed — added `test_device_asset_id_unique_per_upload`, which runs the pipeline twice and asserts the two synthesized `device_asset_id`s differ. A hoisted/reused UUID would now fail it.

routers/api/assets.py
taggartgorman: Fixed — moved the uniqueness/dedup rationale into the `GUMNUT_UPLOAD_DEVICE_ID` docstring (single source of truth); both call-site comments are now the same one-line pointer.

PR #252, opened 2026-07-07
Rework shared-links to Immich v3 token/key/slug access model

taggartgorman:
## What
- Drop `SharedLinkResponseDto.token` from all shared-link stub responses; drop the removed `password`/`token` query params from `GET /shared-links/me` (v3 keeps `key`/`slug` there).
- Drop the removed `key`/`slug` query params from `PUT /shared-links/{id}/assets`, `PUT /albums/{id}/assets`, and `PUT /albums/assets` (they were declared but unused). `GET /albums/{id}` **keeps** `key`/`slug` — v3 dropped only `withoutAssets` there.
- Pass the already-`UUID` `id`/`userId` values directly (no `str()` wrap) so dropping `token=` doesn't unmask pre-existing `str`→`UUID` typing errors.
- Add runnable v3-conformance guards: `tests/unit/api/test_shared_links.py` (`model_fields` + `inspect.signature`) and `tests/unit/api/test_albums_v3_params.py` (`ast.parse`-based, since `albums.py` can't be imported on this branch).
- Doc: record the runnable-test technique in the design doc's "Known blockers" section.

## Why
Immich v3 removed the shared-link token/key/slug access surface. `routers/api/shared_links.py` is a pure stub (returns fake data — no real anonymous-access logic), so this is a v3 **shape-conformance** change, not an auth reimplementation. Every param/field change was verified against the fetched Immich v3.0.0 GA OpenAPI spec and the regenerated models. `SharedLinkEditDto.changeExpiryTime` was already absent after the regen; no code referenced it.

This PR merges to `migration/immichv3`, **not** `main` — it is step 4 of the Immich v3 retarget; the integration branch merges to `main` once all steps land.

## Known follow-up (out of scope)
`POST /shared-links/login` is missing v3's optional `key`/`slug` query params — a **pre-existing** (pre-v3) gap, an *addition* rather than a removal, with no functional impact (FastAPI silently ignores unknown query params on the stub). Left out to keep this PR purely subtractive; can be picked up separately.

## Test plan
The `migration/immichv3` branch is intentionally red (documented `Error1`-import and pattern-constrained-DTO instantiation blockers), so verification is scoped rather than a green suite:
- [x] `uv run pyright routers/api/shared_links.py routers/api/albums.py` — zero new errors vs. baseline (net −5: shared_links 5→0, albums unchanged)
- [x] `uv run pytest tests/unit/api/test_shared_links.py tests/unit/api/test_albums_v3_params.py` — 7 passed
- [x] Grep sweep: no `token` in `shared_links.py`, no `changeExpiryTime` references, `key`/`slug` remain only on `get_album_info`
- [x] Regression: `test_search.py` results identical with and without this change (stash diff)

### DIFF
--- a/tests/unit/api/test_albums_v3_params.py
+++ b/tests/unit/api/test_albums_v3_params.py
@@ -1,8 +1,10 @@
 """Immich v3 query-param conformance for the albums asset-add endpoints.
 
 Immich v3 dropped the `key`/`slug` query params from `PUT /albums/{id}/assets`
-and `PUT /albums/assets`, while KEEPING them on `GET /albums/{id}` (v3 dropped
-only `withoutAssets` there). These guard that intentional asymmetry.
+and `PUT /albums/assets`, while KEEPING them on `GET /albums/{id}` (of that
+endpoint's params, v3 dropped only `withoutAssets` — which the adapter still
+declares as a documented no-op for client compatibility, so the parsed source
+below still shows it). These guard that intentional asymmetry.
 
 `routers/api/albums.py` cannot be imported on the `migration/immichv3` branch —
 it imports `Error1`, which the v3 model regen removed, so `inspect.signature`
@@ -42,7 +44,7 @@ def test_put_albums_assets_dropped_key_and_slug():
 
 
 def test_get_album_info_retains_key_and_slug():
-    """GET /albums/{id} keeps key/slug in v3 (only withoutAssets was dropped)."""
+    """GET /albums/{id} keeps key/slug in v3."""
     params = _param_names("get_album_info")
     assert "key" in params
     assert "slug" in params
--- a/tests/unit/api/test_albums_v3_params.py
+++ b/tests/unit/api/test_albums_v3_params.py
@@ -1,0 +1,48 @@
+"""Immich v3 query-param conformance for the albums asset-add endpoints.
+
+Immich v3 dropped the `key`/`slug` query params from `PUT /albums/{id}/assets`
+and `PUT /albums/assets`, while KEEPING them on `GET /albums/{id}` (v3 dropped
+only `withoutAssets` there). These guard that intentional asymmetry.
+
+`routers/api/albums.py` cannot be imported on the `migration/immichv3` branch —
+it imports `Error1`, which the v3 model regen removed, so `inspect.signature`
+(which requires importing the module) is unavailable here. Parsing the source
+with `ast` inspects the signatures without executing the module.
+"""
+
+import ast
+from pathlib import Path
+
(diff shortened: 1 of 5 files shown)


tests/unit/api/test_albums_v3_params.py
```diff
@@ -0,0 +1,48 @@
+"""Immich v3 query-param conformance for the albums asset-add endpoints.
+
+Immich v3 dropped the `key`/`slug` query params from `PUT /albums/{id}/assets`
+and `PUT /albums/assets`, while KEEPING them on `GET /albums/{id}` (v3 dropped
+only `withoutAssets` there). These guard that intentional asymmetry.
```
taggartgorman: Minor / non-blocking: this docstring says v3 "dropped only `withoutAssets`" from `GET /albums/{id}`, which is accurate about the v3 spec — but this file `ast`-parses `albums.py`, whose `get_album_info` still declares `withoutAssets` (intentionally retained as a documented no-op for client compatibility). A reader cross-checking the parsed source might briefly read the present `withoutAssets` as contradicting the docstring. Consider a clarifying clause, e.g. "(v3 dropped only `withoutAssets` there, though the adapter still declares it as a no-op for client compatibility)".

taggartgorman (commented):
Reviewed as step 4 of the Immich v3 retarget (targets `migration/immichv3`). Clean, tightly-scoped subtractive change: removes the v3-dropped shared-link token/key/slug surface and the unused `key`/`slug` params on the album/shared-link asset-add PUTs, and adds runnable guard tests that stay green despite the branch's documented instantiation blockers.

Verified independently:
- `SharedLinkResponseDto` has no `token` field and types `id`/`userId` as `UUID`, so dropping `token=` and the `str()` wraps is correct and serializes identically on the wire.
- `shared_links.py` imports cleanly (so the `inspect.signature` guards work); `albums.py` can't be imported (`Error1` removed), which justifies the `ast.parse` approach. Both new test files pass (7 tests).
- Dropped params were inert on the stub (`password=""` hardcoded, `token` fed a now-removed field), and FastAPI silently ignores unknown query params — no client behavior lost, no real auth surface changed.

No blocking findings; one optional clarity nit inline. LGTM.

tests/unit/api/test_albums_v3_params.py
taggartgorman: Fixed in db8b207 — clarified the module docstring to note the adapter still declares `withoutAssets` as a documented no-op (so the parsed source showing it no longer reads as a contradiction), and dropped the now-redundant `withoutAssets` parenthetical from the function docstring.

PR #253, opened 2026-07-07
Add Immich v3 Sync-v2 entity mappings to the sync stream

taggartgorman:
## What
- Extend the sync stream's existing faces V1/V2 dual pattern to **assets** and **albums**: new `gumnut_asset_to_sync_asset_v2` (= V1 with `duration` as integer milliseconds — the only asset payload delta) and `gumnut_album_to_sync_album_v2` (= V1 minus `ownerId`, which the v3 GA model dropped). Wire `AssetsV2`/`AlbumsV2` into `_SYNC_TYPE_ORDER` and route them in `convert_entity_to_sync_event`.
- Accept `AssetOcrV1`, `PartnerAssetsV2`, and `AlbumAssetsV2` as **emit-nothing no-ops** (Gumnut has no OCR data, no partner sharing, and album assets are the user's own — already streamed via `AssetsV2` + the unconditional `AlbumToAssetsV1` join). The client tolerates their absence.
- Set the now-**required** `SyncAssetV1.createdAt` (from the asset's upload time) at both construction sites, including the WebSocket upload-ready payload.
- Generalize the hardcoded "skip V1 faces when V2 requested" into a `_V1_SUPERSEDED_BY_V2` table covering assets, albums, and faces — no entity double-emits.
- Self-review follow-through: extend the album-cover causal-consistency override and the FK checkpoint map (`_GUMNUT_TYPE_TO_SYNC_TYPES`) to the V2 types, since the v3 client streams assets/albums exclusively as V2; add invariant tests that fail a half-wired future V2 addition.

## Why
The adapter already reports server version 3.0.x, so the v3 mobile client requests the V2 sync entity types. Without these mappings, **no assets or albums sync** for a v3 client (they were silently logged as "unsupported"). This is step 3 of the Immich v3 retarget (design doc §5 / §7.3).

## Base branch
Merges to **`migration/immichv3`**, not `main` — it's one step of the multi-PR Immich v3 retarget; the integration branch merges to `main` once all steps land.

## Notes
- Scope boundary: the now-dangling `AssetDeltaSyncDto` import in `routes.py` (from the removed `/sync/delta-sync` + `/sync/full-sync` endpoints) is a separate removed-endpoints cleanup, out of scope here.
- The `migration/immichv3` branch is intentionally red: the regenerated models carry `pattern`-constrained UUID/datetime fields that raise at instantiation under the pinned pydantic + Python 3.14 (tracked separately). So tests assert on `model_fields`, stream module constants, and `inspect.getsource` rather than constructing the DTOs, and verification is by inspection + scoped checks.

## Test plan
- [x] `uv run pytest tests/unit/api/sync/test_sync_v2.py` — 13 passed (payload-shape, converters, stream wiring, no-op acceptance, event routing, and cross-site invariants)
- [x] `uv run pyright` — net +2 vs base, entirely the pre-existing `str→UUID` branch-blocker family unmasked by adding the required `createdAt` (the new V2 converters add zero); no new defect class
- [x] No regression in pre-existing sync/asset tests (identical failure counts with and without the change)

### DIFF
--- a/routers/api/sync/converters.py
+++ b/routers/api/sync/converters.py
@@ -178,8 +178,8 @@ def gumnut_asset_to_sync_asset_v2(asset: AssetResponse, owner_id: str) -> SyncAs
     """Convert Gumnut AssetResponse to Immich SyncAssetV2 format.
 
     SyncAssetV2 is SyncAssetV1 with ``duration`` as integer milliseconds instead
-    of the interval string — the only payload difference between the two (Immich
-    v3 §5). Delegate to V1 and swap that one field.
+    of the interval string — the only payload difference between the two (see the
+    ``immich-v3-api-changes.md`` design doc, §5). Delegate to V1 and swap it.
     """
     fields = gumnut_asset_to_sync_asset_v1(asset, owner_id).model_dump()
     fields["duration"] = duration_ms(asset.duration)
--- a/routers/api/sync/converters.py
+++ b/routers/api/sync/converters.py
@@ -295,8 +295,8 @@ def gumnut_album_to_sync_album_v2(album: AlbumResponse, owner_id: str) -> SyncAl
     """Convert Gumnut AlbumResponse to Immich SyncAlbumV2 format.
 
     In the Immich v3 GA model SyncAlbumV2 is SyncAlbumV1 without ``ownerId``
-    (the design doc's "byte-identical to V1" predates the GA drop of that field).
-    Delegate to V1 and drop the field V2 no longer carries.
+    (see the ``immich-v3-api-changes.md`` design doc, §5). Delegate to V1 and
+    drop the field V2 no longer carries.
     """
     fields = gumnut_album_to_sync_album_v1(album, owner_id).model_dump()
     fields.pop("ownerId", None)
--- a/routers/api/sync/converters.py
+++ b/routers/api/sync/converters.py
@@ -16,15 +16,18 @@ from routers.immich_models import (
     AssetVisibility,
     SyncAlbumToAssetV1,
     SyncAlbumV1,
+    SyncAlbumV2,
     SyncAssetExifV1,
     SyncAssetFaceV1,
     SyncAssetFaceV2,
     SyncAssetV1,
+    SyncAssetV2,
     SyncAuthUserV1,
     SyncPersonV1,
     SyncUserV1,
 )
 from routers.utils.asset_conversion import (
+    duration_ms,
     exif_dims_and_orientation,
     format_duration,
     mime_type_to_asset_type,
@@ -148,6 +151,8 @@ def gumnut_asset_to_sync_asset_v1(asset: AssetResponse, owner_id: str) -> SyncAs
     return SyncAssetV1(
         id=str(safe_uuid_from_asset_id(asset.id)),
         checksum=resolve_immich_checksum(asset),
+        # "Uploaded to Immich at" — required on SyncAssetV1 in Immich v3.
+        createdAt=asset.created_at,
         isFavorite=False,  # Gumnut doesn't track favorites
         isEdited=False,
         originalFileName=asset.original_file_name,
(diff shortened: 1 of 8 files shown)


routers/api/sync/converters.py
```diff
@@ -274,6 +291,18 @@ def gumnut_album_to_sync_album_v1(album: AlbumResponse, owner_id: str) -> SyncAl
     )
 
 
+def gumnut_album_to_sync_album_v2(album: AlbumResponse, owner_id: str) -> SyncAlbumV2:
+    """Convert Gumnut AlbumResponse to Immich SyncAlbumV2 format.
+
+    In the Immich v3 GA model SyncAlbumV2 is SyncAlbumV1 without ``ownerId``
+    (the design doc's "byte-identical to V1" predates the GA drop of that field).
+    Delegate to V1 and drop the field V2 no longer carries.
```
taggartgorman: **[Consider]** The docstring's parenthetical re-derives history this same PR just removed from the design doc (which now says "minus `ownerId`", not "byte-identical"). The sibling asset V2 converter above states current behavior and cites §5 instead — mirroring that keeps the rationale in one place.

```suggestion
    In the Immich v3 GA model SyncAlbumV2 is SyncAlbumV1 without ``ownerId``
    (see the ``immich-v3-api-changes.md`` design doc, §5). Delegate to V1 and
    drop the field V2 no longer carries.
```

tests/unit/api/sync/test_sync_v2.py
```diff
@@ -0,0 +1,174 @@
...
+    assert "createdAt=" in inspect.getsource(
+        asset_conversion.build_asset_upload_ready_payload
+    )
+
+
+def test_asset_v2_converter_emits_int_ms_duration():
+    """The V2 asset converter swaps duration to int-ms via duration_ms."""
+    src = inspect.getsource(converters.gumnut_asset_to_sync_asset_v2)
+    assert "duration_ms" in src
+    assert 'fields["duration"]' in src
```
taggartgorman, line 79: **[Consider]** These two source-string assertions (this line and the `fields.pop("ownerId"` check just below at :85) couple to the exact local-variable name and call syntax rather than behavior: a behavior-preserving refactor (e.g. `del fields["ownerId"]`, or an inline `SyncAssetV2(**{**data, "duration": ...})`) would spuriously fail them, and a name→converter mis-wiring wouldn't fail them since both tokens still appear. They do catch outright line removal, and the paired `model_fields` invariant tests plus the value-tested `duration_ms` helper already carry the real signal — so this is belt-and-suspenders, not a coverage gap. Fine as-is under the red-branch constraint (behavioral converter tests are impossible while DTO instantiation raises); worth a follow-up to swap these for direct converter value tests once that blocker lifts.

taggartgorman (commented):
## Review: Sync-v2 entity mappings

Reviewed correctness, data integrity, security, silent failures, performance, test coverage, clarity, and minimalism across the diff plus the surrounding sync package (converters, event dispatch, FK-integrity, the stream loop, the `/sync/ack` path, and the regenerated models).

The core change is solid. The V2 converters correctly mirror the established `gumnut_face_to_sync_face_v2` delegate pattern — I verified the `model_dump()` round-trip: the models use camelCase field names with no separate aliases, so `SyncAssetV2(**fields)` and the `ownerId` pop behave as intended. `_V1_SUPERSEDED_BY_V2` keeps the stream single-emission for only-V1 / only-V2 / both-requested clients, and extending `_GUMNUT_TYPE_TO_SYNC_TYPES` to the V2 types correctly suppresses the spurious FK warning for a v3 client that checkpointed under V2 (the ack path already persists any valid `SyncEntityType`, so V2 checkpoints round-trip; it can't mask a real problem since actual FK safety is keyed on the unchanged gumnut-type string). The invariant tests that fail a half-wired future V2 addition are a nice touch, and generalizing the hardcoded face-skip into a table leaves the code cleaner than it found it.

No blocking issues — three non-blocking `[Consider]` nits (two inline, one below).

**[Consider]** `docs/architecture/sync-stream-architecture.md:3` — stale `last-updated: 2026-04-16`. This PR rewrites the "add a sync type version" checklist and adds the invariant-tests paragraph in the body, but leaves the frontmatter date unchanged — while the sibling `immich-v3-api-changes.md` was bumped to `2026-07-07` in this same PR. Worth bumping this one to `2026-07-07` too for consistency. (Anchored in the body rather than inline because line 3 isn't in the diff.)

routers/api/sync/converters.py
taggartgorman: Fixed in 0b7b74e — the docstring now cites the design doc §5 (matching the asset-V2 converter above) instead of re-deriving the byte-identical history this PR removed from that doc.

tests/unit/api/sync/test_sync_v2.py
taggartgorman: Skipping for this PR. Agreed the two source-string assertions are belt-and-suspenders; behavioral converter value tests are impossible while DTO instantiation raises under the pattern blocker (which is exactly why the source-string approach exists here). The model_fields invariants plus the value-tested duration_ms helper carry the real signal. Good follow-up to swap these for direct converter value tests once that separate blocker lifts.

PR #254, opened 2026-07-07
Strip pattern constraints from uuid/date-time fields in model generator

taggartgorman:
## What

- **Model generator** strips the `pattern` constraint from any OpenAPI schema whose `format` maps to a non-string Python type (`uuid`, `date-time`, `date`, `time`) before the spec reaches `datamodel-code-generator`, so codegen can't emit a string-only `pattern` on `UUID` / `AwareDatetime` / `date` / `time` fields. The transform lives in a new dependency-free module `tools/spec_preprocess.py` (`strip_non_string_patterns`), imported by the generator; genuine string-field patterns (email, hex color, PIN, etc.) are retained. `fetch_spec` is unified so URL and local-file inputs share one parse → strip → temp-file path (`json.dump(default=str)` keeps the write robust to native datetime scalars a YAML spec may carry).
- **Regenerate** `routers/immich_models.py` from the Immich v3.0.0 spec: 246 field-level patterns removed, and 8 now-redundant `RootModel[UUID]` id wrappers (`AlbumId`, `AssetId`, …) collapse to plain `UUID` (none referenced anywhere in the repo).
- **Tests**: `tests/unit/test_spec_preprocess.py` unit-tests the transform directly (each non-string format stripped, string patterns kept, nested recursion, and a dict-valued `format` guard); `tests/unit/test_immich_models.py` locks the committed artifact (uuid/date-time DTOs instantiate, a non-v4 UUID is accepted, patterns stripped on uuid/date-time but retained on string fields).
- **Docs/comments**: refresh the retarget design doc's "Known blockers" note, two sibling test docstrings, and the development-tools reference.

## Why

The v3 GA spec annotates id fields as `format: uuid` and timestamps as `format: date-time` while *also* carrying a string `pattern` (a v4-UUID / RFC3339 regex). `datamodel-code-generator` maps those formats to `UUID` / `AwareDatetime` and copies the string-only `pattern` onto the non-string field. Under the pinned pydantic + Python 3.14, validating a value for any such field raises `TypeError: Unable to apply constraint 'pattern' … for schema of type 'uuid'` (and `'datetime'`, `'date'`, `'time'`). The module imports fine, but every asset/album/user/sync response would 500 and every DTO-building test errored at setup — this was the blocker keeping the integration branch un-runnable end to end.

The `pattern` is a redundant re-encoding of what `format` already declares, so dropping it loses no meaningful validation (`UUID` / `AwareDatetime` still parse and reject bad input). Ids map to plain **`UUID`, not `UUID4`**, because the adapter reproduces the Gumnut API's UUIDs verbatim and does not control their version — a `UUID4` constraint could reject the adapter's own responses.

## Target branch

This PR merges to **`migration/immichv3`**, not `main` — it is a step in the Immich v3 retarget; the integration branch merges to `main` once all steps land. A sibling blocker (the `bulk.py` error-enum rename) remains and is tracked separately; with both fixed, the branch's test suite runs green.

## Test plan

- [x] `uv run pyright` — no new errors (branch error count unchanged at 170; the remainder is the intentionally-red branch's separate blockers)
- [x] `uv run pytest tests/unit/test_spec_preprocess.py tests/unit/test_immich_models.py` — 8 passed
- [x] Regenerated `immich_models.py` is identical to the prior committed version apart from the header timestamp (0 uuid/date-time patterns remain; 24 string-field patterns retained)
- [x] Full collectable suite: 677 passing with no regressions (remaining failures/collection errors are the separate branch blockers)


taggartgorman (commented):
## Review Summary

Solid fix for the v3 retarget blocker. Spec-preprocessing is the right layer (beats hand-editing generated output, which gets clobbered on regen, or pinning pydantic). I verified the strip is correct and complete for the current v3.0.0 artifact: 0 uuid/date-time patterns remain, all 24 retained patterns are genuine string fields (email, session token, hex color, PIN, HH:MM), no `example` payloads corrupted, and the eight `RootModel[UUID]` → `UUID` collapses are unreferenced. Plain `UUID` (not `UUID4`) is the correct call for reproducing Gumnut UUIDs verbatim.

Four comments below — two worth a look before this lands, two optional. **None affect the correctness of the committed artifact.**

**Should Fix**
- Strip covers `uuid`/`date-time` but not `date`/`time`, which raise the identical pydantic `TypeError` — a latent gap for future regenerations.
- Generator docstring (and the design-doc Resolved note) describe the error as firing at instantiation; it fires at value validation, contradicting this PR's own test file and `development-tools.md`.

**Consider**
- YAML fallback + always-`json.dump` can leak a temp file on a datetime scalar.
- Generator helpers are exercised only via the committed artifact, not directly.

tools/generate_immich_models.py
```diff
@@ -48,20 +48,60 @@ def get_container_tag() -> str:
...
+    generated DTO that carries an id or timestamp un-constructable. The regex is a
+    redundant re-encoding of what ``format`` already guarantees, so drop it before
+    codegen sees the spec while leaving ``pattern`` on genuine string fields intact.
+
+    Mutates ``node`` in place; returns the number of patterns removed.
     """
-    Fetch OpenAPI spec from URL or file path and return path to local file.
+    removed = 0
+    if isinstance(node, dict):
+        if node.get("format") in ("uuid", "date-time") and "pattern" in node:
```
taggartgorman: **[Should Fix]** The guard matches only `format: uuid`/`date-time`, but `format: date` and `format: time` raise the *identical* pydantic-core error (`TypeError: Unable to apply constraint 'pattern' … for schema of type 'date'`) at value validation — the same class of bug this pass exists to prevent.

Harmless for the current v3.0.0 artifact (its `date` fields carry no `pattern`), so this is latent: the generator is meant to be re-run against future Immich specs by container tag, and Immich already added v-patterns to uuid/date-time — the same on a `date` field would silently reintroduce the crash, and the "Stripped N" count wouldn't reflect the miss. `development-tools.md` even says to "extend that pass" for another non-string format that trips this error — `date`/`time` are already that class today.

```python
if node.get("format") in ("uuid", "date-time", "date", "time") and "pattern" in node:
```

tools/generate_immich_models.py
```diff
@@ -48,20 +48,60 @@ def get_container_tag() -> str:
...
 
 
-def fetch_spec(spec_path: str) -> tuple[str, bool, str | None]:
+def _strip_non_string_patterns(node: object) -> int:
+    """Remove ``pattern`` from schemas declared as ``format: uuid`` or ``date-time``.
+
+    datamodel-codegen maps ``format: uuid`` to ``UUID`` and ``format: date-time`` to
+    ``AwareDatetime``, then copies the (string-only) ``pattern`` constraint onto that
+    non-string field. pydantic-core rejects a ``pattern`` constraint on a uuid/datetime
+    schema with a ``TypeError`` when the model is first instantiated, which makes every
```
taggartgorman: **[Should Fix]** This says the error fires "when the model is first instantiated, which makes every generated DTO … un-constructable," but pydantic-core raises only at **value validation** — bare construction of a model whose only constrained field is optional succeeds (`SharedLinkEditDto()` is fine; only `SharedLinkEditDto(expiresAt=<datetime>)` trips it).

This contradicts two of this PR's own files, which state it correctly: `tests/unit/test_immich_models.py` ("applied when a datetime value is validated … no-argument construction is only a default-value smoke check") and `docs/references/development-tools.md` ("raise TypeError at value validation"). Worth aligning the wording here and in the design-doc Resolved note (`immich-v3-api-changes.md`) so all three agree — e.g. "pydantic-core rejects the constraint with a TypeError when a value for such a field is validated, so every real id/timestamp-bearing response 500s."

tools/generate_immich_models.py
```diff
@@ -120,7 +139,17 @@ def fetch_spec(spec_path: str) -> tuple[str, bool, str | None]:
...
+        source_label = path.name
+        raw_text = path.read_text()
+
+    spec_data = _parse_spec(raw_text)
+    removed = _strip_non_string_patterns(spec_data)
+    print(f"Stripped {removed} pattern constraint(s) from uuid/date-time fields")
+
+    # Write the preprocessed spec to a temp file for datamodel-codegen to consume.
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".json", delete=False) as f:
+        json.dump(spec_data, f, indent=2)
```
taggartgorman: **[Consider]** The new YAML fallback in `_parse_spec` can produce native `datetime` objects (from unquoted RFC3339 scalars, common in OpenAPI `example:` values), which `json.dump` can't serialize — raising an uncaught `TypeError` here. Because the temp file is `delete=False` and the exception escapes before `fetch_spec` returns, `main`'s `finally` never runs, so the temp file leaks and the tool aborts with a raw traceback instead of the tidy `sys.exit(1)` used elsewhere. Low likelihood since real inputs are JSON, but a cheap fix in just-refactored code:

```python
json.dump(spec_data, f, indent=2, default=str)
```

tests/unit/test_immich_models.py
```diff
@@ -0,0 +1,73 @@
...
+
+Guards the pattern-strip fix in ``tools/generate_immich_models.py`` (see
+``_strip_non_string_patterns`` for the mechanism): the v3 spec puts string ``pattern``
+constraints on ``format: uuid`` / ``date-time`` fields, which pydantic-core rejects
+with a ``TypeError`` when a value for such a field is validated — so every DTO carrying
+a populated id or timestamp is un-constructable. These tests lock the committed
+artifact: uuid/date-time DTOs must construct, and patterns on genuine string fields
+must survive.
+
+We assert against the committed generated module rather than importing the generator,
```
taggartgorman, line 11: **[Consider]** These tests assert against the committed generated artifact, never `_strip_non_string_patterns` / `_parse_spec` directly (the module can't import here because `requests`/`click`/`pyyaml` aren't in the test venv — as this docstring notes). That guards the runtime guarantee well, but the generator's own logic is unguarded: editing the `("uuid", "date-time")` tuple or the dict-vs-list recursion and running pytest *without* regenerating still passes all four tests against the stale file. Optional follow-up: add a dev-dependency group (or extract the two stdlib-only helpers into a dependency-free module) so the pure functions can be unit-tested with hand-built spec dicts — which would also let the `date`/`time` gap in the comment above be covered directly.

tools/generate_immich_models.py
taggartgorman: Fixed — broadened the stripped formats to `{uuid, date-time, date, time}` (verified `date`/`time` raise the identical `TypeError` at value validation). The logic now lives in `tools/spec_preprocess.py` and is covered directly by `test_spec_preprocess.py`.

tools/generate_immich_models.py
taggartgorman: Fixed — the strip function docstring (now in `tools/spec_preprocess.py`), the design-doc *Resolved* note, and `development-tools.md` all say "value validation" now, matching the test file.

tools/generate_immich_models.py
taggartgorman: Fixed — added `default=str` to the `json.dump`. Verified `yaml.safe_load` yields a native `datetime` that plain `json.dump` rejects (which would leak the `delete=False` temp file since `main`s cleanup runs only after `fetch_spec` returns).

tests/unit/test_immich_models.py
taggartgorman: Done — extracted the stdlib-only `strip_non_string_patterns` into `tools/spec_preprocess.py` and added `tests/unit/test_spec_preprocess.py`, which loads it by path and tests each format directly (incl. a dict-valued `format` guard). This is what makes the new `date`/`time` cases testable, since the committed-artifact test can not exercise them.

PR #255, opened 2026-07-07
Drop Immich v3 removed endpoints (clean cut, no shims)

taggartgorman:
## What
Immich v3 removed 7 endpoints. This drops their handlers from the adapter (clean cut, no shims):

- `POST /sync/delta-sync`, `POST /sync/full-sync` — Sync v2 supersedes them over the existing `/sync/stream`
- `POST /assets/exist`
- `GET /assets/random` — use `POST /search/random`
- `GET /assets/device/{deviceId}`
- `PUT /assets/{id}/original` — **method drop only**; the `GET /assets/{id}/original` download on the same path stays
- `GET /server/theme` — also removed from the auth-middleware unauthenticated-path set

`GET /plugins/triggers` was never implemented in the adapter, so there was nothing to drop.

Also removes the now-orphaned imports of DTOs deleted from the v3 models (`AssetDeltaSync*`, `AssetFullSyncDto`, `CheckExistingAssets*`, `AssetMediaReplaceDto`, `ServerThemeDto`), drops the four dead handler test classes, and records the drop decision in the v3 API-change and gap-analysis design docs.

## Why
The adapter retarget to Immich 3.0 is a clean cut with no dual-support window (product is in alpha), so removed endpoints need not stay as shims — none of these paths exist in the v3 spec, so no supported v3 client (Immich mobile/web v3) calls them. Keeping dead handlers that import deleted model symbols also left latent breakage.

**This PR merges to `migration/immichv3`, not `main`** — it is one step of the multi-PR Immich v3 retarget; the integration branch merges to `main` once all steps land.

## Test plan
- [x] `uv run ruff check` / `ruff format` clean
- [x] `uv run pyright`: 162 errors, down 8 from the branch baseline of 170 — **zero new** errors
- [x] Node-ID diff of pytest failures vs the base commit: **0 previously-passing tests regressed**
- [x] Removing the dead deleted-DTO imports unblocks test collection (15 → 11 collection errors); the 24 sync ack/session tests that directly exercise the edited router all pass
- Note: `migration/immichv3` is intentionally red due to a separate `Error1` import blocker (a different issue). The 5 newly-visible `test_sync_stream.py` failures are pre-existing `SyncAssetV1.createdAt` Mock-fixture issues from the merged Sync-v2 work — un-masked by this PR unblocking collection, not caused by it.

### DIFF
--- a/docs/architecture/adapter-architecture.md
+++ b/docs/architecture/adapter-architecture.md
@@ -1,6 +1,6 @@
 ---
 title: "Immich Adapter Architecture"
-last-updated: 2026-07-03
+last-updated: 2026-07-07
 ---
 
 # Immich Adapter Architecture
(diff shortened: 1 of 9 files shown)


taggartgorman (commented):
**Reviewed: Drop Immich v3 removed endpoints (clean cut, no shims)**

The removal is clean and well-scoped. Verified:
- No dangling references to any removed handler or DTO; all removed imports are genuinely unused in retained code (`List`, `AssetMediaResponseDto`, etc. correctly kept).
- `GET /assets/{id}/original` download is intact after the `PUT` drop (FastAPI keys routes by path+method).
- Removing `/api/server/theme` from `UNAUTHENTICATED_PATHS` is strictly more restrictive — exact-match set, endpoint fully gone, no bypass and no effect on the other paths.
- The sync-suite test failures are pre-existing `SyncAssetV1.createdAt` Mock-fixture issues un-masked by this PR unblocking collection (the base branch's sync suite failed to collect entirely) — not regressions.

One non-blocking doc follow-up:

**[Should Fix]** `docs/architecture/adapter-architecture.md` — stale "Fully implemented" endpoint table (this file isn't in the PR's diff, so noting it here rather than inline)

AGENTS.md designates this doc as the "endpoint status" reference, and this PR's removals make two rows factually wrong:
- **Sync row (`docs/architecture/adapter-architecture.md:384`)**: "Full sync, delta sync, stream, ack" → the full-sync and delta-sync handlers were deleted. Should read "stream, ack".
- **Assets row (`docs/architecture/adapter-architecture.md:377`)**: lists "existence check" → `POST /assets/exist` was deleted. Should drop "existence check".

A reader consulting this table would believe the adapter still serves endpoints that now return 404. The other removed endpoints (random/device/replace/theme) were stubs and correctly aren't in this table. Worth a 2-row edit in this PR to keep the endpoint-status doc accurate.

PR #260, opened 2026-07-08
Retarget test_assets.py removed Action enum to v3 AssetUploadAction

taggartgorman:
## What
- Repoint `tests/unit/api/test_assets.py` from the removed anonymous `Action` enum to the Immich v3 `AssetUploadAction` (same `accept`/`reject` values): one import line plus the five `Action.accept`/`Action.reject` assertion sites.
- Test-only — no production code references the old name.

## Why
- The Immich v3 GA model regeneration renamed the anonymous bulk-upload-check enum, so the stale import made `test_assets.py` fail collection with an `ImportError` — the suite's last remaining collection error on the integration branch. With this change the module collects again (147 tests) and the retargeted enum assertions pass.
- `AssetUploadAction` types the `action` field on the bulk-upload-check result DTO, so the enum-member comparisons carry over unchanged.
- This PR merges to `migration/immichv3`, not `main` — it is one step of the Immich v3 retarget; the integration branch merges to `main` when the whole migration lands.
- Out of scope, per the tracked plan: the 9 pre-existing v3-migration failures the module now reveals (id `str`→`UUID` assertion residue on untouched lines) are part of a separate mechanical sweep.

## Test plan
- [x] `uv run pytest tests/unit/api/test_assets.py --collect-only` — collects 147 tests, no `ImportError`
- [x] `grep -n '\bAction\b' tests/unit/api/test_assets.py` — no bare `Action` reference remains
- [x] `uv run pytest tests/unit/api/test_assets.py` — 138 passed / 9 failed, all 9 confirmed pre-existing str→UUID residue on lines this diff does not touch
- [x] `uv run pyright tests/unit/api/test_assets.py` — zero new errors vs. base (stash-diff baseline); fixes one baseline error (the unknown `Action` import)

### DIFF
--- a/tests/unit/api/test_assets.py
+++ b/tests/unit/api/test_assets.py
@@ -42,7 +42,6 @@ from routers.api.assets import (
 )
 from routers.utils.gumnut_id_conversion import uuid_to_gumnut_asset_id
 from routers.immich_models import (
-    Action,
     AssetBulkUploadCheckDto,
     AssetBulkUploadCheckItem,
     AssetCopyDto,
@@ -57,6 +56,7 @@ from routers.immich_models import (
     AssetMetadataUpsertItemDto,
     AssetMediaSize,
     AssetMediaStatus,
+    AssetUploadAction,
 )
 
 
@@ -88,7 +88,7 @@ class TestBulkUploadCheck:
 
         # Assert
         assert len(result.results) == 2
-        assert all(item.action == Action.accept for item in result.results)
+        assert all(item.action == AssetUploadAction.accept for item in result.results)
         assert result.results[0].id == "asset1"
         assert result.results[1].id == "asset2"
         mock_client.assets.check_existence.assert_called_once()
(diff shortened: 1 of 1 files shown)


taggartgorman (commented):
**Review: no blocking findings** (0 Must Fix / 0 Should Fix / 0 Consider).

_Posted as a COMMENT rather than a formal approval — GitHub does not allow approving your own PR._

Verified every claim against the code, not just the diff:

- `AssetUploadAction` (`immich_models.py:313-315`) has identical `accept`/`reject` values to the removed `Action` enum, and types `AssetBulkUploadCheckResult.action` (`immich_models.py:4039`) — so the enum-member comparisons carry over unchanged.
- No bare `Action` reference remains in `test_assets.py`, and the enum is fully removed from `immich_models.py`.
- `uv run ruff check tests/unit/api/test_assets.py` → clean. Import placement follows the block's existing convention.
- Ran the retargeted tests: the `action == AssetUploadAction.*` assertions **pass**. The failures that remain land on the untouched `assetId == str(sample_uuid)` lines (e.g. line 169) — the documented, out-of-scope `str`→`UUID` residue, not a regression from this change.

**Value check:** restores collection of the previously-uncollectable module (was a hard `ImportError`), at exactly minimal scope. LGTM.

PR #261, opened 2026-07-08
Clear Immich v3 test residue: 147 failures to green

taggartgorman:
## What
- Adapts tests to the regenerated v3 models — ids typed `UUID`, stricter `AwareDatetime` and pattern validation: UUID-vs-str assertion updates, license fixtures using the shared stub constant, real `created_at` datetimes in sync mocks (`SyncAssetV1` gained a required `createdAt`), valid UUIDs replacing sentinel id strings, and a `RandomSearchDto` rating fixture bumped to the new `ge=1` floor (16 test files). New pinned tests: `minFaces == 3`, `SyncAssetV1.createdAt` mapping (distinct sentinel), `realtimeTranscoding` in the unimplemented-flags tuple.
- Fixes the real production defects the newly-collecting suite exposed:
  - `album_conversion.py` — `albumThumbnailAssetId` used `""` as the no-cover sentinel; v3 types it `UUID | None`, so every coverless album failed response validation. Emit `None`.
  - `server.py` — supply newly required fields: `ServerVersionResponseDto.prerelease`, `ServerConfigDto.minFaces`, `ServerFeaturesDto.realtimeTranscoding`.
  - `oauth.py` — `LoginResponseDto.userId` received the raw Gumnut user id; route it through the existing `safe_uuid_from_user_id` conversion.
  - `people.py` — bulk people update double-wrapped an already-`UUID` id (`UUID(UUID(...))` → `AttributeError`), crashing every `PUT /api/people`; malformed ids are now rejected with 422 at the DTO boundary, and the unreachable `ValueError` branch is removed.
  - **License stubs** — keys failing v3's `UserLicense` pattern 500ed their endpoints; now a shared `STUB_LICENSE_KEY` in `routers/api/constants.py` (carrying the pattern rationale) used at all 11 production sites.
  - **Stub id literals** — non-UUID string ids in stub endpoints (`admin`, `tags`, `activities`, `libraries`, `memories`, `api_keys`, `stacks` incl. `primaryAssetId`) were latent 500s under v3 response validation; replaced with `uuid4()` (11 sites, enumerated via pyright's literal-assignment errors).
  - `current_user.py` — `get_current_user_id` double-wrapped the now-`UUID` `UserAdminResponseDto.id`; dead `str()` wraps dropped at two UUID construction sites.
- Documents the post-regeneration pyright sweep in `docs/references/development-tools.md`.

## Why
- These failures were pre-existing v3 residue, masked until the app could import and the suite could collect (unblocked by the previous integration-branch PRs). Several were real user-facing crashes under v3 — coverless albums, bulk people updates, and every stub endpoint returning a literal that the regenerated models now reject.
- The generated v3 models are kept strict throughout: no `UUID`/`AwareDatetime` typing or pattern was weakened; tests were adapted to the contract and production call sites fixed to honor it.
- This PR merges to `migration/immichv3`, not `main` — it is part of the Immich v3 retarget; the integration branch merges to `main` when the whole migration lands.

## Known verification item (deliberately unchanged)
- Three sites fall back to `""` for a missing user email and now feed `EmailStr`-typed fields (`oauth.py` login, `current_user.py`, `users.py`). Accounts are Clerk-backed, so a null email is assumed impossible; if that assumption ever breaks, these 500 loudly. Decision: leave as-is this PR, verify the upstream guarantee separately.

## Follow-up candidates (pre-existing, untouched)
- `jobs.py` / `system_config.py` have the missing-required-field family of latent 500s (visible in the pyright backlog) — same residue-burndown series, separate change.

## Test plan
- [x] `uv run pytest` — **1125 passed / 0 failed** (was 147 failed / 976 passed)
- [x] `uv run ruff check` / `uv run ruff format --check` — clean
- [x] `uv run pyright` — 116 errors vs. 155 on base: 39 fixed, zero new (list-diffed against base)
- [x] Untested stub endpoints smoke-tested via TestClient (`GET /api/server/license`, `GET /api/stacks/{id}` → 200; both previously 500)
- [x] Two independent review passes (pre-PR + two self-review iterations this round); all findings fixed or explicitly adjudicated

### DIFF
--- a/routers/api/people.py
+++ b/routers/api/people.py
@@ -134,9 +134,10 @@ async def _update_one_person(
     `GumnutError`, raised from either `_resolve_thumbnail_face_id`'s inner
     `client.faces.list` or the final `client.people.update`) is delegated to
     `classify_bulk_item_call`, mirroring the per-chunk policy used by
-    `chunked_per_item_bulk` for chunked bulk endpoints. The pre-call
-    exception specific to this endpoint (`HTTPException` from
-    `_resolve_thumbnail_face_id`'s "missing face" branch) stays here.
+    `chunked_per_item_bulk` for chunked bulk endpoints. The exception
+    specific to this endpoint (`HTTPException` from
+    `_resolve_thumbnail_face_id`'s "missing face" branch) stays here — it
+    is not an SDK error, so `classify_bulk_item_call` lets it propagate.
     """
     log_extra = {"person_id": person_item.id}
 
--- a/routers/api/people.py
+++ b/routers/api/people.py
@@ -135,24 +135,14 @@ async def _update_one_person(
     `client.faces.list` or the final `client.people.update`) is delegated to
     `classify_bulk_item_call`, mirroring the per-chunk policy used by
     `chunked_per_item_bulk` for chunked bulk endpoints. The pre-call
-    exceptions specific to this endpoint (`ValueError` from UUID parsing,
-    `HTTPException` from `_resolve_thumbnail_face_id`'s "missing face"
-    branch) stay here.
+    exception specific to this endpoint (`HTTPException` from
+    `_resolve_thumbnail_face_id`'s "missing face" branch) stays here.
     """
     log_extra = {"person_id": person_item.id}
 
-    try:
-        # immich openapi specs switch between str and UUID for people id, so
-        # UUID(...) can raise on malformed input.
-        gumnut_person_id = uuid_to_gumnut_person_id(UUID(person_item.id))
-    except ValueError as ve:
-        logger.warning(
-            "Invalid person id in bulk update",
-            extra={**log_extra, "error": str(ve)},
-        )
-        return BulkIdResponseDto(
-            id=person_item.id, success=False, error=BulkIdErrorReason.unknown
-        )
+    # v3 types `PeopleUpdateItem.id` as UUID, so malformed ids are rejected
+    # with a 422 at the request boundary before reaching this handler.
+    gumnut_person_id = uuid_to_gumnut_person_id(person_item.id)
 
     try:
         sdk_error = await classify_bulk_item_call(
@@ -396,7 +386,7 @@ async def merge_person(
     )
 
     return [
-        BulkIdResponseDto(id=str(source_uuid), success=True, error=None)
+        BulkIdResponseDto(id=source_uuid, success=True, error=None)
         for source_uuid in request.ids
     ]
 
--- a/routers/api/server.py
+++ b/routers/api/server.py
@@ -53,6 +53,8 @@ server_features = {
     "configFile": False,
     "email": False,
     "ocr": False,
+    # No server-side transcoding; assets are served as stored.
+    "realtimeTranscoding": False,
 }
 
 
@@ -77,6 +79,9 @@ def get_fake_config() -> dict:
         "publicUsers": True,
         "mapDarkStyleUrl": "https://tiles.immich.cloud/v1/style/dark.json",
         "mapLightStyleUrl": "https://tiles.immich.cloud/v1/style/light.json",
+        # Immich's people-recognition server default; face detection runs in
+        # the Gumnut API, so this only feeds the client's settings display.
+        "minFaces": 3,
     }
 
 
@@ -276,6 +281,8 @@ async def get_server_version() -> ServerVersionResponseDto:
         major=version.major,
         minor=version.minor,
         patch=version.patch,
+        # Tracked container tags are GA releases, never pre-releases.
+        prerelease=None,
     )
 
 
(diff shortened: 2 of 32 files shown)


taggartgorman (commented):
**Review — Immich v3 test residue cleanup**

Reviewed the full diff (6 production files + 15 test files) against the regenerated v3 models. The production fixes are correct and contract-aligned:

- **`oauth.py` / `current_user.py`**: `userId` and `get_current_user_id` now return the `UUID` the v3 DTOs require (raw Gumnut id / double-`UUID(...)` removed). Verified against `LoginResponseDto.userId: UUID` and `UserAdminResponseDto.id: UUID`; the oauth value round-trips through `safe_uuid_from_user_id` to match the persisted session id and `/users/me`.
- **`people.py`**: removing the `try/except ValueError` is safe — `PeopleUpdateItem.id` is now `UUID`, so the branch is unreachable; malformed ids 422 at the DTO boundary (matches upstream Immich's own `@IsUUID()`).
- **`album_conversion.py`**: `None` (not `""`) for no-cover matches `albumThumbnailAssetId: UUID | None`.
- **`server.py`**: `prerelease` / `minFaces` / `realtimeTranscoding` supplied for the newly-required v3 fields.
- **License stubs**: the new `IMSV-…` value matches the v3 `UserLicense` pattern `^IM(SV|CL)(-[\dA-Za-z]{4}){8}$`; the old slash-wrapped form did not (the cause of the 500s).

No blocking issues (0 Must Fix / 0 Should Fix). Three non-blocking suggestions inline.

_FYI (not posted inline):_ the replacement `test_update_people_malformed_uuid_rejected_at_dto_boundary` asserts pydantic raises rather than exercising the endpoint's 422 — but that path is pure FastAPI validation, so an end-to-end test would largely test the framework. Reasonable to leave as-is.

routers/api/server.py
```diff
@@ -53,6 +53,8 @@
     "configFile": False,
     "email": False,
     "ocr": False,
+    # No server-side transcoding; assets are served as stored.
+    "realtimeTranscoding": False,
```
taggartgorman, line 58: **[Consider]** Pin `realtimeTranscoding` to `False` so an accidental flip to `True` is caught.

This new flag is only incidentally covered — the endpoint returning 200 validates presence + bool-ness, not the value. `tests/integration/test_server_features.py::test_unimplemented_features_are_disabled` exists precisely to assert unimplemented flags read `False`; adding `"realtimeTranscoding"` to its tuple pins it cheaply.

routers/api/people.py
```diff
@@ -135,24 +135,14 @@ async def _update_one_person(
     `client.faces.list` or the final `client.people.update`) is delegated to
     `classify_bulk_item_call`, mirroring the per-chunk policy used by
     `chunked_per_item_bulk` for chunked bulk endpoints. The pre-call
-    exceptions specific to this endpoint (`ValueError` from UUID parsing,
-    `HTTPException` from `_resolve_thumbnail_face_id`'s "missing face"
-    branch) stay here.
+    exception specific to this endpoint (`HTTPException` from
```
taggartgorman: **[Consider]** Drop the now-stale "pre-call" qualifier in the docstring (it's on the preceding line).

Removing the `ValueError` mention was right, but "pre-call" originally distinguished that `ValueError` (raised before the `try` block) from the SDK errors. The only remaining exception — `HTTPException` from `_resolve_thumbnail_face_id`'s "missing face" branch — is raised *inside* `_do_person_update` (during the wrapped call), so the real discriminator is exception *type* (HTTPException vs SDK error), not call timing. Suggested wording:

> ... for chunked bulk endpoints. The exception specific to this endpoint (`HTTPException` from `_resolve_thumbnail_face_id`'s "missing face" branch) stays here (it is not an SDK error, so `classify_bulk_item_call` lets it propagate).

routers/api/server.py
```diff
@@ -299,7 +306,8 @@ async def get_server_license() -> LicenseResponseDto:
     This is a stub implementation returning basic license info.
     """
     return LicenseResponseDto(
-        licenseKey="/IMSV-AAAA-AAAA-AAAA-AAAA-AAAA-AAAA-AAAA-AAAA/",
+        # Must match the v3 UserLicense key pattern or the endpoint 500s.
+        licenseKey="IMSV-AAAA-AAAA-AAAA-AAAA-AAAA-AAAA-AAAA-AAAA",
```
taggartgorman: **[Consider]** Optional: hoist the stub license key into a shared constant.

`IMSV-AAAA-…` is now inlined at 11 production sites, and the load-bearing "must match the v3 pattern or the endpoint 500s" rationale lives at only this one. A `STUB_LICENSE_KEY` constant (e.g. in `routers/api/constants.py`, alongside the existing named constants) would carry the value + rationale once. Defensible to leave as-is — sibling stub fields like `activationKey` are also inlined per existing convention; the pattern constraint is the only thing tipping this toward centralizing.

routers/api/server.py
taggartgorman: Fixed — `realtimeTranscoding` added to the unimplemented-flags tuple in `test_unimplemented_features_are_disabled`, so a flip to `True` now fails the suite.

routers/api/people.py
taggartgorman: Fixed — reworded per the suggestion: the docstring now says the HTTPException stays here because it is not an SDK error (so `classify_bulk_item_call` lets it propagate), dropping the stale pre-call framing. Verified against `classify_bulk_item_call`'s catch set before adopting.

routers/api/server.py
taggartgorman: Fixed — `STUB_LICENSE_KEY` added to `routers/api/constants.py` (carrying the pattern rationale once) and used at all 11 production sites plus the 5 test fixtures. Note: `routers/utils/current_user.py` now imports it from `routers.api.constants` — the first utils→api import, but constants.py is a pure-constants module so no cycle is possible.

PR #262, opened 2026-07-08
Add v3 createdAt array to timeline bucket response

taggartgorman:
## Summary

Immich v3's `TimeBucketAssetResponseDto` requires a `createdAt` array — "Array of UTC timestamps when each asset was originally uploaded" — which the adapter previously omitted from `GET /api/timeline/bucket`. This PR populates it.

## Changes

- **`routers/utils/asset_conversion.py`**: new `resolve_created_at` helper alongside the existing capture-time resolvers. It maps the Gumnut asset's `created_at` (upload time) through `to_actual_utc`, matching the semantics of the upstream Immich server, which emits this field as the actual-UTC upload instant.
- **`routers/api/timeline.py`**: `get_time_bucket` builds and returns the `createdAt` array, formatted like the other bucket timestamp arrays (ISO 8601, no timezone suffix, exactly 3 digits of milliseconds). The loop's `created_at` variable is renamed to `captured_at` now that two different timestamps flow through it.
- **`tests/unit/api/test_timeline.py`**: `test_get_time_bucket_success` pins the new field, including the UTC conversion (`createdAt` is the actual-UTC instant, distinct from `fileCreatedAt`'s wall-clock convention — a `-05:00` capture asserts `15:00` upload vs `10:00` capture). The missing-attributes test mock now carries a real `created_at`, since the Gumnut SDK declares the field non-nullable.

## Out of scope, documented

While verifying the new field against upstream, we found that the bucket's pre-existing `fileCreatedAt` emits capture wall-clock time where Immich v3 expects actual UTC (v3 clients reconstruct display time as `fileCreatedAt + localOffsetHours`, so known-timezone captures display shifted). That bug and the related fractional-offset truncation in `localOffsetHours` predate this change and are tracked separately; this PR adds a code comment marking the gap so the correct UTC handling on the adjacent `createdAt` lines doesn't read as an inconsistency.

## Testing

- `uv run pytest` — 1125 passed
- `uv run ruff check` / `uv run ruff format --check` — clean

### DIFF
--- a/routers/api/timeline.py
+++ b/routers/api/timeline.py
@@ -17,6 +17,7 @@ from routers.utils.asset_conversion import (
     duration_ms,
     mime_type_to_asset_type,
     resolve_capture_datetime,
+    resolve_created_at,
 )
 from routers.utils.current_user import get_current_user_id
 from routers.utils.gumnut_client import get_authenticated_gumnut_client
@@ -207,6 +208,11 @@ async def get_time_bucket(
     asset_count = len(filtered_assets)
 
     asset_ids = []
+    created_at_list = []  # Upload timestamps (actual UTC)
+    # Capture timestamps, currently emitted as wall-clock local time. Immich v3
+    # expects actual UTC here (clients derive local time by adding
+    # localOffsetHours), so known-timezone captures display shifted — a known
+    # gap, fixed separately from the createdAt addition.
     file_created_at_list = []
     is_image_list = []
     ratio_list = []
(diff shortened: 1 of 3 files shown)


taggartgorman (commented):
**Review: clean pass — no blocking or non-blocking findings.**

(Posted as a COMMENT review because GitHub does not permit an APPROVE event on one's own PR; the review below is otherwise an approval on the merits.)

### What the PR does
Populates the required `createdAt: list[str]` field on Immich v3's `TimeBucketAssetResponseDto`, which `GET /api/timeline/bucket` previously omitted. Each entry is the asset's Gumnut upload time (`created_at`) converted to actual UTC via a new `resolve_created_at` helper, formatted like the sibling timestamp arrays.

### Verification
- `resolve_created_at`'s `to_actual_utc(created_at)` cannot hit `None`: the SDK types `AssetResponse.created_at` as a required, non-nullable `datetime`. The test-mock change from `created_at = None` to a real datetime is both required (the old `None` would crash the new `.strftime`) and more faithful to the SDK contract.
- The `to_actual_utc` wrap is intentional: `.strftime()` prints wall-clock in the datetime's own tz, so wrapping first makes it emit the true UTC instant — mirroring the existing `resolve_file_created_at` pattern. The test pins this with `-05:00` and `+02:00` captures, distinct from `fileCreatedAt`'s wall-clock convention.
- The `captured_at` rename is complete across all four loop references; no collision with the new `created_at_list`.
- The alphabetized return dict now matches the DTO's field order; every key still maps to its correct parallel array (verified key-by-key). Cosmetic, zero functional risk.
- Local run: 48 timeline tests pass; `ruff check` / `ruff format --check` clean; `pyright` shows only pre-existing str→UUID / float→int errors in untouched functions — the new helper and `timeline.py` changes are type-clean.

**Findings: 0 Must Fix · 0 Should Fix · 0 Consider.**

PR #264, opened 2026-07-09
Clear all 116 pyright errors from the Immich v3 model regeneration

taggartgorman:
## Summary

The v3 model regeneration left 116 pyright errors on the integration branch. This PR clears all of them (116 → 0), in five commits grouped by root cause. Most were static-only residue of the v3 spec tightening (pydantic coerces at runtime, so the suite stayed green), but the sweep surfaced and fixes **three real runtime bugs** the type checker caught:

1. **`GET /api/system-config` and `GET /api/jobs` returned 500 on every request.** v3 added required config fields (`ffmpeg.realtime`, `job.integrityCheck`, `newVersionCheck.channel`, three OAuth fields, the `integrityChecks` job tree, the `integrityCheck` queue) that the stubs didn't supply, so response construction raised `ValidationError`. Neither file had test coverage. New values mirror the Immich v3.0.0 server defaults. Two adjacent crashes in the same builder were masked: `externalDomain`/`mobileRedirectUri` are now typed `str` (stale `AnyUrl(...)` wrappers fail validation), and `machineLearning.urls` gained `min_length=1` — a value constraint invisible to pyright, caught by the new smoke test.

2. **Upload/update WebSocket events were silently dropped.** v3 types `UserResponseDto.id` as `UUID`, so the upload and asset-update paths began passing a `UUID` object to `emit_user_event`, which uses the value as the socket.io room key. Rooms are registered at connect time with the UUID *string* form, and socket.io matches rooms by exact value — `UPLOAD_SUCCESS`, `ASSET_UPLOAD_READY_V1`, and `ASSET_UPDATE` emitted into an empty room with no error anywhere. Fixed at the seam: the emit helpers accept `str | UUID` and normalize where the room key is built, with a regression test pinning the normalization.

## Commits

| Commit | Bucket | Errors |
|---|---|---|
| `0a70959` | New required v3 config fields → live 500s in system-config/jobs stubs, plus construction smoke tests | 7 |
| `dd5fe35` | WebSocket room-key normalization (silent event drop) | 4 + 1 |
| `db4be94` | str → UUID sweep: pass native UUIDs to v3 UUID-typed id fields; retype the sync-converter `owner_id`/`asset_uuid` chain end-to-end | 71 |
| `7bbd5d1` | float → int: drop v2-era `float()` casts on asset dims (the Gumnut SDK already returns `int \| None`) and integer literals in config stub / test fixtures | 17 |
| `696b74b` | `LicenseResponseDto` is `RootModel[UserLicense]` in v3 — construct with an explicit `root=UserLicense(...)` | 16 |

Notes on the sweep:

- Every str→UUID site was checked for the raw-Gumnut-ID hazard (a non-UUID string would raise `ValidationError` at runtime once unwrapped); all 71 pass properly decoded ids.
- The sync stream keeps a separate string form of the owner id for logging `extra` dicts — Sentry serializes unknown objects via `repr`, which would turn user ids into `UUID('...')` and break search/correlation.
- `SyncAlbumV1.thumbnailAssetId` and `SyncAssetFaceV1.personId` keep their string conversions: v3 left those fields typed as plain strings.

## Testing

- `uv run pyright`: **0 errors** (was 116)
- `uv run pytest`: **1132 passed** (adds system-config/jobs/license construction smoke tests and the WebSocket room-key regression test)
- `uv run ruff check` / `ruff format`: clean
- Previously-crashing endpoints (`GET /api/system-config`, `GET /api/jobs`, license get/set) invoked directly and verified working


routers/api/users.py
```diff
@@ -125,9 +125,11 @@ async def get_user_license() -> LicenseResponseDto:
     This is a stub implementation that returns fake license data.
     """
     return LicenseResponseDto(
-        licenseKey=STUB_LICENSE_KEY,
-        activationKey=str(uuid4()),
-        activatedAt=datetime(1900, 1, 1, 0, 0, 0, tzinfo=timezone.utc),
+        root=UserLicense(
```
taggartgorman, line 128: **[Consider]** License `RootModel` rewrite has no construction smoke test

Commit `696b74b` rewrites the four license stubs (`get_user_license`/`set_user_license` here, `get_server_license`/`set_server_license` in `server.py`) to `LicenseResponseDto(root=UserLicense(...))`. That's the same "construct a generated DTO by hand" failure surface the PR explicitly covers for system-config (`test_system_config.py`) and jobs (`test_jobs.py`) — but the license endpoints got no test.

Behaviorally this is a no-op today (the old kwargs form and the `root=` form emit byte-identical JSON, and `STUB_LICENSE_KEY` satisfies the `licenseKey` regex), so regression risk is low — hence Consider, not Should Fix. The value is symmetry: a future regen that tightens `UserLicense` or reshapes the `RootModel` would silently 500 `GET /api/users/me/license` and `/api/server/license` with no test failing.

Suggest a trivial smoke test mirroring `test_jobs.py`:

```python
assert isinstance(await get_user_license(), LicenseResponseDto)
```

(and one for `get_server_license`), so all three construction-fix groups in this PR are covered symmetrically.

routers/api/albums.py
```diff
@@ -189,23 +189,23 @@ async def add_assets_to_album(
         if asset_uuid in errors_by_uuid:
             results.append(
                 BulkIdResponseDto(
-                    id=asset_uuid_str,
+                    id=asset_uuid,
```
taggartgorman, line 191: **[Consider]** `asset_uuid_str` is now a cold-path-only local (defined at line 187)

After this sweep replaced every `id=asset_uuid_str` in `add_assets_to_album` with `id=asset_uuid`, the `asset_uuid_str = str(asset_uuid)` local at line 187 has exactly one remaining consumer — the `asset_id=asset_uuid_str` field of the rare contract-drift warning at line 217 — yet it's still computed on every loop iteration. The same PR fully removed the identical local from the sibling `remove_asset_from_album`, so this is a small leftover parallel-structure asymmetry.

Suggest dropping the local and inlining it in the warning's `extra`:

```python
extra={"album_id": str(id), "asset_id": str(asset_uuid)}
```

Negligible either way — raise only for consistency with the sibling function's cleanup.

taggartgorman (commented):
## Review: auto-approvable (0 Must Fix / 0 Should Fix)

Reviewed the full diff across correctness, data-integrity, security, silent-failures, performance, test-coverage, clarity, and minimalism. This clears 116 pyright errors and fixes three real runtime regressions (system-config/jobs 500s, silently-dropped WebSocket events, broken license endpoints), with byte-identical wire output on the mechanical str→UUID / float→int sweeps.

Verified independently: `pyright` → 0 errors, `ruff check` clean, new tests pass. The WebSocket room-key normalization was traced end-to-end — `str(user_id)` deterministically reproduces the room `connect()` registers from `session.user_id`, so there's no cross-user delivery risk; it strictly restores delivery to the correct room. `STUB_LICENSE_KEY` and `LicenseKeyDto` both satisfy the `UserLicense.licenseKey` regex, so the license endpoints don't 500.

**Value check: passes** — removes 116 static errors + 3 live bugs, minimal scope, no new abstraction, simpler call sites.

No blocking findings. Two optional **Consider** items inline below. (Posted as `COMMENT` rather than `APPROVE` only because GitHub doesn't allow approving your own PR.)

services/websockets.py
```diff
@@ -229,27 +230,33 @@ async def emit_user_event(
...
     Args:
         event: The event type (from WebSocketEvent enum)
-        user_id: The Gumnut user ID
+        user_id: The user's id — the UUID form of the Gumnut user id, or its string form
         payload: Event data - Pydantic model (auto-serialized), string, list, or None
 
     Raises:
         pydantic.ValidationError: If payload is a Pydantic model that fails serialization
     """
+    room = str(user_id)
```
taggartgorman, line 251: **[Consider]** WebSocket delivery depends on a room-key invariant that no test asserts

The normalization added here (`room = str(user_id)` in `emit_user_event`) restores realtime delivery only because the emit-time room key is byte-identical to the key the room was registered under at connect time. That invariant holds today, but it's implicit and untested:

- **Connect** registers the room with `session.user_id` (`connect` → `enter_room(sid, user_id)`), set at session creation as `str(safe_uuid_from_user_id(result.user.id))` — the canonical UUID string.
- **Emit** stringifies `current_user.id`, itself `safe_uuid_from_user_id(user.id)`.

Both sides derive from `safe_uuid_from_user_id` on the same user id, so they stringify to the same canonical form and the rooms match. Fine right now.

The gap: the new regression test (`test_normalizes_uuid_user_id_to_string_room`) only pins the `UUID → str` conversion in isolation — it does not assert the round trip (a client joins room X at connect, an `emit_user_event` for that user reaches room X). So if either derivation later drifts — e.g. session storage switches to the raw prefixed Gumnut user id, or `current_user.id` changes shape again — realtime events would silently drop into an empty room with **no failing test**, which is exactly the class of bug `dd5fe35` just fixed.

Not blocking, and out of scope for clearing the pyright errors — raising it so it's tracked before the v3 cutover. Suggest an integration-style test around cutover that registers a room via the connect path (or a stubbed session) and asserts an `emit_user_event` for that user is delivered to that room, rather than checking the string conversion alone.

routers/api/users.py
taggartgorman: Fixed — added `tests/unit/api/test_license.py` with `isinstance` construction smoke tests for `get_user_license` and `get_server_license`, mirroring `test_jobs.py`. All three construction-fix groups (system-config, jobs, license) are now covered symmetrically.

routers/api/albums.py
taggartgorman: Fixed — dropped the `asset_uuid_str` local and inlined `str(asset_uuid)` into the contract-drift warning `extra`, matching the sibling `remove_asset_from_album` cleanup.

services/websockets.py
taggartgorman: Deferring to the v3 cutover — a test that catches the drift here (session storage switching to the raw prefixed id, or `current_user.id` reshaping) needs to exercise the real connect→session→emit chain where `safe_uuid_from_user_id` lives, i.e. a genuine integration test, out of scope for this pyright-clearing PR. Leaving this thread open to track it.

PR #270, opened 2026-07-13
Add album map-markers endpoint (Immich v3)

taggartgorman:
## What
- Add `GET /api/albums/{id}/map-markers` — map markers for a single album's GPS-tagged assets (new in Immich v3).
- Extract the shared geotagged-marker paging/caps/logging from the global map endpoint into `routers/utils/map_markers.collect_geotagged_markers`, parameterized by an optional `album_id` (AND-combined with the world bbox); `get_map_markers` becomes a thin wrapper.
- A missing album returns 404 (pre-`retrieve`, matching Immich); `key`/`slug` shared-link params are accepted and dropped.

## Why
Immich v3 introduced this album-scoped endpoint and dropped the `albumIds` filter from the global `/map/markers`, so the v3 web album view calls it on every album open. The adapter returned 404, surfacing a console error and a user-visible "Something went wrong" toast on a high-traffic screen (the album still rendered). The album's markers reuse the existing geotagged-bbox retrieval over the Gumnut API, so a faithful implementation (real pins) costs only marginally more than an empty stub.

## Base
Targets `migration/immichv3`, not `main` — this is part of the Immich v3 retarget; the integration branch merges to `main` when the retarget lands.

## Test plan
- [x] `uv run pytest` — full suite green; adds album-scoped tests (markers built, coordinate-less skipped, `album_id` forwarded, empty album, missing album → 404 before listing, route registered)
- [x] Existing `/map/markers` tests pass unchanged — the refactor is behavior-preserving
- [x] `uv run ruff check` / `ruff format --check` / `uv run pyright` all clean

### DIFF
--- a/routers/api/albums.py
+++ b/routers/api/albums.py
@@ -21,6 +21,7 @@ from routers.immich_models import (
     AlbumsAddAssetsDto,
     AlbumsAddAssetsResponseDto,
     AlbumStatisticsResponseDto,
+    MapMarkerResponseDto,
     UpdateAlbumDto,
     UpdateAlbumUserDto,
     AddUsersDto,
@@ -32,6 +33,7 @@ from routers.utils.gumnut_id_conversion import (
     uuid_to_gumnut_asset_id,
 )
 from routers.utils.album_conversion import convert_gumnut_album_to_immich
+from routers.utils.map_markers import collect_geotagged_markers
 from pydantic.json_schema import SkipJsonSchema
 
 logger = logging.getLogger(__name__)
@@ -123,6 +125,36 @@ async def get_album_info(
         current_user,
         asset_count=gumnut_album.asset_count,
     )
+
+
+@router.get("/{id}/map-markers")
+async def get_album_map_markers(
+    id: UUID,
+    key: Annotated[str | SkipJsonSchema[None], Query(alias="key")] = None,
+    slug: Annotated[str | SkipJsonSchema[None], Query(alias="slug")] = None,
+    client: AsyncGumnut = Depends(get_authenticated_gumnut_client),
+) -> List[MapMarkerResponseDto]:
+    """Return map markers for the GPS-tagged assets in a single album.
+
+    New in Immich v3 (the global `/map/markers` dropped its `albumIds` filter),
+    the v3 web album view calls this on every album open. Markers are the
+    album's geotagged assets, collected by the shared `collect_geotagged_markers`
+    helper with the album filter AND-combined with the world bbox.
+
+    `key` / `slug` are accepted for client compatibility (shared-link tokens)
+    but dropped — this adapter doesn't support shared links. A missing album
+    404s: `albums.retrieve` raises `NotFoundError`, mapped to 404 by the global
+    handler (matching Immich, and consistent with `get_album_info`).
+    """
+    _ = key, slug  # accepted for client compat, dropped
+
+    gumnut_album_id = uuid_to_gumnut_album_id(id)
+
+    # Strict 404 for a missing album — retrieve first so a nonexistent album
+    # returns 404 rather than an empty marker list.
+    await client.albums.retrieve(gumnut_album_id)
+
+    return await collect_geotagged_markers(client, album_id=gumnut_album_id)
 
 
 @router.post("", status_code=201)
--- a/routers/api/albums.py
+++ b/routers/api/albums.py
@@ -150,8 +150,7 @@ async def get_album_map_markers(
 
     gumnut_album_id = uuid_to_gumnut_album_id(id)
 
-    # Strict 404 for a missing album — retrieve first so a nonexistent album
-    # returns 404 rather than an empty marker list.
+    # Retrieve first so a missing album 404s instead of returning an empty list.
     await client.albums.retrieve(gumnut_album_id)
 
     return await collect_geotagged_markers(client, album_id=gumnut_album_id)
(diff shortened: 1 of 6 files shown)


routers/api/albums.py
```diff
@@ -125,6 +127,35 @@ async def get_album_info(
...
+    the v3 web album view calls this on every album open. Markers are the
+    album's geotagged assets, collected by the shared `collect_geotagged_markers`
+    helper with the album filter AND-combined with the world bbox.
+
+    `key` / `slug` are accepted for client compatibility (shared-link tokens)
+    but dropped — this adapter doesn't support shared links. A missing album
+    404s: `albums.retrieve` raises `NotFoundError`, mapped to 404 by the global
+    handler (matching Immich, and consistent with `get_album_info`).
+    """
+    _ = key, slug  # accepted for client compat, dropped
```
taggartgorman, line 149: **[Consider]** The `key`/`slug` drop here is untested, unlike the sibling global endpoint which explicitly verifies its dropped params don't leak into the SDK call (`test_partner_and_shared_album_filters_are_dropped` in `test_map.py`). Low-risk today (the params are only `_ = key, slug`), but a matching test would guard against a future edit accidentally wiring them into the `retrieve`/`list` calls.

Suggested test in `TestGetAlbumMapMarkers`: call with `key`/`slug` set and assert `mock_client.assets.list.call_args.kwargs` stays `{"album_id", "limit", "include", "bbox"}` (neither `key` nor `slug` forwarded).

routers/api/map.py
```diff
@@ -1,16 +1,14 @@
...
 from pydantic.json_schema import SkipJsonSchema
 
-from routers.api.constants import GUMNUT_API_MAX_PAGE_SIZE
 from routers.immich_models import MapMarkerResponseDto, MapReverseGeocodeResponseDto
-from routers.utils.asset_conversion import ASSET_INCLUDE_METADATA_ONLY
 from routers.utils.gumnut_client import get_authenticated_gumnut_client
-from routers.utils.gumnut_id_conversion import safe_uuid_from_asset_id
+from routers.utils.map_markers import collect_geotagged_markers
 
 logger = logging.getLogger(__name__)
```
taggartgorman: **[Consider]** This module logger is now orphaned — its only `logger.info(...)` call moved into `collect_geotagged_markers`, so both this line and the `import logging` on line 1 are dead (ruff/pyright won't flag an unused module-level binding). Safe to remove both.

Caveat: `users.py` also carries an unused module logger, so if the intent is an "always declare a module logger" convention, keeping it is defensible — your call.

taggartgorman (commented):
## Automated review — clean ✅

Reviewed across correctness, data integrity, security, silent failures, performance, test coverage, clarity, and minimalism. **0 Must Fix, 0 Should Fix, 3 Consider.** No blocking issues.

The refactor is behavior-preserving: the extracted `collect_geotagged_markers` keeps the paging loop, both caps, and cap-hit logging identical to the prior inline `get_map_markers` — the only delta (dropping `str(...)` around `safe_uuid_from_asset_id`) is a no-op since `MapMarkerResponseDto.id: UUID`. The 404-before-listing check mirrors `get_album_info`, and `key`/`slug` are dropped without ever being consulted for auth (no shared-link bypass). Value check passes — both call sites get genuinely simpler and the shared marker logic can no longer drift.

I'm the PR author, so GitHub blocks a formal `APPROVE`; posting as `COMMENT`. All three findings are non-blocking.

### Pre-approve checklist
- ✅ **Findings**: 0 Must Fix / 0 Should Fix (3 Consider — inline + below)
- ✅ **Codebase value**: extraction removes ~94 inline lines from the global endpoint and lets the album endpoint reuse the shared marker logic instead of copy-pasting; beats do-nothing (user-visible error toast), an `[]` stub (permanently empty map), and copy-paste (drift risk)
- ✅ **Security/auth**: no auth/authz logic touched — `key`/`slug` are bound then dropped (`_ = key, slug`), never used for auth; fails closed via the existing authenticated-client dependency
- ✅ **Data deletion**: read-only endpoint; no destructive operations
- ✅ **Confidence**: refactor verified byte-identical to the prior loop; covered by existing `test_map.py` plus new `TestGetAlbumMapMarkers`
- ✅ **Purpose clarity**: purpose and approach clear from description and code

### [Consider] `docs/design-docs/immich-adapter-gap-analysis.md:5` — bump `last-updated`
_(Anchored in the body rather than inline: the frontmatter line isn't part of this PR's diff hunk.)_ This PR edits the doc body (flipping the album map-markers entry to "implemented"), but `last-updated:` still reads `2026-07-07`. `docs/references/code-practices.md:130` codifies bumping it when a gap entry is closed — set it to `2026-07-13`.

routers/api/albums.py
taggartgorman: Fixed — added `test_key_and_slug_are_dropped_not_forwarded` to `TestGetAlbumMapMarkers`. It calls the endpoint with `key`/`slug` set and asserts neither `albums.retrieve` (kwargs and positional args) nor `assets.list` (kwargs stay `{album_id, limit, include, bbox}`) receives them — covering the `retrieve` path the sibling global test does not.

routers/api/map.py
taggartgorman: Fixed — removed the orphaned `logger = logging.getLogger(__name__)` and its `import logging` from `map.py` (the only `logger.info` moved into `collect_geotagged_markers`). Left `users.py`'s pre-existing unused logger alone: it lives on the base branch, outside this PR's scope.

PR #271, opened 2026-07-14
Emit owner AlbumUserV1 link so v3 client displays albums

taggartgorman:
## Problem

Immich v3's `SyncAlbumV2` dropped `ownerId`, so the mobile client no longer derives an album's owner from the album event itself. Its album-list query **inner-joins on an owner-role album-user row**, so albums that synced into the client DB were filtered out and never displayed.

## Fix

Emit the owner album-user link on the separate `AlbumUsersV1` stream, derived from the same `album` events:

- **`AlbumUsersV1` is a first-class entry in `_SYNC_TYPE_ORDER`**, mapped to the same `album` gumnut entity as `AlbumsV1/V2`. It streams **after** the album (its FK parent) and after the owner `UserV1`. Each album fans out to two sync entities — the album (`AlbumV1/V2`) and its owner link (`AlbumUserV1`) — both derived from the same `AlbumResponse`.
- **Single-user model**: Gumnut has no album sharing, so every album has exactly one album-user — the owner (`role=owner`).
- **Upsert-only** (`_DERIVED_UPSERT_ONLY_TYPES`): album deletes are owned by the album pass (which emits `AlbumDeleteV1`), and the client's `remoteAlbumUserEntity.albumId` FK cascades on album deletion. Re-emitting the delete from the album-user pass would duplicate `AlbumDeleteV1`, so this pass skips deletes (the cursor still advances). This assumes an `Albums{V1,V2}` pass is co-requested to emit the delete — the v3 client always requests `AlbumsV2` alongside `AlbumUsersV1`. No `AlbumUserDeleteV1` is emitted — Gumnut has no unshare operation.

## Tests

- `test_streams_owner_album_user_for_albums_v2` — the v3 client (`AlbumsV2` + `AlbumUsersV1`) gets an owner `AlbumUserV1` link, ordered after the album.
- `test_album_users_v1_skips_album_delete_events` — the album-user pass does not re-emit album deletes (exactly one `AlbumDeleteV1`).
- `test_album_user_converter_maps_album_and_owner_distinctly` — direct converter test asserting `albumId` and `userId` map to distinct sources (a field swap fails; the stream test uses a single UUID for both, so can't catch one).
- `test_sync_v2.py` structural checks — sync-type ordering, upsert-only membership, converter role, and event routing.

Full suite: **1147 passed**; `ruff format`, `ruff check`, and `pyright` all clean.

## Docs

Documents the new album-user link in the sync-stream-architecture doc and the v3 API-change design doc, and records the single-`TEST_UUID`-masks-field-swaps testing gotcha in `code-practices.md`.


taggartgorman (commented):
Reviewed the owner-`AlbumUserV1`-link fix. The change is correct, minimal, and well-tested — traced the risk paths with concrete inputs and found no correctness, data-integrity, security, or silent-failure issues:

- **Cursor advances on skipped deletes** — `emit_deletes=False` skips only the *emission*; `last_cursor` still advances, so no events are lost or re-scanned within a pass.
- **`AlbumDeleteV1` dedup holds** — emitted exactly once (by the AlbumsV2 pass); the AlbumUsersV1 pass skips before `make_delete_sync_event`.
- **FK ordering safe** — owner `UserV1` precedes `AlbumUserV1`, and when the user is unchanged its client-side row persists.
- **Checkpoints isolated** — AlbumV2 and AlbumUserV1 iterate the same `album` events but use separate checkpoint keys / ack prefixes; no corruption.

One non-blocking doc-consistency nit below. (The `AlbumUsersV1` pass does a second scan of `album` events, but at single-user scale that's negligible and the single-pass alternative isn't worth the complexity — the trade-off looks right.)

**[Consider]** `docs/design-docs/immich-v3-api-changes.md:274` — the §7 retarget-plan step 3 still reads `(AlbumV2 == V1; AssetOcrV1 emits nothing)`, which this PR's own §5 update ("`SyncAlbumV2` is `SyncAlbumV1` **minus `ownerId`**") and the entire premise of this change contradict. Since you're already editing this doc, consider pointing that parenthetical at §5 / the owner-link requirement so the plan summary doesn't tell a reader "sync v2 is trivial, AlbumV2 == V1" when it actually required the separate `AlbumUsersV1` stream.

_(Couldn't anchor this inline — line 274 isn't part of the diff.)_

PR #274, opened 2026-07-14
Refine adapter handling of v3 mobile sync request types

taggartgorman:
## Summary

Refines how the adapter handles the sync request types the Immich v3 mobile client sends on every `/sync/stream` call. Two independent improvements, each verified against the v3.0.0 mobile client's actual read paths:

### 1. Quiet the false "unsupported types" warning
The v3 client requests a broad set of types with no equivalent in a single-user Gumnut library: `partnersV1`, `partnerAssetExifsV1`, `partnerStacksV1`, `partnerAssetsV2`, `stacksV1`, `memoriesV1`, `memoryToAssetsV1`, `assetMetadataV1`, `albumAssetExifsV1`. These were logged as `Client requested unsupported sync types` on **every** sync. They now live in `_NOOP_REQUEST_TYPES`, so the adapter records them as intentional no-ops (matching the existing `assetOcrV1` / `albumAssetsV2` handling) and the misleading warning stops. Logging-only change — nothing was ever streamed for them.

### 2. Show people with 1–2 faces in the mobile People tab
The v3 client reads `people.minimumFaces` from the `UserMetadataV1` sync stream to decide which people appear in its People tab, defaulting to **3** when the stream is absent — hiding clusters of 1–2 faces. The adapter now synthesizes a single `UserMetadataV1` *preferences* row with `minimumFaces=1`. Only `people.minimumFaces` is load-bearing; the other sections mirror the client's own `Preferences.fromMap` defaults (which the client applies to any absent section anyway). It is emitted right after `UserV1` (its `userId` FK parent) and keyed off the same user cursor as `UserV1` (the user's `updated_at`), so it acks once and re-syncs only when the user record changes. Only the `preferences` key is synthesized — onboarding UI is gated on the login response, not the sync stream.

## Testing
- Unit tests for both behaviors in `test_sync_stream.py` and `test_sync_v2.py`, including: `UserV1`→`UserMetadataV1` FK ordering when both are requested, re-emit when the user record changes, the emit-once skip when the checkpoint cursor matches, and a no-op request emitting nothing without tripping the "unsupported types" log.
- Full suite: **1154 passed**; `ruff check` / `ruff format` / `pyright` all clean.

## Notes
- Targets `migration/immichv3`, alongside the already-merged owner `AlbumUserV1` fix (#271).

### DIFF
--- a/routers/api/sync/converters.py
+++ b/routers/api/sync/converters.py
@@ -146,10 +146,14 @@ def gumnut_user_to_sync_user_metadata_v1(owner_id: UUID) -> SyncUserMetadataV1:
     with at least one face is shown; all other fields mirror the client's own
     defaults.
 
-    ``value`` is the server's nested ``UserPreferences`` JSON shape — the client
-    parses it via ``Preferences.fromMap`` (reads ``value["people"]["minimumFaces"]``
-    etc.), storing the map verbatim. Only the keys the client reads are included;
-    it defensively defaults everything else.
+    ``value`` is the server's nested ``UserPreferences`` JSON shape, which the
+    client parses via ``Preferences.fromMap`` (reads
+    ``value["people"]["minimumFaces"]`` etc.) and stores verbatim.
+    ``Preferences.fromMap`` reads every section null-safely and falls back to its
+    own default for anything absent, so ``people.minimumFaces`` is the only
+    load-bearing field here. The other sections each restate the client's own
+    default and are included solely to mirror the full canonical shape a real
+    Immich server emits — none of them are required by ``Preferences.fromMap``.
     """
     return SyncUserMetadataV1(
         key=UserMetadataKey.preferences,
--- a/routers/api/sync/converters.py
+++ b/routers/api/sync/converters.py
@@ -28,7 +28,9 @@ from routers.immich_models import (
     SyncAssetV2,
     SyncAuthUserV1,
     SyncPersonV1,
+    SyncUserMetadataV1,
     SyncUserV1,
+    UserMetadataKey,
 )
 from routers.utils.asset_conversion import (
     duration_ms,
@@ -131,6 +133,36 @@ def gumnut_user_to_sync_user_v1(user: UserResponse) -> SyncUserV1:
         profileChangedAt=user.updated_at,
         avatarColor=None,
         deletedAt=None,
+    )
+
+
+def gumnut_user_to_sync_user_metadata_v1(owner_id: UUID) -> SyncUserMetadataV1:
+    """Synthesize a UserMetadata *preferences* row.
+
+    Gumnut has no per-user preferences, but the v3 mobile client reads the
+    ``people.minimumFaces`` threshold from this stream to decide which people
+    appear in the People tab — defaulting to 3 when absent, which would hide
+    Gumnut clusters of 1–2 faces. We emit ``minimumFaces=1`` so every person
+    with at least one face is shown; all other fields mirror the client's own
+    defaults.
+
+    ``value`` is the server's nested ``UserPreferences`` JSON shape — the client
+    parses it via ``Preferences.fromMap`` (reads ``value["people"]["minimumFaces"]``
+    etc.), storing the map verbatim. Only the keys the client reads are included;
+    it defensively defaults everything else.
+    """
+    return SyncUserMetadataV1(
+        key=UserMetadataKey.preferences,
+        userId=owner_id,
+        value={
+            "folders": {"enabled": False},
+            "memories": {"enabled": True},
+            "people": {"enabled": True, "minimumFaces": 1},
+            "ratings": {"enabled": False},
+            "sharedLinks": {"enabled": True},
+            "tags": {"enabled": False},
+            "purchase": {"showSupportBadge": True},
+        },
     )
 
 
@@ -329,7 +361,13 @@ def gumnut_album_to_sync_album_user_v1(
 
 
 def gumnut_face_to_sync_face_v1(face: FaceResponse) -> SyncAssetFaceV1:
-    """Convert Gumnut FaceResponse to Immich SyncAssetFaceV1 format."""
+    """Convert Gumnut FaceResponse to Immich SyncAssetFaceV1 format.
+
+    ``imageWidth``/``imageHeight`` are emitted as 0: the v3 mobile client stores
+    them but has no read path (face/person thumbnails are served from a server
+    URL, and there is no local bounding-box overlay). Emit real dimensions only
+    when a client consumer actually reads them.
+    """
     bounding_box = face.bounding_box
 
     person_id = None
--- a/routers/api/sync/stream.py
+++ b/routers/api/sync/stream.py
@@ -147,7 +147,7 @@ _DELETE_TYPE_ORDER: list[SyncEntityType] = [
 #   - StacksV1:            Gumnut has no stacks; assets carry stackId=None and
 #                          render individually, so absent stack rows hide nothing.
 # UserMetadataV1 is deliberately NOT here — it is emitted (a synthesized
-# preferences row); see _stream_user_metadata.
+# preferences row); see gumnut_user_to_sync_user_metadata_v1.
 _NOOP_REQUEST_TYPES: dict[SyncRequestType, SyncEntityType] = {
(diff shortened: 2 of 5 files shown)


routers/api/sync/stream.py
```diff
@@ -125,24 +126,60 @@
...
+#                          AssetsV2 + AlbumToAssetsV1 (this stream is for assets
+#                          shared into an album by other users, which Gumnut lacks).
+#   - AlbumAssetExifsV1:   EXIF for other users' album assets; own-album-photo
+#                          EXIF arrives via AssetExifsV1.
+#   - MemoriesV1 /
+#     MemoryToAssetsV1:    Gumnut has no memories feature; the tab renders empty.
+#   - StacksV1:            Gumnut has no stacks; assets carry stackId=None and
+#                          render individually, so absent stack rows hide nothing.
+# UserMetadataV1 is deliberately NOT here — it is emitted (a synthesized
+# preferences row); see _stream_user_metadata.
```
taggartgorman: **[Should Fix]** These comments point to `_stream_user_metadata`, but no such function exists — grep finds only these two comment lines. The emission is inline in `generate_sync_stream` and the converter is `gumnut_user_to_sync_user_metadata_v1` (which the line-609 comment already names correctly). Same fix needed at line 175.

```suggestion
# preferences row); see gumnut_user_to_sync_user_metadata_v1.
```

routers/api/sync/stream.py
```diff
@@ -125,24 +126,60 @@
...
+# Cursor for the synthesized UserMetadataV1 preferences row. The payload is
+# static, so a constant cursor makes the client ack it once and skip it on every
+# subsequent sync. Bump the suffix if the emitted preferences change, to force a
+# one-time re-emit to already-synced clients.
+_USER_METADATA_CURSOR = "preferences-v1"
+
+# Supported SyncRequestTypes (used to detect unsupported types requested by
+# client). AuthUsersV1/UsersV1/UserMetadataV1 are handled specially (not via
+# _SYNC_TYPE_ORDER): the first two stream the user record, and UserMetadataV1
+# streams a synthesized preferences row (see _stream_user_metadata).
```
taggartgorman: **[Should Fix]** Same dangling `_stream_user_metadata` reference as line 150 — the symbol doesn't exist.

```suggestion
# streams a synthesized preferences row (see gumnut_user_to_sync_user_metadata_v1).
```

tests/unit/api/sync/test_sync_stream.py
```diff
@@ -570,6 +570,53 @@ async def test_streams_faces_when_requested(self):
...
+            updated_at=updated_at,
+            cursor="preferences-v1",
+        )
+        checkpoint_map = {SyncEntityType.UserMetadataV1: checkpoint}
+
+        events = await collect_stream(
+            generate_sync_stream(mock_client, request, checkpoint_map, mock_user)
+        )
+
+        assert [e["type"] for e in events] == ["SyncCompleteV1"]
```
taggartgorman, line 619: **[Consider]** The emit-after-`UserV1` ordering (so the `userId` FK parent exists on the client first) is called out in the code comment at `stream.py:608` and the new architecture doc, but no test exercises both types together — both new tests request only `UserMetadataV1`, so a reorder regression would go uncaught. The repo already tests the analogous `AlbumV2`→`AlbumUserV1` ordering in `test_streams_owner_album_user_for_albums_v2`. Consider adding a test requesting `[UsersV1, UserMetadataV1]` with an empty `checkpoint_map`, asserting emitted types are `['UserV1', 'UserMetadataV1', 'SyncCompleteV1']`. Low priority — the ordering is two adjacent `if`-blocks.

routers/api/sync/stream.py
```diff
@@ -566,6 +603,21 @@ async def generate_sync_stream(
                 )
                 logger.debug("Streamed user", extra={"user_id": owner_id})
 
+        # Stream a synthesized user-preferences row if requested. Static payload
+        # keyed off a constant cursor, so the client acks it once and skips it
+        # thereafter. Emitted after UserV1 so its userId FK parent exists on the
+        # client. See gumnut_user_to_sync_user_metadata_v1 for the why.
+        if SyncRequestType.UserMetadataV1 in requested_types:
```
taggartgorman, line 605: **[Consider]** This is the first place the adapter synthesizes `UserMetadataV1`. Real Immich emits *two* rows per user — `preferences` and `onboarding: {isOnboarded: true}` — and the adapter emits only `preferences`. If the v3 client gates onboarding UI on that row, a synced user could see the onboarding flow. Can you confirm whether a missing `onboarding` row triggers onboarding UI? If so, a second synthesized row (keyed to its own constant cursor) alongside this one would cover it; if not, no change needed.

routers/api/sync/converters.py
```diff
@@ -134,6 +136,36 @@ def gumnut_user_to_sync_user_v1(user: UserResponse) -> SyncUserV1:
...
+
+    ``value`` is the server's nested ``UserPreferences`` JSON shape — the client
+    parses it via ``Preferences.fromMap`` (reads ``value["people"]["minimumFaces"]``
+    etc.), storing the map verbatim. Only the keys the client reads are included;
+    it defensively defaults everything else.
+    """
+    return SyncUserMetadataV1(
+        key=UserMetadataKey.preferences,
+        userId=owner_id,
+        value={
```
taggartgorman, line 161: **[Consider]** Only `people.minimumFaces` is load-bearing here, but `value` carries seven sections. The docstring says "Only the keys the client reads are included" yet also says the others "mirror the client's own defaults" and that the client "defensively defaults everything else" — mildly contradictory, and it implies the six non-`people` sections change nothing. Either they're required by `Preferences.fromMap` (a complete-shape requirement worth stating explicitly) or they can trim toward `{"people": {"enabled": True, "minimumFaces": 1}}`. Not blocking — matching a full canonical server shape is a defensible choice.

taggartgorman (commented):
Reviewed both changes — the no-op reclassification and the synthesized `UserMetadataV1` preferences row. Correctness, data-integrity, security, and silent-failure dimensions came back clean: FK ordering holds because the v3 client always co-requests `UsersV1` with `UserMetadataV1`, the constant-cursor emit-once gate and ack round-trip are correct, and the no-op set masks nothing streamable. Good test + doc coverage.

Findings below are all non-blocking: one **Should Fix** (a dangling `_stream_user_metadata` symbol in two comments) plus three **Consider** items (an FK-ordering test gap and two payload-completeness verification requests).

routers/api/sync/stream.py
taggartgorman: Fixed — the comment now references `gumnut_user_to_sync_user_metadata_v1` (the symbol `_stream_user_metadata` never existed; the logic is inline in `generate_sync_stream`).

routers/api/sync/stream.py
taggartgorman: Fixed — same correction; this comment now references `gumnut_user_to_sync_user_metadata_v1`.

tests/unit/api/sync/test_sync_stream.py
taggartgorman: Added `test_streams_user_then_user_metadata_in_fk_order` — requests `[UsersV1, UserMetadataV1]` with an empty checkpoint map and asserts emitted types `["UserV1", "UserMetadataV1", "SyncCompleteV1"]`, so a reorder regression fails. (Also added a re-emit-on-cursor-change test while here.)

routers/api/sync/stream.py
taggartgorman: Skipping — verified against the v3 mobile client: the synced `UserMetadata` is consumed only for `preferences` (the People tab, via `userMetadataPreferencesProvider`); nothing reads the synced `onboarding` row, and onboarding UI is gated on the login response's `isOnboarded` (`LoginResponseDto`), a separate path. So a synthesized `onboarding` row would change nothing. Documented this in the architecture doc's User Preferences section.

routers/api/sync/converters.py
taggartgorman: Fixed the contradiction — rewrote the docstring: `Preferences.fromMap` reads every section null-safely and defaults anything absent, so only `people.minimumFaces` is load-bearing. The other six sections are kept to mirror the full canonical server shape (each already equals the client's own default) — the docstring now says so explicitly rather than claiming they're required.

PR #278, opened 2026-07-15
Retarget adapter to Immich v3.0.3

taggartgorman:
## What

- Bump the target Immich version to **v3.0.3** in both pinned sites (`.immich-container-tag` and the `Dockerfile` `ARG IMMICH_VERSION` default), and regenerate `routers/immich_models.py` against the v3.0.3 spec.
- Supply the fields the regen makes required on hand-built stub DTOs: `recentlyAdded` on every `UserPreferencesResponseDto` construction site, and `resolutions`/`videoCodecs` on the FFmpeg realtime config.
- Add construction smoke tests for the preference stubs, which had none, and pin the memories `for` param's dual-wire-form contract.
- Collapse the two byte-identical admin preference trees into one module-level constant, so the next required-field bump edits one site instead of two.
- Fix the `on_server_version` socket payload, which still carried Immich v2's semver.js object and made the web sidebar read `v3.0.3-rc.`.

## Why

Immich v3.0.1 added a required `recentlyAdded` user preference and v3.0.2 added required `resolutions`/`videoCodecs` to `SystemConfigFFmpegRealtimeDto`. Adapter stubs hand-construct both DTOs, so regenerating the models against v3.0.1+ without supplying the new fields fails construction outright. These were the blockers preventing any bump past v3.0.0.

Nothing was broken in production. The adapter serves the web client pinned by `.immich-container-tag`, so users were on the v3.0.0 client, which has no `recentlyAdded` code. This is what unblocks the bump.

The v3.0.3 web client reads `preferences.recentlyAdded.sidebarWeb` with **no optional chaining** (`UserSidebar.svelte`), so once we ship that client, omitting the field is a TypeError on every page that renders the sidebar — not a cosmetic gap. The stub emits `sidebarWeb=False`, hiding the link, which matches both upstream's own default and every other stubbed sidebar toggle.

Real-time HLS remains an intentional gap: `enabled=False` keeps both clients on direct playback. The resolution/codec lists are inert while disabled but are required by the schema, so they mirror the upstream server defaults.

### Notes for review

- **The version socket fix is a bug this branch inherited, not one it caused.** `services/websockets.py` emitted a parsed semver.js object whose `prerelease` is `[]`. Immich v2's client read only `major/minor/patch` so it was inert, but v3 formats the payload as ``v${major}.${minor}.${patch}${prerelease === null ? '' : `-rc.${prerelease}`}`` — and `[]` is not `null`, so the sidebar rendered a trailing `-rc.`. It arrived with the v3.0.0 GA retarget and reads `v3.0.0-rc.` on `main` today; fixed here because it is the same retarget's fallout. `GET /server/version` was already correct, so the two had silently drifted. Mobile does not subscribe to this event.
- **The `Dockerfile` bump is load-bearing**, not cosmetic: Render builds the Dockerfile without a way to inject a build-arg from a file, so its default is what ships to production. CI's `check-immich-version-sync` enforces the two pins match.
- The regen also retypes `SystemConfigIntegrityChecksumJob.percentageLimit` from `int` to `float` (`0..1`). The stub already passes upstream's default of `1`, which always meant "100% of assets", so it stays valid and unchanged in meaning — a near-miss rather than a break.
- The **admin** preference stubs were not in the original scope — pyright surfaced them. They had no test coverage at all, so the break would have stayed hidden until a client hit the route.
- `/memories?for=` was retyped `date-time` → `date` in v3.0.3. No adapter change needed (pydantic coerces `yyyy-MM-dd`, and only y/m/d are read), but the param stays typed `datetime` deliberately so both wire forms keep working; narrowing it to match the spec would 422 older clients. That contract is now pinned by a test — nothing guarded it before.
- `POST /search/metadata` forwards the clients' `takenAfter`/`takenBefore` untouched. Worth knowing that **web and mobile disagree** on what they send: web uses the `keepLocalTime` hack (fictitious offset), mobile converts to real UTC. Pre-existing and unchanged here; recorded at the call site because reconciling them would change user-visible search results.

## Test plan

- [x] Full suite: **1162 passed**; `ruff check`, `ruff format --check`, and `pyright` all clean
- [x] Smoke tests cover DTO construction for every preference stub, plus the `recentlyAdded` wire payload and `realtime.enabled=False`
- [x] The memories wire-form test is mutation-verified: narrowing the param to `date` fails the pre-v3.0.3 case with `date_from_datetime_inexact`, while the rest of the suite stays green
- [x] API compatibility validator against the v3.0.3 spec: `server` endpoint (what CI gates on) reports **no differences**; a full-endpoint run is byte-identical to `main`'s (37 errors / 149 warnings, all pre-existing unimplemented endpoints), confirming no new incompatibilities
- [x] Verified end-to-end against the real artifact: extracted the v3.0.3 web client and confirmed the shipped bundle contains the unguarded `recentlyAdded.sidebarWeb` read, which now resolves to `false` instead of throwing
- [x] Confirmed the served payload is `"recentlyAdded": {"sidebarWeb": false}` and realtime serializes as `{"enabled": false, "resolutions": [480, 720, 1080], "videoCodecs": ["h264", "hevc"]}`, matching upstream's defaults
- [x] Confirmed the admin hoist is payload-preserving: both endpoints return identical payloads, and `users.py`'s tree stays distinct
- [x] Ported the v3.0.3 client's `semverToName` and ran both payloads through it: the old one renders `v3.0.3-rc.` (reproducing the reported display), the new one renders `v3.0.3`. The emitted payload now matches upstream's own assertion, `{major, minor, patch, prerelease: null}`


docs/references/development-tools.md
```diff
@@ -95,10 +97,10 @@ The workflow checks the `server` endpoint by default, but this can be customized
 The `dump_openapi_json.py` tool prints the adapter's OpenAPI specification from the FastAPI app to stdout, without running a server:
 
 ```bash
-uv run tools/dump_openapi_json.py > /tmp/spec.json
+uv run tools/dump_openapi_json.py 2>/dev/null | sed -n '/^{/,$p' > /tmp/spec.json
```
taggartgorman: **[Should Fix]** `2>/dev/null` discards the tool's only error channel

The prose above this is accurate and I confirmed both halves of it — logs really do go to **stdout**, and the bare `>` really does produce `Extra data: line 1 column 5`. But that diagnosis motivates only the `sed`. The `2>/dev/null` is unexplained by the prose and works against it.

On a success run stderr is **0 bytes**, so the redirect discards nothing. What it discards is the failure case: `dump_openapi_json.py` prints its only diagnostics to stderr and exits 1 — `Error importing main app: {e}` / `Make sure you're running this from the project root directory` (lines 35-38), and `Error generating OpenAPI spec: {e}` (line 53).

Reproduced: run the documented recipe from the wrong cwd and you get a 0-byte `/tmp/spec.json` and no message at all — leaving you to debug a validator rejecting an empty file instead of reading the tool's own "run this from the project root" hint.

Dropping it lets the `sed` do the whole job the prose describes. Verified this still yields valid JSON (142 paths) while letting stderr through:

```suggestion
uv run tools/dump_openapi_json.py | sed -n '/^{/,$p' > /tmp/spec.json
```

(Fixing the tool to log to stderr would be tidier, but that means touching app-wide logging config at import of `main` — fairly scoped out of a version bump. Documenting the `sed` is the right call; it's just the `2>/dev/null` that should go.)

docs/references/development-tools.md
```diff
@@ -53,10 +53,12 @@ The generator's `datamodel-code-generator` dependency is unpinned (`>=0.25.0`, r
...
 
-### After Regenerating: Sweep Stub Literals via Pyright
+### After Regenerating: Sweep Stub Breakage via Pyright
 
 A regeneration that adds typing or pattern constraints (e.g. `str` → `UUID` ids, regex-patterned keys) silently turns hardcoded literals in stub endpoints into latent 500s — stubs have no test coverage, so the suite stays green while the endpoint fails response validation on every call. Don't hunt these by grep (partial sweeps have missed sites repeatedly); enumerate them from pyright's error list — `Literal['...'] cannot be assigned to parameter ... of type UUID` (or a pattern-constrained field) pinpoints every offending literal. Dynamic `str(...)`-of-UUID values coerce fine at runtime and are style cleanup, not defects; invalid *literals* are the class that 500s.
 
+A regen that makes a field **required** breaks the same stubs through a different error — `Argument missing for parameter "<name>"` at every hand-construction site. Sweep it the same way: pyright is the only pre-runtime signal for a stub that has no smoke test yet, and the smoke tests [code practices](./code-practices.md#bumping-the-immich-version) prescribes only cover stubs that already have one.
```
taggartgorman: **[Consider]** This overstates what pyright catches, and the closing clause is circular

Two things in the last sentence.

**The trailing clause reads as a tautology** — "the prescribed smoke tests only cover stubs that already have one." It also doesn't match the section it cites: code-practices § "Bumping the Immich Version" says *"Keep a construction smoke test per hand-built-DTO stub"*, i.e. a test for every hand-built stub, not only for ones that already have a test. (The anchor itself resolves fine.)

**More substantively:** pyright only catches the type-shaped half of this problem. A regen that adds a pydantic *constraint* to an existing field is invisible to it, because the annotation doesn't change. I tested this — adding `min_length=1` to `ReverseGeocodingStateResponseDto.lastUpdate` (a plain `str | None`, consumed by `system_metadata.py::get_reverse_geocoding_state`, which hardcodes `lastUpdate=""`) leaves pyright at **0 errors** and the suite at **1160 passed** — this PR's exact test-plan numbers — while the endpoint raises `ValidationError` -> 500 on every call.

This regen contains a live instance of that class: `percentageLimit` went `int (le=9007199254740991)` -> `float (ge=0.0, le=1.0)`. It stayed valid, and it happened to sit in `system_config.py`, which has a smoke test. As written, the paragraph will read to the next bumper as "pyright has you covered."

```suggestion
A regen that makes a field **required** breaks the same stubs through a different error — `Argument missing for parameter "<name>"` at every hand-construction site. Sweep it the same way: for a stub with no smoke test yet, pyright is the only pre-runtime signal. Note the limit of that signal — pyright catches a missing required argument or a changed annotation, but **not** a new `Field(pattern=…/min_length=…/ge=…/le=…)` constraint on an unchanged annotation, which only fails at value validation. For that shape, a construction smoke test is the only thing that catches it.
```

Separately and explicitly *not* asking for it here: eight router modules with hand-built DTO returns are imported by no test at all (`activities`, `api_keys`, `download`, `libraries`, `notifications`, `stacks`, `system_metadata`, `tags`). That's a reasonable follow-up, not scope for a version bump.

docs/references/code-practices.md
```diff
@@ -36,7 +36,7 @@ Style, patterns, and conventions for the immich-adapter codebase.
...
... a string produced by the web client's `asLocalTimeISO`, which does `setZone('utc', { keepLocalTime: true })`. The wire value's date and time components are the user's **local wall-clock**, with `Z` appended so it transports as a string — the offset is fictitious. Pull `.year/.month/.day/.hour/.minute` off the parsed datetime as-is; do **not** apply timezone math, or you'll shift the user's local "today" by their UTC offset. As of Immich v3.0.3 the client still applies the hack to the search filter's `takenAfter`/`takenBefore`, which `POST /search/metadata` forwards to the Gumnut API as `captured_after`/`captured_before` (`routers/api/search.py`) — whether that passthrough wants the fictitious offset or a real one depends on the Gumnut API's capture-time comparison, which is worth confirming before relying on it. `GET /memories?for=` was the canonical example until v3.0.3 retyped the param to a plain `date` — the client now sends `yyyy-MM-dd`, so its time components are always midnight, and `_local_today` accepts both forms deliberately. The hack may reappear on any endpoint where the client wants the server to interpret a value in the user's local time without exposing the offset.
```
taggartgorman: **[Consider]** Bullet parks an open question in a prescriptive doc and restates the `_local_today` docstring

Every factual claim here checks out — I verified each against source: v3.0.3's `SearchFilterModal.svelte` does still wrap `takenAfter`/`takenBefore` in `asLocalTimeISO`; `search.py:260-261` does forward them as `captured_after`/`captured_before`; the `for` param did retype to `date`; `_local_today` does accept both forms. Two hygiene points, not accuracy ones.

**The open question.** The bullet ends with "*worth confirming before relying on it*". A convention doc's job is to state the convention — a reader who arrives to learn the rule leaves with an unactionable maybe, no owner, and no way to tell whether anyone ever confirmed it. It's also slightly off against the code: `search.py` already forwards these in shipping code, so "before relying on it" describes a decision already made. This would do more work as a comment at the `captured_after`/`captured_before` forward in `routers/api/search.py`, and/or a tracked issue.

**The restatement.** The bullet went 111 -> 187 words and now teaches four things in one paragraph: the hack, its search call site, the memories history, and the open question. Two of those overlap the `_local_today` docstring (memories.py:85-92), which links *to* this doc and independently states that v3.0.3+ sends `yyyy-MM-dd`, that pydantic parses it to midnight, and that the param stays `datetime` to accept both — against § Project Conventions' "Never link to a doc *and* restate it". The docstring is the right home (symbol-scoped, and it holds the actual trap: "narrowing it would 422 the older one"), so the doc is the copy to trim. The docstring itself is proportionate — I'd leave it alone.

With the open question relocated and the memories sentence pointed at the docstring, the bullet lands back near its former length:

```suggestion
- **Immich web "today" wire format**: Some query params receive a string produced by the web client's `asLocalTimeISO`, which does `setZone('utc', { keepLocalTime: true })`. The wire value's date and time components are the user's **local wall-clock**, with `Z` appended so it transports as a string — the offset is fictitious. Pull `.year/.month/.day/.hour/.minute` off the parsed datetime as-is; do **not** apply timezone math, or you'll shift the user's local "today" by their UTC offset. As of Immich v3.0.3 the client still applies the hack to the search filter's `takenAfter`/`takenBefore`, which `POST /search/metadata` forwards to the Gumnut API as `captured_after`/`captured_before` (`routers/api/search.py`). `GET /memories?for=` was the canonical example until v3.0.3 retyped the param to a plain `date` — see `_local_today` in `routers/api/memories.py` for why the param stays typed `datetime`. The hack may reappear on any endpoint where the client wants the server to interpret a value in the user's local time without exposing the offset.
```

tests/unit/api/test_admin.py
```diff
@@ -0,0 +1,47 @@
...
+        """The stub preferences must validate against the generated models."""
+        result = await get_user_preferences_admin(uuid4())
+
+        assert isinstance(result, UserPreferencesResponseDto)
+
+    @pytest.mark.anyio
+    async def test_update_constructs_valid_dto(self):
+        """The update stub builds the same tree and must validate too."""
+        result = await update_user_preferences_admin(
+            uuid4(), UserPreferencesUpdateDto()
```
taggartgorman: **[Consider]** Empty update DTO means this can't distinguish "ignored" from "applied"

`update_user_preferences_admin` ignores `request` entirely and hardcodes a tree byte-identical to `get_user_preferences_admin`'s. Called with `UserPreferencesUpdateDto()`, this assertion is satisfied by exactly the construction path `test_get_constructs_valid_dto` already exercises — so it adds no coverage, and nothing anywhere pins the stub's one distinguishing behavior. If someone made this endpoint start applying the payload, no test would fail.

You already worked this out for the non-admin twin — `test_users.py::test_update_ignores_the_request` deliberately passes a non-empty payload, with the docstring: *"The update must be non-empty to distinguish 'ignored' from 'applied' — an all-`None` payload echoes unchanged preferences either way."* Same reasoning applies here.

Mirroring the sibling covers the construction path *and* the ignore contract in one test:

```python
result = await update_user_preferences_admin(
    uuid4(), UserPreferencesUpdateDto(recentlyAdded=RecentlyAddedUpdate(sidebarWeb=True))
)

assert isinstance(result, UserPreferencesResponseDto)
assert result.recentlyAdded.sidebarWeb is False
```

(Not offering this as an applyable suggestion since it also needs `RecentlyAddedUpdate` added to the `routers.immich_models` import at the top of the file.)

To be clear about what's working: reverting the admin `recentlyAdded` fix does fail all three tests in this file, so it genuinely closes the gap this PR is about. This is just about the update test earning its own place.

routers/api/admin.py
```diff
@@ -281,6 +283,7 @@ async def update_user_preferences_admin(
         people=PeopleResponse(enabled=False, sidebarWeb=False),
         purchase=PurchaseResponse(hideBuyButtonUntil="", showSupportBadge=False),
         ratings=RatingsResponse(enabled=False),
+        recentlyAdded=RecentlyAddedResponse(sidebarWeb=False),
```
taggartgorman: **[Consider]** This is the second byte-identical copy of the same 16-line tree

`get_user_preferences_admin` (243-263) and `update_user_preferences_admin` (267-289) build byte-identical `UserPreferencesResponseDto` trees — I diffed them, zero differences.

The duplication predates this PR, so pushing back on scope is fair. Raising it because this change adds `recentlyAdded` into *both* trees, which is exactly the tax the duplication imposes: paid twice in this file, three times across the PR. And the next required-field bump pays it again — which is notable given this PR exists *because* required-field bumps break hand-construction sites.

Hoisting the shared admin tree to a module-level constant (mirroring the existing `userPreferencesResponse` pattern in `users.py`) and returning it from both endpoints is a ~16-line delete that halves the per-bump edit.

Honest caveat that makes this smaller than it looks: the third tree in `users.py:57-72` is **not** identical (`memories=MemoriesResponse(duration=7, enabled=True)` vs `enabled=False` here), so a 3-way merge isn't the free win. Just the admin pair.

Entirely reasonable to defer this to a follow-up rather than widen a version bump.

taggartgorman (commented):
Reviewed with three parallel dimension agents (correctness/data-integrity/security/silent-failures; performance/test-coverage; clarity/minimalism). **0 Must Fix, 1 Should Fix, 4 Consider** — none in shipped adapter behavior. The version bump itself verified clean.

### What I verified rather than assumed

- **The regen is faithful.** Re-ran `tools/generate_immich_models.py` against the v3.0.3 spec and `ruff format`-ed it: byte-identical to the committed `routers/immich_models.py` modulo the timestamp header. Zero hand-edits. The `SystemConfigFFmpegDto`/`SystemConfigFFmpegRealtimeDto` reorder is codegen's own topological output, caused by the new `HlsVideoResolution` dependency.
- **The constraint sweep is complete.** A programmatic diff of the v3.0.0 and v3.0.3 specs across `components.schemas` *and* `paths` parameters shows the entire surface change is exactly: `recentlyAdded` (required), `realtime.resolutions`/`videoCodecs` (required), the `percentageLimit` retype, the `for` retype on `/memories` + `/memories/statistics`, and two optional params on `/search/metadata`. Every one is handled or provably inert. There are also no `model_validate`/dict-unpack construction sites of the affected DTOs, so pyright genuinely saw them all.
- **The `percentageLimit` claim holds exactly.** Upstream comments `percentageLimit: 1, // 100% of assets` at *both* tags — the old `maximum: MAX_SAFE_INTEGER` was a spec bug that v3.0.3 fixed. Bound probe: `1` valid, `100` rejected. Value valid, meaning unchanged.
- **`_local_today` works on both wire forms**, confirmed by execution (`2026-07-15` -> midnight; `...T23:30:00.000Z` -> same y/m/d). The justification checks out too: narrowing to `date` would 422 the old form (`date_from_datetime_inexact`) for any non-midnight wall-clock, i.e. essentially every real old-client request.
- **Defaults match upstream byte-for-byte.** `realtime` serializes to `{"enabled": false, "resolutions": [480, 720, 1080], "videoCodecs": ["h264", "hevc"]}`. `recentlyAdded.sidebarWeb=False` matches upstream's default and all four other stubbed sidebar toggles. `UserSidebar.svelte` does read it with no optional chaining — the PR's core justification is accurate.
- **The added tests earn their place**, checked by reverting each fix in a throwaway worktree: reverting the `admin.py` fix fails all three new admin tests; reverting `system_config.py` fails two pre-existing tests; reverting `users.py` errors at collection.

### The findings

All five are documentation and test hygiene. The one Should Fix is a documented shell recipe that swallows the tool's only error channel — worth fixing because it will cost a future reader a debugging session. The four Considers are optional; the `admin.py` duplication one is a fair follow-up rather than something to widen this PR with.

One finding is worth calling out beyond its severity: the new `development-tools.md` paragraph frames pyright as the sweep mechanism, but pyright cannot see a *constraint* added to an existing field. That's demonstrated inline, and this very regen contains an instance of the class (`percentageLimit`). It stayed valid by luck, and it happened to live in a module that has a smoke test. Worth stating the limit honestly while the paragraph is being written.

Nice work on the verification depth in the PR body — checking the shipped web bundle for the unguarded `recentlyAdded.sidebarWeb` read is exactly the evidence that makes this reviewable.

docs/references/development-tools.md
taggartgorman: Fixed — dropped `2>/dev/null`, kept the `sed`.

Confirmed the diagnosis end to end: running the documented recipe from the wrong cwd gives a 0-byte `/tmp/spec.json` and **no message at all**; without the redirect the tool's own error surfaces (in my repro, the settings ValidationError branch rather than the ImportError one — different path, same lesson). Verified the recipe still yields valid JSON (142 paths); stderr was never part of the pipe, so letting it through can't corrupt the file.

Agreed on scoping out the logging fix — that means touching app-wide config at `main` import, which doesn't belong in a version bump.

docs/references/development-tools.md
taggartgorman: Fixed — and this one went further than the suggestion, because applying it verbatim would have made the section self-contradictory.

Reproduced your demonstration independently: a `Field(min_length=1)` on an unchanged `str | None` leaves pyright at **0 errors** while runtime raises `string_too_short`. You're right, and the circular clause is fixed too.

The catch: your text says pyright does not see `Field(pattern=…)` — but the **pre-existing** paragraph directly above claimed pyright's error list "pinpoints every offending literal" for "a pattern-constrained field". Probed both shapes: pyright flags `Literal['not-a-uuid'] cannot be assigned to parameter of type UUID` and **misses** the pattern violation, while both fail at runtime. Patterns emit as `Annotated[str, Field(pattern=…)]` (24 in the artifact), and `STUB_LICENSE_KEY` is a live literal governed by one. So I narrowed that paragraph to *retypes* and dropped its false parenthetical — otherwise the two paragraphs assert opposite things.

One correction to my own first attempt, caught in self-review: I initially cited `percentageLimit` as an instance of the class. It isn't — it's a near-miss (untouched by this branch, valid only because upstream's default sits exactly on the new bound), and calling it an instance contradicted this branch's own commit message. Also fixed the claim that a widening retype fails at value validation: `int`→`float` with `v=100` passes: only the constraint can fail.

On the eight uncovered router modules — agreed, out of scope for a version bump.

docs/references/code-practices.md
taggartgorman: Fixed — applied the trimmed bullet, and moved the open question to the `search.py` call site.

Both points land. Confirmed the convention at `code-practices.md:22`: *"Home by scope: one symbol → its docstring... Never link to a doc **and** restate it"* — so I violated a rule documented in the file I was editing, and the same line also settles where the open question belongs (symbol-scoped, at the call site).

One correction to your wording before adopting it: `/memories?for=` retyped in **v3.0.3**, not v3.0.1+ — checked all four specs (`date-time` at v3.0.0/1/2, `date` at v3.0.3).

The relocated comment then turned out to need fixing itself, caught in self-review: I wrote that these values arrive in the keepLocalTime format, which is true of **web only**. Mobile serializes `value.toUtc().toIso8601String()` (`metadata_search_dto.dart:598`) from a local picker date — a genuine instant. So the two clients disagree, and stated flatly the comment would invite a "fix" that shifts mobile searches by the user's offset. Now scoped, with the divergence recorded and the open question sharpened: forwarding as-is can't satisfy both, and reconciling changes user-visible results — a decision, not a fix.

tests/unit/api/test_admin.py
taggartgorman: Fixed — mirrored the sibling's shape.

Verified the premise before acting: `update_user_preferences_admin` accepts `request` but never references it, and I diffed the two trees programmatically — byte-identical. So the empty-DTO call exercised exactly the path `test_get_constructs_valid_dto` already covered, and nothing pinned the ignore contract. Renamed to `test_update_ignores_the_request` to match the twin, added the `RecentlyAddedUpdate` import as you noted.

Confirmed it now has teeth rather than assuming: against an implementation that honors the request it fails; against the real stub it passes.

Worth noting this composed with the duplication thread — after the hoist both endpoints return the same constant, which makes the value assertion the only part carrying weight.

routers/api/admin.py
taggartgorman: Done rather than deferred — thanks for the out, but the argument for doing it was the stronger one: this PR exists *because* required-field bumps break hand-construction sites, so shipping it while adding a third site to maintain seemed like the wrong trade. It's a net −9 lines, behavior-preserving, and the tests added in this PR guard both endpoints.

Confirmed byte-identical via a programmatic diff of both trees before touching them, and confirmed the hoist is payload-preserving afterwards: both endpoints return identical payloads, all 12 required keys, `memories.enabled=False` intact — and `users.py`'s tree stays distinct (`enabled=True`), so no 3-way merge, exactly as you flagged.

One deviation: I mirrored the *structure* of `users.py`'s pattern but not its naming. `userPreferencesResponse` is a camelCase outlier; the repo's convention for module-level constants is UPPER_SNAKE (`STUB_LICENSE_KEY`, `GUMNUT_API_MAX_PAGE_SIZE`, `_YEAR_WINDOW`), so it's `_USER_PREFERENCES_RESPONSE` — module-private, since only admin.py uses it.

Side effect worth knowing: the hoist moved admin's failure mode from request-time 500 to import-time, which is strictly better — and promptly falsified `test_admin.py`'s docstring, which existed to explain the *inline* contrast. Caught in self-review and rewritten.

PR #281, opened 2026-07-16
fix(people): exclude hidden people when withHidden is omitted

taggartgorman:
## What

- `GET /people` and `GET /search/person` now exclude hidden people when `withHidden` is omitted, matching upstream Immich's `!withHidden` rule — an explicit `true` is the only opt-in.
- Both params move from `bool = Query(default=None)` to `Annotated[bool, Query()] = False`, correcting a declared type that claimed `bool` but defaulted to `None`. `closestAssetId` / `closestPersonId` get the same correction (both are inert — accepted but never read — so this is type-only).
- Documents a related divergence in `adapter-architecture.md`: the adapter's `total` is a post-filter count, whereas upstream's is unfiltered.

## Why

Both endpoints filtered only on `if withHidden is False`, so an omitted param fell through and returned hidden people — the opposite of upstream, where absent, `undefined`, and `false` all exclude. Any client relying on the default saw hidden people leak through:

- **Immich web**: the people-page search box calls `searchPerson({ name })` with no `withHidden`, so hiding a person and then searching their name surfaced them anyway.
- **Immich mobile**: calls `getAllPeople()` bare, so it was leaking too.

The `/people` grid itself was never affected — it passes `withHidden: true` deliberately and filters client-side.

### On the declaration form

`Annotated[bool, Query()] = False` is load-bearing, not style. Unit tests call handlers directly, bypassing FastAPI's parameter resolution, so a `Query(default=False)` param hands the test the raw `Query` object — which is **truthy**, making `not withHidden` skip the filter. The added omitted-case test fails against both the old condition *and* the `Query(default=False)` form; it passes only with the `Annotated` form, which yields a real `False`. This is the shape the repo's parameter convention already mandates.

### On `total`

Making the filter the default moves an existing divergence onto the default path: `total` is counted after filtering while `hidden` counts every hidden person, so `total < hidden` is reachable (upstream counts both unfiltered). Every web reader of `total` was checked and none break:

- **People page and people-manage page** — both load with `withHidden=true`, so no filter runs and their counts match upstream.
- **Explore page** — gates its People section on `total > 0` with `withHidden=false`. The counts genuinely differ here, in the adapter's favor: a fully-hidden people set hides the section, where upstream renders the header over an empty grid.
- **Face editor** — compares `candidates.length` against `total`, having built the list from the same filtered response.

Mobile reads only the `people` array. Left as-is and documented rather than changed, since altering `total` would also change existing explicit-`withHidden=false` behavior. A third-party client assuming upstream's `total >= hidden` invariant would be affected.

## Test plan

- [x] `uv run pytest` — 1165 passed
- [x] `uv run ruff check` / `uv run ruff format` / `uv run pyright` — clean
- [x] Verified the new omitted-case tests fail against the pre-fix condition (they catch the real bug, one per endpoint)
- [x] Verified they also fail against the `Query(default=False)` form (the truthy-`Query` trap is real)
- [x] Confirmed `hidden_count` is still computed before filtering, so `PeopleResponseDto.hidden` stays correct
- [ ] Manual: hide a person, search their name on the people page — they should no longer appear


taggartgorman (commented):
The fix is correct and I verified the core claim independently against the fork at v3.0.3: `person.repository.ts` uses literally `$if(!withHidden, …)` in both `getAllForUser` and `getByName`, so excluding hidden on an omitted param is upstream-faithful. Mobile's `person_api.repository.dart:15` calls `getAllPeople()` bare and is the one client this changes — toward parity. Mutation testing confirms both new omitted-case tests fail against the pre-fix code and against the `Query(default=False)` form, so they genuinely pin the default.

One note for future readers: the `Annotated[bool, Query()] = False` change is what actually fixes the bug. Reverting `not withHidden` back to `withHidden is False` while keeping the `Annotated` form leaves the suite green — once FastAPI coerces to a real `bool`, the two conditions are semantically identical. `not withHidden` is the clearer form and worth keeping, but the declaration is the load-bearing part, exactly as the PR body says.

Leaving `total` divergent-but-documented is the right call — I checked all four web readers and none break, and making `total` upstream-exact would make the explore page strictly worse (upstream renders an empty People section when everything is hidden; the adapter hides it).

All my findings are in the prose, and they share a shape worth naming: this PR's central new convention is *"verify claims about client behavior before writing them down — they're plausible-sounding and frequently wrong."* Both new prose claims fail that check. The convention is a good one; it just needs to survive its own first application.

docs/references/code-practices.md
```diff
@@ -121,6 +121,7 @@ The SDK is auto-generated (Stainless), so a version bump can add **newly-require
...
    - For loosely-specified response values (free-form strings, group/field names), also check how the Immich **clients** consume the response (`immich/web/src/` and `immich/mobile/lib/`) — clients hard-match values the OpenAPI spec doesn't constrain. E.g., the explore page renders only the group with `fieldName == "exifInfo.city"`, so a spec-valid response with different group names would silently render nothing.
+   - The same duty applies to any **claim about client behavior you write down** — in a comment, docstring, test name, or doc — not just to implementation decisions. "Web passes `X`" / "no client reads this field" are plausible-sounding and frequently wrong: a param is usually passed differently across a dozen call sites (`withHidden` is passed `true` in four, `false` in six, and omitted in one). Grep the fork for every caller before asserting one, and prefer stating what each caller does over a blanket claim. A false rationale outlives the change that introduced it.
```
taggartgorman: **[Should Fix]** The counts here don't hold. Grepping the fork at v3.0.3: `withHidden` is passed `true` at **six** `getAllPeople` call sites (eight including the two `searchPerson` sites), `false` at six, omitted at one. "Four" isn't right under any grouping — not by call site (6 or 8), not by distinct file (5). The `false` and omitted counts are both correct, which is what makes the wrong one easy to skim past.

Since this is the teaching example inside the bullet telling readers to grep the fork before asserting a client-behavior claim — and it closes with "a false rationale outlives the change that introduced it" — a wrong number undercuts the rule it's illustrating.

Suggest dropping the exact tally rather than correcting it: it rots on the next fork sync and nothing in CI pins it, and this PR is itself evidence the number is easy to get wrong. Something durable and still concrete would carry the point: *"a param is usually passed differently across many call sites — `withHidden` is passed both `true` and `false` across the web client's `getAllPeople`/`searchPerson` calls, and omitted at one in mobile."*

docs/architecture/adapter-architecture.md
```diff
@@ -149,6 +149,10 @@ Used when Immich clients expect offset-based pagination or need the full result
 4. Slice for the requested page: `all_items[(page-1)*size : page*size]`
 5. Return with `total`, `hasNextPage`, and other metadata
 
+**`GET /api/people` — `total` is a post-filter count, unlike upstream's.** Upstream sources both `total` and `hidden` from a single count that ignores `withHidden`, so it guarantees `total >= hidden`. The adapter instead counts `total` after the hidden filter runs, while `hidden` still counts every hidden person — so `total < hidden` is reachable, and `total - hidden` is only meaningful when no filter ran.
+
+Immich web reads `total` in three places, and none break: the people page derives its visible count as `total - hidden` but loads with `withHidden=true`, so no filter runs and the arithmetic matches upstream; the explore page gates its People section on `total > 0` with `withHidden=false`, where a fully-hidden people set hides the section instead of rendering an empty grid; and the face editor's candidate loop compares `total` against a list it built from the same filtered response. Mobile reads only the `people` array. A third-party client assuming upstream's `total >= hidden` invariant would be affected.
```
taggartgorman: **[Should Fix]** There are four readers, not three. Missing: `web/src/routes/(user)/people/manage/+page.svelte:24` — `const totalPeopleCount = $derived(data.people.total)`, rendered as the page-header count.

The conclusion survives: `people/manage/+page.ts:8` loads with `withHidden: true`, so no filter runs and `total` matches upstream — the same rationale already given for the people page. But this paragraph's persuasive force is the exhaustiveness of the list; "none break" only means something if all of them were found, and a reader who re-derives it will find four and not know which claim to trust.

At four items, each with its own `withHidden` value and rationale, this has outgrown one sentence. Suggest a bullet list: people page (`total - hidden`, `withHidden=true` → matches upstream); people-manage page (header count, `withHidden=true` → matches upstream); explore page (`total > 0`, `withHidden=false` → fully-hidden set hides the section rather than rendering an empty grid); face editor (`candidates.length === total` against the same filtered response). Keep the closing note about third-party clients.

docs/references/code-practices.md
```diff
@@ -81,7 +81,7 @@ For the full error handling strategy including rate limit protection and per-ite
 
 ### Defining Endpoint Parameters
 
-- Use `Annotated` to specify attributes, such as `Query()`, `Path()`, `Body()` functions, or numeric or string validations, but do not use `Default` — the default value should be specified as part of the Python declaration.
+- Use `Annotated` to specify attributes, such as `Query()`, `Path()`, `Body()` functions, or numeric or string validations, but do not use `Default` — the default value should be specified as part of the Python declaration. This is not just style: a `Query(default=X)` param hands unit tests the raw `Query` object rather than `X`, because they call handlers directly (see the Testing section), so a defaulted param's behavior can't be exercised through its own default.
```
taggartgorman: **[Consider]** This and the Testing bullet at line 432 now both explain that `Query(default=X)` hands directly-calling unit tests the raw `Query` object — and each cross-references the other while also restating it. Keep the pointer, cut the restatement.

The Testing bullet is the better owner: it covers the truthiness, the branch that silently flips, the legacy-shape caveat, and the `get_asset_statistics` case. But the "why isn't this just style?" hook here is doing real work at the point where someone decides whether to follow the convention, so a bare pointer would lose something. Middle ground: *"…the default value should be specified as part of the Python declaration. This is not just style — the `Query(default=X)` shape makes a param's default untestable; see the Testing section."*

docs/references/code-practices.md
taggartgorman: Fixed — dropped the tally rather than correcting it, for the reason you gave. Verified your count: `withHidden: true` at six `getAllPeople` sites, six `false`, two `searchPerson` true. Root cause was my original grep ending in `head -12`, which silently cut exactly the two `people/manage/*` lines that were `true` — the same truncation that hid the fourth `total` reader in the other thread.

One deviation from the suggested wording: it says "omitted at one in mobile", but web omits too (`PeopleSearch.svelte:86`), which is this PR's actual repro. The bullet now names no call site it would have to keep true, and I cut the closing aphorism and a `head`-cap trap that only bit if you were producing a count — which the rule now forbids anyway.

docs/architecture/adapter-architecture.md
taggartgorman: Fixed — you were right, and it was the same truncated grep as the tally thread: `head -12` cut `people/manage/*`, which is simultaneously why the count said four and why this list missed a reader. Added `people/manage/+page.svelte:24` (loader passes `withHidden: true`, so your rationale holds), and confirmed exhaustively there is no fifth reader.

Adopted the bullet list, with two changes. Merged the people and people-manage bullets since both reduce to "loads `withHidden=true` → matches upstream". And dropped the leading count entirely — self-review pointed out that `Immich web reads total in four places` is exactly the tally the rule added two files over forbids, and it had already rotted inside this PR, having said three a round ago.

I also nearly shipped a new false claim in the rewrite: I first wrote that the explore page hiding the section "is upstream's behavior", when it is the opposite — upstream renders the header over an empty grid. It now states the divergence, in our favor. Same fix applied to the PR description, which restated the same incomplete list.

docs/references/code-practices.md
taggartgorman: Fixed — agreed, and it violated this doc's own "never link to a doc *and* restate it" rule. Took your middle ground close to verbatim:

> …the default value should be specified as part of the Python declaration. This is not just style: the `Query(default=X)` shape makes a param's default untestable — see the Testing section.

Kept the Testing bullet as the owner of the mechanism.

PR #293, opened 2026-07-22
Map rating 0 to unrated in Immich EXIF conversion

taggartgorman:
## What
- `normalize_rating` now bounds a rating to the Immich DTO's valid 1-5 range, mapping 0, the deprecated -1, and any other out-of-range value to None.
- Added a parametrized unit test for the range plus end-to-end guards through `extract_exif_info` (REST) and `extract_sync_exif` (sync).
- Generalized the code-practices convention note so it names the "a Gumnut value violates a stricter Immich DTO constraint -> 500" class (rating range alongside the existing naive-datetime instance).

## Why
`ExifResponseDto.rating` is constrained `ge=1, le=5`, but cameras write a rating of 0 (XMP:Rating) — and the deprecated -1 — to mean "unrated". `normalize_rating` previously mapped only -1 to None and let 0 pass through, so a 0 rating raised a Pydantic `ValidationError` and returned a 500 from every endpoint that builds `exifInfo` (memories, timeline, single-asset, search). "Unrated" 0 is a common EXIF value (in practice the second-most-frequent rating), so the crash recurs whenever such a photo surfaces.

## Test plan
- [x] `uv run pytest tests/unit/utils/test_asset_conversion.py` — 73 passed
- [x] `uv run ruff check` and `ruff format --check` — clean
- [x] `uv run pyright` — clean
- New tests fail on the pre-fix code: the `0` and `6` unit cases return the raw value, and the end-to-end tests raise the `ValidationError` the fix prevents.


ternarybits (approved):
Thanks for looking into this.

PR #295, opened 2026-07-22
Support criterion-less POST /api/search/metadata (immich-go enumeration)

ternarybits:
## What
- Route a **criterion-less** `POST /api/search/metadata` to a plain asset listing instead of the Gumnut API's `search.search`. immich-go enumerates the whole library this way (its body carries only pagination + `visibility`/`order`/`withExif`, and for its trashed pass `trashedAfter` — no query/date/person).
- `_list_all_assets` follows the adapter's documented **Pattern 1** (load-all + client-side pagination, like `/api/assets/statistics`): maps `withDeleted→all` / `trashedAfter→trashed` / else `live` for the Gumnut `state`, filters `visibility` in memory, honors `order=asc`, slices `page`/`size` (clamped to 200), and returns a real `total` plus a string `nextPage` (`None` on the last page).
- A request carrying any real criterion keeps the existing `search.search` path unchanged. The classifier is an **allow-list** of enumeration-honorable fields, so a future OpenAPI-regenerated filter field fails safe onto the search path rather than being silently ignored.

## Why
- The Gumnut API's search mandates a criterion and returns 400 ("At least one search criterion must be provided") on an empty one, so immich-go's asset sweep aborted before uploading anything. Real Immich returns *all* assets for a filter-less metadata search; this makes the adapter match that contract.
- This is the adapter-side, production-ready replacement for a throwaway evaluation patch whose `nextPage=None` / `total=len(page)` only worked for a single-page library — real pagination here carries a client correctly through libraries larger than one page.

## Test plan
- [x] `uv run pytest` — 1209 passed (new `TestCriterionLessEnumeration` + `TestSearchMetadataEnumeration`: routing, multi-page `nextPage`/`total`, size clamp 1000→200, trashed/withDeleted state mapping, visibility filter, order reverse, real-criterion routing, and boundary cases — exact-full-final-page, page-past-end, empty library)
- [x] `uv run ruff format` / `uv run ruff check` / `uv run pyright` clean

### DIFF
--- a/routers/api/search.py
+++ b/routers/api/search.py
@@ -321,6 +321,22 @@ async def _list_all_assets(
     with no total, while Immich clients page by offset and read `total` /
     `nextPage`, so the full set is loaded once and sliced here.
     """
+    # Every Gumnut asset converts to `timeline` visibility (hardcoded in
+    # `convert_gumnut_asset_to_immich`), so a non-timeline query matches nothing.
+    # immich-go sweeps archive/timeline/hidden, so 4 of its 6 enumeration queries
+    # are non-timeline — short-circuit here, before the load, so those don't each
+    # fetch the whole library only to discard it.
+    if (
+        request.visibility is not None
+        and request.visibility != AssetVisibility.timeline
+    ):
+        return SearchResponseDto(
+            albums=SearchAlbumResponseDto(count=0, facets=[], items=[], total=0),
+            assets=SearchAssetResponseDto(
+                count=0, facets=[], items=[], nextPage=None, total=0
+            ),
+        )
+
     # `withDeleted` widens to live+trashed; immich-go's trashed pass instead
     # sends `trashedAfter` (a min date meaning "all trashed ever"). The Gumnut
     # API has no trash-date filter, so `trashedAfter`'s value isn't honored —
@@ -356,16 +372,6 @@ async def _list_all_assets(
             for asset in all_assets
             if asset.trashed_at is None or asset.trashed_at >= request.trashedAfter
         ]
-
-    # Every Gumnut asset converts to `timeline` visibility (hardcoded in
-    # `convert_gumnut_asset_to_immich`), so a non-timeline query — immich-go
-    # sweeps archive/timeline/hidden — matches nothing rather than returning the
-    # live/trashed set once per visibility.
-    if (
-        request.visibility is not None
-        and request.visibility != AssetVisibility.timeline
-    ):
-        all_assets = []
 
     # The Gumnut API lists in descending order (capture time for live/all, trash
     # time for trashed); honor an explicit ascending request by reversing it
--- a/routers/api/search.py
+++ b/routers/api/search.py
@@ -249,6 +249,31 @@ async def search_asset_statistics(
     return SearchStatisticsResponseDto(total=total)
 
 
+# Fields on MetadataSearchDto that a plain asset listing can honor (or that
+# don't restrict the result set), so their presence does NOT disqualify a
+# request from the criterion-less enumeration path: pagination, response-shape
+# hints, sort order, and the visibility/trash selectors `_list_all_assets`
+# translates. Every *other* field is a restricting filter. Framing this as an
+# allow-list (rather than enumerating the restricting fields) fails safe:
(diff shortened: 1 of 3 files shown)


routers/api/search.py
```diff
@@ -248,13 +249,157 @@ async def search_asset_statistics(
...
+    # sweeps archive/timeline/hidden) correctly matches nothing here rather than
+    # returning the live/trashed set once per visibility.
+    if request.visibility is not None:
+        immich_assets = [
+            asset for asset in immich_assets if asset.visibility == request.visibility
+        ]
+
+    # The Gumnut API lists in descending order (capture time for live/all,
+    # trash time for trashed); honor an explicit ascending request by reversing
+    # that order (immich-go sends order="asc"). Mirrors the timeline endpoint.
```
ternarybits: Create a ticket to support this natively in the list_assets endpoint and then use it in immich-adapter.

routers/api/search.py
```diff
@@ -248,13 +249,157 @@ async def search_asset_statistics(
...
+    # its presence just selects the trashed set, which is exactly immich-go's
+    # all-trashed intent.
+    if request.withDeleted:
+        state = "all"
+    elif request.trashedAfter is not None:
+        state = "trashed"
+    else:
+        state = "live"
+
+    all_assets = [
```
ternarybits, line 352: all_assets could be huge.

do we really need everything in ASSET_INCLUDE?

Could we incrementally convert to immich_assets and filter the list rather than try to pull all assets into memory at once?

routers/api/search.py
```diff
@@ -248,13 +249,157 @@ async def search_asset_statistics(
...
+    """
+    # `withDeleted` widens to live+trashed; immich-go's trashed pass instead
+    # sends `trashedAfter` (a min date meaning "all trashed ever"). The Gumnut
+    # API has no trash-date filter, so `trashedAfter`'s value isn't honored —
+    # its presence just selects the trashed set, which is exactly immich-go's
+    # all-trashed intent.
+    if request.withDeleted:
+        state = "all"
+    elif request.trashedAfter is not None:
+        state = "trashed"
```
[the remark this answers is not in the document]
ternarybits: Fixed — the listing now filters by `trashed_at >= trashedAfter` (live assets always kept), so a real bound is honored instead of returning the whole trash. Applied whenever `trashedAfter` is set, so it is correct for `state=trashed` and the `withDeleted` `state=all` case too. immich-go sends the min sentinel, for which this keeps everything. Covered by two new tests.

routers/api/search.py
ternarybits: Filed an internal tracking issue to add native ascending-order support (plus offset/total) to the Gumnut API asset listing, and switch the adapter to it — which drops this in-memory reverse. Leaving the reverse in place until that lands, since immich-go requests ascending order.

routers/api/search.py
ternarybits: Reduced the in-memory footprint: `_list_all_assets` now converts only the requested page slice to Immich DTOs (heavy conversion bounded to `size`), instead of converting the whole library. The listing itself is still fully loaded because offset paging over the cursor-based asset listing needs the full ordered set for `total`/slicing — that full-load (and the include set) is what the native offset/total tracking issue above covers.

routers/api/search.py
```diff
@@ -248,13 +249,177 @@ async def search_asset_statistics(
...
+            asset
+            for asset in all_assets
+            if asset.trashed_at is None or asset.trashed_at >= request.trashedAfter
+        ]
+
+    # Every Gumnut asset converts to `timeline` visibility (hardcoded in
+    # `convert_gumnut_asset_to_immich`), so a non-timeline query — immich-go
+    # sweeps archive/timeline/hidden — matches nothing rather than returning the
+    # live/trashed set once per visibility.
+    if (
```
ternarybits: Shouldn't the request visibility check optimally go before fetching all_assets?

ternarybits: Good call — moved it. The non-timeline short-circuit now returns the empty result before loading the library, so immich-go's archive/hidden passes (4 of its 6 enumeration queries) no longer fetch the whole listing just to discard it. Added an assertion that `assets.list` isn't called in that case.

PR #318, opened 2026-07-29
Add shared burst-stack translation primitives

taggartgorman:
## What

- **SDK bump** `gumnut-sdk` 0.137.0 → 0.140.0 — the first published release whose stack resource is complete (0.138 and 0.139 still lack `create_stack` and `add_assets_to_stack`).
- **Stack ID conversion** — `safe_uuid_from_stack_id` / `uuid_to_gumnut_stack_id` over the existing generic shortuuid implementation, on the compound `asset_stack` prefix.
- **New `routers/utils/stack_conversion.py`** — member hydration, effective-cover resolution, bounded-concurrency hydration for a collection, and a synchronous `StackResponseDto` builder.
- **Test factories** — `make_gumnut_stack` / `make_gumnut_stack_members`, a shared `MockPaginatedListing`, plus a consolidation of the duplicated asset fixtures onto one builder.
- **Corrections to the fan-out contract** — see *Beyond the stated scope* below.

No endpoint behavior changes; `routers/api/stacks.py` is still stubbed.

## Why

Immich stack support lands across several surfaces — the `/stacks` reads and writes, the timeline's per-asset stack tuples, an asset's own `stack` block. Each needs the same two things before it can answer: the stack's members, and a decision about which member Immich shows as the cover. Building those once here keeps the surfaces from drifting into showing a different frame for the same burst, and keeps the follow-on endpoint PRs small.

The cover decision is the part that genuinely needs a single home. Immich requires a non-null `primaryAssetId`, but a Gumnut auto-detected burst deliberately has none — the backend pins one only when a user picks it, and leaves the default to the client. So the adapter has to choose, and the rule is now written down once: the pinned cover when present (kept even when trashed, since the backend preserves a trashed cover's ID until permanent deletion and dropping it would silently override the user), otherwise the first live member, otherwise the first trashed one.

## The DTO shape, and why it matches upstream

Two guarantees the OpenAPI spec cannot express, both taken from the Immich source at the tag this repo pins (`.immich-container-tag`, `v3.0.3`):

- **`assets` carries live members only.** `server/src/repositories/stack.repository.ts` filters `asset.deletedAt is null`. `StackResponseDto` has no count field, so clients read `assets.length` as the stack's size — `web/src/lib/utils/actions.ts` writes it into the timeline's per-asset stack block, and `keepThisDeleteOthers` feeds the array back into a bulk delete. Emitting trashed frames would inflate the badge against a timeline that counts live assets, and re-delete already-trashed ones.
- **`assets[0]` is the cover.** `server/src/dtos/stack.dto.ts::mapStack` partitions `[...primary, ...others]`. The asset viewer navigates to `stack.assets[0]` after removing an asset from a stack, and the filmstrip renders in server order.

Together these mean a **trashed pin names an asset absent from `assets`** — which is upstream's behavior too, since it preserves `primaryAssetId` from the stack row while its member query drops the asset. Callers must not assume the primary can be found in `assets`. The member read still uses `state="all"` so a trashed pin stays resolvable; the filtering is a response-shape rule applied at DTO build.

The read passes `order="asc"` explicitly rather than inheriting the server default. The default is newest-first, which made "otherwise the first live member" resolve to the *last* frame of a burst — the opposite of both the docstring and Immich's own convention (`POST /stacks` takes `assetIds[0]` as the primary). Pinning it also stops a future change to the server default from silently moving the cover of every auto-detected burst.

A member-less row yields `None` rather than a fabricated UUID. It can't form a valid `StackResponseDto`, and inventing an ID would only move the failure to the client.

## Beyond the stated scope

Self-review caught a docstring claiming `asyncio.gather` cancels pending siblings on the first exception. **It does not** — the exception propagates, but the siblings run to completion, so an aborted fan-out still costs its full upstream cost. That claim pre-existed in four other places, all corrected here.

Two further gaps in the same contract closed during review:

- **The claim had no test.** `test_propagates_exception` passes identically under a cancelling primitive, so nothing in the suite distinguished the two. `test_siblings_run_to_completion_after_a_failure` observes the siblings' side effects instead, and is the only test that fails when the helper is swapped for a cancelling gather.
- **Propagation also discards.** Every sibling exception after the first vanishes — no propagation, no loop exception handler, no warning. `delete_people` is destructive and 204-on-all-success, so a batch where two deletes fail surfaces one error and records nothing for the other. Now documented, along with the fact that "first" means completion-order-first, not input-order-first.

Chasing this through `test_concurrency.py` also turned up a test that didn't test what it named: `test_cancellation_does_not_warn_unawaited_coroutines` triggered on a sibling exception — which cancels nothing — so the `coro.close()` guard never ran and the test passed with the guard deleted. It now cancels the gather itself.

Happy to split these out if you'd rather keep the PR to stacks alone — they're here because the new module's documented contract depends on that helper being described accurately.

## Test plan

- [x] `uv run ruff format` / `uv run ruff check` clean
- [x] `uv run pytest` — 1280 passing (1238 on `main`)
- [x] Cover-resolution branches covered: pinned, pinned-and-trashed, pinned-but-absent, unpinned-live, all-trashed, member-less
- [x] DTO shape pinned: live-only `assets`, cover-first ordering, trashed pin absent from `assets`, all-trashed stack yields empty `assets`
- [x] Mutation-checked rather than assumed: dropping the trashed filter, dropping the cover-first sort, dropping `order="asc"`, widening the pinned-cover warning, swapping the bounded fan-out for a bare `gather`, swapping it for a *cancelling* gather, adding a non-conforming class to the Protocol list, and deleting the `coro.close()` guard each fail their intended test
- [x] SDK contract pinned both ways — method parameters and response-row fields — with the row check enforced statically by pyright
- [x] Concurrency bound verified by peak-counter (peak ≤ 10 with the helper, 30 without)

## Follow-on

A workflow-documentation learning from this build lands separately in the shared agent-tooling repo; it has no effect on this adapter's behavior and nothing here depends on it.


taggartgorman (commented):
Reviewed across correctness/data-integrity/security/silent-failures, performance/test-coverage, and clarity/minimalism. **0 Must Fix, 5 Should Fix, 5 Consider.** Nothing here ships behavior — the routes are still stubbed — so none of it is urgent, but the two DTO-contract items are worth settling before the endpoint PRs build on this module.

What I checked and found solid, so it doesn't clutter the comments below:

- **The `gather` correction is right.** Verified against CPython semantics and by running the helper directly: the exception propagates, siblings are not cancelled, and queued coroutines still start because of `_run`'s `finally` release. All five corrected sites are accurate.
- **Cover resolution is sound.** Every branch traced with concrete inputs — it can never return a non-member, never returns `None` for a non-empty stack, and keeping a trashed pin matches upstream Immich, which reassigns the primary only at permanent delete.
- **The prefix claim holds.** `asset_` and `asset_stack_` genuinely cannot cross-decode: shortuuid's default alphabet is 57 characters with no underscore, so both directions raise.
- **No security findings.** The only external input is a backend-authored row; the client-facing direction doesn't exist yet.
- The committed diff is clean of private-repo paths and tracker IDs.

The two contract findings (#1, #2 inline) come from reading the Immich source at the exact tag in `.immich-container-tag` (`v3.0.3`) rather than from the OpenAPI spec — the spec can't express either guarantee.

---

### Two findings that have no line in this diff to anchor to

**[Should Fix] `docs/design-docs/immich-adapter-gap-analysis.md:151` — this PR contradicts an active design doc.**

That doc is `status: active` and is listed under Active Design Docs in `AGENTS.md`, with the stated purpose "Prioritizing adapter work, evaluating stub endpoints, assessing feature gaps." Its § 6 Stacks says three things the SDK version this PR pins disproves:

- L151 — "**Dependency**: **Both** — the Gumnut API would need a grouping/stacking concept"
- L153 — "**Effort**: **L** — Backend needs a parent-child asset relationship model, group CRUD, and primary asset designation logic"
- L155 — "Consider when burst/HDR detection is added to Gumnut"
- L753 — the Tier-3 table repeats it: `| #6 Stacks | L | Both | Low user demand |`

On 0.140.0 the stacks resource carries `create_stack`, `add_assets_to_stack`, `list_stacks`, `retrieve_stack`, `set_cover`, `remove_assets`, `delete`; `list_stacks` takes `origin: Literal['auto_burst', 'user']` (burst detection exists) and `primary_asset_id` (primary designation exists); and `assets.list` accepts the `stack_id` filter `fetch_stack_members` depends on. The backend half is done and the named blocker is gone, so the effort is now adapter-only.

This is the doc someone opens to decide whether to start stacks work, so it's worth flipping Dependency to adapter-only, restating Effort, replacing the recommendation with current status, and updating the Tier-3 row.

**[Should Fix] The PR description's "Follow-on" line points at a private sibling repo.**

`AGENTS.md` puts PR bodies in scope and names this exact shape as a violation that has shipped before: a cross-link whose number resolves to a 404 for anyone outside the org. This repo is public and that sibling repo is private, so an outside reader gets a 404 plus "a workflow-documentation learning," which carries no information. Either drop the line or describe the follow-on's content without the unreachable reference.

---

The five `[Consider]` items are non-blocking; #9 and #10 in particular are cleanup you may reasonably want to defer.

routers/utils/stack_conversion.py
```diff
@@ -0,0 +1,232 @@
...
+    hydrated: HydratedStack, current_user: UserResponseDto
+) -> StackResponseDto:
+    """Convert an already-hydrated stack into Immich's `StackResponseDto`."""
+    return StackResponseDto(
+        id=hydrated.id,
+        primaryAssetId=hydrated.primary_asset_id,
+        assets=[
+            convert_gumnut_asset_to_immich(member, current_user)
+            for member in hydrated.members
+        ],
```
taggartgorman: **[Should Fix]** `assets[0]` must be the primary asset

Immich guarantees that `StackResponseDto.assets[0]` *is* the primary. At the tag this repo pins (`v3.0.3`), `server/src/dtos/stack.dto.ts::mapStack` partitions and re-concatenates specifically to ensure it:

```ts
const primary = stack.assets.filter((asset) => asset.id === stack.primaryAssetId);
const others  = stack.assets.filter((asset) => asset.id !== stack.primaryAssetId);
return { id, primaryAssetId, assets: [...primary, ...others].map(...) };
```

Here the members go out in the Gumnut API's own ordering, so `assets[0]` is whichever frame sorts first — generally not the resolved cover whenever a pin exists, or whenever the fallback skipped a trashed member.

There's a concrete client dependency: `web/src/lib/components/asset-viewer/AssetViewer.svelte:325` runs `cursor.current = stack.assets[0]` after `REMOVE_ASSET_FROM_STACK`, i.e. it navigates to what it assumes is the cover. The filmstrip renders server order too, so the cover wouldn't appear first there either.

Since this is the shared builder every planned stack surface will call, the deviation propagates to all of them — which is the drift the module docstring says it exists to prevent.

```suggestion
        assets=[
            convert_gumnut_asset_to_immich(member, current_user)
            for member in sorted(
                hydrated.members,
                key=lambda m: safe_uuid_from_asset_id(m.id)
                != hydrated.primary_asset_id,
            )
        ],
```

(`sorted` is stable, so the cover moves to the front and everything else keeps its relative order — the same semantics as `mapStack`.) Worth a test asserting `response.assets[0].id == response.primaryAssetId` for a stack whose pin is *not* the API's first-returned member; `test_primary_is_always_one_of_the_returned_assets` currently checks membership, not position.

routers/utils/stack_conversion.py
```diff
@@ -0,0 +1,232 @@
...
+    return await gather_with_concurrency(
+        [hydrate_stack(client, stack) for stack in stacks]
+    )
+
+
+def build_stack_response(
+    hydrated: HydratedStack, current_user: UserResponseDto
+) -> StackResponseDto:
+    """Convert an already-hydrated stack into Immich's `StackResponseDto`."""
+    return StackResponseDto(
```
taggartgorman, line 278: **[Should Fix]** Trashed members ship in `assets`, but upstream Immich sends live-only

`state="all"` is well-justified for the *fetch* — the pin has to stay resolvable when trashed — but every member then flows into `assets`. Upstream doesn't do that: at `v3.0.3`, `server/src/repositories/stack.repository.ts` filters `.where('asset.deletedAt', 'is', null)`, so a real Immich `/stacks` response carries live members only.

`StackResponseDto` has no count field, so clients derive counts from `assets.length`. Verified consumers at that tag:

- `web/src/lib/utils/actions.ts:78` — `assetCount: stack.assets.length` is written into the timeline's per-asset stack block. Immich's own timeline count is live-only, so the stack badge would disagree with the timeline for the same burst.
- `web/src/lib/utils/asset-utils.ts:349` (`keepThisDeleteOthers`) — sends `stack.assets.filter(a => a.id !== keepAsset.id)` to `deleteAssets`, re-deleting already-trashed frames.
- `AssetViewer.svelte` filmstrip — renders trashed frames inline in a live stack.

Concrete case: a 5-frame burst where the user trashed 2 reports `assetCount: 5` on the tile while the timeline computes 3.

I can see this was deliberate — `test_builds_dto_with_converted_members` asserts it, with "Trashed members belong in the response; Immich renders them in trash" — so I'm raising it as a contract call rather than a bug. Two ways out: emit live members only (matching upstream, and accepting a trashed pin outside `assets`, which is exactly what upstream does), or keep them and document the deviation plus the inflated `assets.length` in the module docstring so the follow-on PRs pick counts deliberately.

Related: `test_primary_is_always_one_of_the_returned_assets`'s docstring says "Immich resolves the cover by looking it up inside `assets`" — upstream actually violates that invariant itself, preserving a trashed `primaryAssetId` while `withAssets` drops the asset. Worth correcting whichever way this lands.

tests/unit/utils/test_concurrency.py
```diff
@@ -75,7 +76,8 @@ async def work() -> None:
 
     @pytest.mark.anyio
     async def test_propagates_exception(self):
-        """First exception bubbles up; pending siblings are cancelled."""
+        """First exception bubbles up (siblings are not cancelled — see
```
taggartgorman, line 79: **[Should Fix]** This PR's headline contract correction isn't pinned by any test

The "gather does not cancel siblings" claim is now asserted in five places — this docstring, `gather_with_concurrency`'s, `delete_people`'s, `hydrate_stacks`', and twice in `code-practices.md`. No committed test checks it.

This is the natural home and the docstring now makes the claim verbatim, but the only assertion is `pytest.raises(RuntimeError)`, which passes identically under a cancelling primitive. The test that *did* exercise the sibling-after-exception path (`[boom()] + [never_runs() ...]`) was rewritten in this PR to cancel the gather instead, so that shape is gone from the suite. `test_upstream_failure_aborts_the_batch` explicitly disclaims proving it and points back at the docstring.

That's the failure mode this PR exists to fix — the *opposite* claim sat unchallenged in these same files until self-review caught it by hand. The docstring even names the alternative ("need a TaskGroup"); if anyone acts on that, all five sites silently become false and `delete_people`'s stated semantics flip with nothing failing.

I verified the shape below by running the helper directly — `completed` is `[0, 1]` at propagation and `[0, 1, 2]` after draining:

```python
@pytest.mark.anyio
async def test_siblings_run_to_completion_after_a_failure(self):
    """An aborted batch still costs its full fan-out."""
    completed: list[int] = []

    async def work(idx: int) -> int:
        await asyncio.sleep(0)
        completed.append(idx)
        return idx

    async def boom() -> int:
        raise RuntimeError("boom")

    with pytest.raises(RuntimeError, match="boom"):
        await gather_with_concurrency([work(0), boom(), work(1), work(2)], limit=2)

    assert sorted(completed) == [0, 1]  # index 2 was still queued
    await asyncio.sleep(0.01)           # queued sibling acquires and runs
    assert sorted(completed) == [0, 1, 2]
```

Index 2 is the load-bearing half — it starts only because of `_run`'s `finally: semaphore.release()`, the most surprising part of the documented claim. This fails under a TaskGroup swap, and once it exists `test_upstream_failure_aborts_the_batch` can point at it instead of at prose.

routers/utils/stack_conversion.py
```diff
@@ -0,0 +1,232 @@
...
+    members; the adapter does not re-sort. The only property relied on is that
+    the order is deterministic, so two reads of an unchanged stack resolve to
+    the same cover.
+
+    Returns `None` only when the stack has no members at all.
+    """
+    if not members:
+        return None
+
+    pinned_id = stack.primary_asset_id
```
taggartgorman, line 156: **[Consider]** A pin that's absent from the members silently overrides the user's chosen cover

When `stack.primary_asset_id` isn't among `members`, this falls through to the live/trashed fallbacks with no signal. `hydrate_stack` logs a warning for the member-less case and the comment there explains why `asset_count` makes that disagreement queryable — but this is the same class of row-vs-member-read disagreement, and it has a directly user-visible effect: the cover becomes a different frame than the one the user pinned.

It's reachable, not just theoretical. A trashed pin keeps its ID until permanent deletion, so a purge landing between the stack-row read and the member read leaves the already-fetched row carrying a stale pin the member read no longer returns; `stacks.remove_assets` racing a listing page does the same. The docstring names the race itself ("the asset left the stack between the two reads").

Two reviewers raised this independently. In `hydrate_stack`, after resolving:

```python
pinned = stack.primary_asset_id
if pinned is not None and primary.id != pinned:
    logger.warning(
        "Stack %s pinned cover %s is not among its members; falling back to %s",
        stack.id,
        pinned,
        primary.id,
        extra={
            "stack_id": stack.id,
            "pinned_asset_id": pinned,
            "effective_asset_id": primary.id,
        },
    )
```

plus a `caplog` assertion on the existing `test_pin_absent_from_members_falls_back`.

routers/utils/stack_conversion.py
```diff
@@ -0,0 +1,232 @@
...
+    omit the very asset the stack points at. `ASSET_INCLUDE` because each member
+    is converted by `convert_gumnut_asset_to_immich`, which reads `metadata`,
+    `people`, and the `file_data` scalars.
+
+    `limit` is the per-page size, not a cap — `async for` walks the SDK's cursor
+    pages, so a stack with more members than one page still returns in full.
+    """
+    return [
+        asset
+        async for asset in client.assets.list(
```
taggartgorman, line 114: **[Consider]** The unpinned cover inherits a server-side default, and "first live member" means *newest*

`order` isn't passed, so this takes the Gumnut API default. Per the SDK 0.140.0 `assets.list` docstring, `state="all"` returns members "ordered by capture time like `live`" with "Ordering defaults to newest ... first." So `resolve_effective_primary`'s rule 2 — "otherwise the first live member" — resolves to the **newest** frame of the burst, not the earliest. That isn't what the prose conveys, and it's the opposite of Immich's own stack-create convention (`assetIds[0]`, the first asset the user selected).

Determinism itself does hold — I checked this specifically, since burst frames share near-identical timestamps, and the SDK documents an asset-ID tie-breaker. But version-stability is also being relied on: if the server default ever flips, the cover for every auto-detected unpinned burst changes silently and no test catches it, because the tests supply their own fixed mock ordering.

Since this is the one rule every stack surface shares, it seems worth being explicit rather than inheriting it — pass `order=` deliberately (`"asc"` if the intended cover is the first frame of the burst, `"desc"` to pin today's behavior), and reword rule 2 to name which end of the burst it picks.

routers/utils/stack_conversion.py
```diff
@@ -0,0 +1,232 @@
...
+    propagates and the partial results are discarded. The siblings are *not*
+    cancelled — `asyncio.gather` lets them run to completion — so an aborted
+    batch still costs its full member fan-out. Only a route knows whether its
+    endpoint should fail or degrade, so a caller wanting to drop just the failed
+    stack (say, one deleted between the listing page and its hydration) should
+    catch per stack rather than change this for everyone. Note the asymmetry
+    with a member-less stack, which is *not* an upstream failure and yields
+    `None`.
+    """
+    return await gather_with_concurrency(
```
taggartgorman, line 247: **[Consider]** The bound caps parallelism, not total cost — and `/stacks` is unpaginated

This is an N+1 by construction: one `assets.list` walk per stack, each with `include=ASSET_INCLUDE`, which carries `people` — the server-side aggregation `ASSET_INCLUDE_NO_PEOPLE` exists specifically to avoid. `gather_with_concurrency` bounds in-flight reads at 10, but not the number of round-trips, the wall-clock, or peak memory, and `stacks` has no ceiling.

The input size isn't hypothetical. In the pinned spec, `searchStacks` takes exactly one optional query param, `primaryAssetId` — no `page`/`size`. So the follow-on route has to return *every* stack, each with its full `assets` array. A library with 2,000 bursts becomes 2,000+ upstream walks at 10 concurrent — roughly 20s at 100ms per round-trip, past client timeouts. That's the same shape as the incident already recorded in `code-practices.md` § Counts and Aggregates ("fans out into N paginated GETs of full asset payloads — this scaled to >10s on large persons").

There is a cheaper shape for that surface specifically: `AssetResponse.stack_id` is populated, so one `assets.list` walk over the library grouped in memory by `stack_id` costs `ceil(library / 200)` requests instead of one walk per stack, and wins whenever stack count exceeds `library_size / 200`.

Nothing is broken as merged — there's no caller. Raising it because these docstrings are the contract the route PRs get written against, and the module already steers timeline surfaces away from `hydrate_stack` while the unpaginated all-stacks read is the bigger cost and currently reads as solved. Worth saying explicitly what the bound doesn't cover, and choosing between the two shapes deliberately in the route PR.

routers/utils/stack_conversion.py
```diff
@@ -0,0 +1,232 @@
...
+) -> list[HydratedStack | None]:
+    """Hydrate a collection of stacks under the shared fan-out bound.
+
+    Each stack costs its own `assets.list` walk, so a full page of stacks would
+    open that many concurrent upstream reads if launched naively.
+    `gather_with_concurrency` bounds the in-flight calls and preserves input
+    order, so the result zips back to `stacks` positionally — including the
+    `None` entries for member-less stacks.
+
+    An upstream failure on any one stack aborts the whole batch: the exception
```
taggartgorman: **[Consider]** Five copies of the rationale corrected — still five copies

The PR rightly makes `gather_with_concurrency`'s docstring the canonical statement, but the same fact is then written out again here (two sentences, two lines above the call to that very helper), in `delete_people`, at `code-practices.md:347`, and again at `code-practices.md:357` — where "Just don't read that as the siblings being called off — per above, they still run" links *and* restates in one clause. Prose on this one fact went from roughly six lines to roughly thirty across five files.

The PR's own narrative is that the claim drifted precisely *because* it lived in five copies. Correcting all five without reducing the count preserves the surface that caused it. `code-practices.md` § Project Conventions states the rule directly: "explain a non-obvious *why* once; cite it elsewhere in one line rather than restating (copies multiply churn and drift)."

Two smaller instances of the same pattern in this PR:

- The `asset_`/`asset_stack_` cross-decode mechanism is spelled out in full three times — `adapter-architecture.md:169`, `safe_uuid_from_stack_id`'s docstring, and `test_stack_and_asset_ids_do_not_cross_decode`'s docstring. (The mechanism is correct; this is placement only.)
- "The `/stacks` routes are still stubbed" appears in the new model-mapping row and at `code-practices.md:117`, while `adapter-architecture.md`'s own **Stub implementations** table — the section that exists for exactly this fact — omits Stacks.

Suggested shape: keep the full statement in `concurrency.py` and one in `code-practices.md:347`; reduce this docstring and `delete_people` to pointers; drop the restatement clause at `:357`. Keep the cross-decode explanation in the docstring and shorten the other two to a line. Add a Stacks row to the Stub implementations table and drop the status tails elsewhere.

tests/unit/utils/test_stack_conversion.py
```diff
@@ -0,0 +1,519 @@
...
+        client = Mock()
+        client.assets.list = Mock(side_effect=_list)
+
+        result = await fetch_stack_members(client, stack.id)
+
+        assert len(result) == total
+        assert listings[0].pages_fetched == 2
+
+
+class _PaginatedListing:
```
taggartgorman: **[Consider]** This is now the *third* verbatim copy of `_PaginatedListing`

I diffed the bodies: this one, `tests/unit/api/test_memories.py:430`, and `tests/unit/api/test_map.py:69` are identical — same class name, same `__init__`/`__aiter__`/`_iter`. Only the docstrings differ.

`code-practices.md:438` (untouched by this PR) tells future authors "See `tests/unit/api/test_memories.py::_PaginatedListing` for the canonical shape" — so the doc names one canonical shape while three identical ones exist, and anyone who improves the canonical copy won't know about the others. The docstrings have already diverged on arrival: this copy drops the "limit is a result cap vs. limit is per-page" explanation that is the whole reason the helper exists.

Since this PR was already editing `tests/conftest.py` to add shared builders, and that file already hosts the sibling paginator mock `MockSyncCursorPage`, the shared home was open. Moving it there, importing it in all three modules, and repointing the `code-practices.md` bullet would close it out.

routers/utils/stack_conversion.py
taggartgorman: Fixed — `build_stack_response` now stably sorts the cover to `assets[0]`, reproducing `mapStack`'s partition. I sorted the *converted* DTOs rather than calling `safe_uuid_from_asset_id` per member as suggested, since `AssetResponseDto.id` is already a UUID. Added `test_primary_leads_the_assets_array`, which pins a member that is not the API's first-returned one — emitting in fetch order fails it. Verified by mutation.

routers/utils/stack_conversion.py
taggartgorman: Settled as live-only, matching upstream. `assets` now filters trashed members; `HydratedStack.members` still carries every state so a trashed pin stays resolvable. This accepts a trashed pin naming an asset absent from `assets` — which is upstream's own behaviour, since it preserves `primaryAssetId` from the stack row while `withAssets` drops the asset. You were right that `test_primary_is_always_one_of_the_returned_assets` asserted an invariant upstream doesn't hold; it's replaced by `test_trashed_pin_is_absent_from_assets`, and `test_builds_dto_with_live_members_only` covers the count. Three tests fail if the filter is removed.

tests/unit/utils/test_concurrency.py
taggartgorman: Fixed — added `test_siblings_run_to_completion_after_a_failure`. I ran your version first and it reproduced exactly (`[0, 1]` at propagation, `[0, 1, 2]` after draining). One change: the pre-drain assertion is now `assert 2 not in completed`, because the `[0, 1]` value depends on `work()` having exactly one `await` — two yields gives `[0]`, three gives `[]` — while the post-drain assertion carries the contract in every case. The drain is a deterministic hop count rather than wall-clock. It's the only test in the suite that fails when the helper is swapped for a cancelling gather, and `test_upstream_failure_aborts_the_batch` now points at it instead of at prose.

routers/utils/stack_conversion.py
taggartgorman: Fixed as documentation — `hydrate_stacks` now says the bound caps in-flight reads but not round-trips or peak memory, which scale with stack count times per-stack member count, and names the transpose (one `assets.list` walk grouped by `stack_id`) for the unpaginated whole-library read. Confirmed `StackSearchSchema` really does take only `primaryAssetId`. I did not add a member cap: with no caller it's complexity for a risk nothing demonstrates yet, and the route PR is where it can be chosen against real code.

routers/utils/stack_conversion.py
taggartgorman: Fixed, and this one was worth more than it looked. The gather claim is now one canonical statement in `gather_with_concurrency` plus pointers from `hydrate_stacks` and `delete_people`, with the restatement clause at `code-practices.md` dropped; cross-decode is one statement plus two pointers; a Stacks row went into the Stub implementations table and the status tails were dropped elsewhere. Worth flagging: self-review caught that my first pass *relocated* the duplication rather than removing it — three fresh copies of the trashed-member rule appeared in the module docstrings, and the follow-up restated the docstring it had just written. Both are cleaned up; I measured the count before and after rather than assuming.

tests/unit/utils/test_stack_conversion.py
taggartgorman: Fixed — hoisted to `tests/conftest.py` as `MockPaginatedListing` beside `MockSyncCursorPage`, all three local copies deleted, and the `code-practices.md` bullet repointed. Confirmed the bodies were byte-identical and that the new copy had dropped the result-cap-vs-per-page explanation. Two follow-ons this surfaced: `test_map.py` was paging at 50 while `collect_geotagged_markers` sends 200, so the hoisted docstring's rule was violated by a site it now governs — both map sites now use `GUMNUT_API_MAX_PAGE_SIZE` and assert `call_args.kwargs["limit"]`, which closes a drift that previously left all 86 map tests green. And the rule is now stated as an invariant rather than one mechanism, since the three call sites satisfy it differently.

routers/utils/stack_conversion.py
taggartgorman: Fixed — `hydrate_stack` now warns when the resolved cover differs from the pinned one, with `pinned_asset_id` and `effective_asset_id` in `extra`, plus the `caplog` assertion you suggested.

One refinement worth noting: the condition is "resolved primary != pinned id", not "pin is absent from `assets`". Those differ now that `assets` is live-only — a trashed pin is honoured but absent from the response, and warning on that would fire for every stack whose cover sits in the trash. `test_trashed_pin_is_not_treated_as_an_override` pins the distinction; widening the condition fails it.

routers/utils/stack_conversion.py
taggartgorman: Fixed — the read now passes `order="asc"` explicitly.

Confirmed your reading of the SDK docstring: the default is newest-first and there is an asset-ID tie-breaker, so determinism did hold, but rule 2 was resolving to the *newest* frame. I went with `asc` rather than pinning today's behaviour, so "the first live member" is true as written and matches Immich's `assetIds[0]` convention; `fetch_stack_members` now records why it is passed rather than inherited.

Two tests cover it: `test_pins_the_member_read_arguments` asserts it at the call site, and `test_asset_list_accepts_stack_id_filter` now includes `order` in the SDK signature check — the call-site assertion is a Mock, so it would survive the SDK dropping the parameter, and the failure would otherwise only surface when a stack route is first exercised.

ternarybits (approved):

PR #324, opened 2026-07-30
Implement the Immich stack read endpoints

taggartgorman:
## What

- `GET /api/stacks` and `GET /api/stacks/{id}` return real burst stacks with their live members, replacing a stub that returned `[]` and one that fabricated a DTO. Both go through the shared translation layer added earlier for member hydration and cover resolution, so they can't disagree with the later stack surfaces about which frame represents a burst.
- `primaryAssetId` is answered by resolving the asset's own stack and comparing *effective* covers, rather than forwarding the Gumnut API's `primary_asset_id` filter.
- The list walks the backend's stack cursor rather than answering with one page, bounded by both a 500-stack cap and a 5000-member hydration budget — whichever binds first — with walked/budgeted/hydrated/returned counts, a truncation flag, and which bound fired all logged.
- The five write routes stay stubbed; a follow-up issue owns them.

## Why

Immich's `searchStacks` takes no pagination parameters, so its unfiltered answer is *every* stack the user has — one backend page would be a wrong answer rather than a partial one, since the client has no way to ask for the rest. But unbounded it is also the one read here whose cost has no ceiling: every stack costs its own member walk, and bounding concurrency doesn't bound round-trips or peak memory. Hence the walk plus explicit bounds, matching how `/map/markers` already answers the same shape.

A cap on *rows* isn't a bound on *work*, though: 500 bursts of 3 frames and 500 of 10,000 both satisfy it at wildly different cost. A listing row carries its own `asset_count`, so each stack's size is known before anything is fetched — the walk spends a member budget from it as it goes and stops on whichever bound binds first. Admission stays all-or-nothing per stack, because clients read `assets.length` as the stack's size: a truncated member array is a wrong answer where a missing stack is merely an incomplete one. That puts the real bound at the budget plus one stack, and a single stack outweighing the entire budget is warned about rather than silently absorbed — dropping a user's largest stack would be a worse answer than a slow one.

Both bounds are tested before admitting a row rather than after appending it, so a library of exactly 500 stacks exhausts the cursor without reporting a truncation that never happened. That flag is the only signal a library has outgrown the endpoint, so a false positive on the boundary would cost it its meaning. Since two bounds can now stop the walk, the boolean is named for the truncation rather than for either bound, and `stack_truncated_by` records which one fired — the remedies differ. The budget is spent in live members while hydration reads `state="all"`, so the summary carries the realized total alongside the budgeted one; the gap is measurable instead of assumed.

Forwarding the backend's `primary_asset_id` filter was the tempting one-liner and the wrong one. That filter matches only covers a user explicitly pinned, while the adapter synthesizes a cover for every auto-detected burst — so forwarding it would answer "no stack" for exactly the bursts whose cover Immich is already displaying. Resolving the asset's stack and comparing against the cover the adapter actually reports costs two extra reads and is correct for both pinned and unpinned stacks. Those reads are an implementation detail, so a not-found from any of them yields an empty result: upstream's search is a plain equality filter that matches nothing for an unknown ID, and it would be wrong for the adapter's extra lookups to turn a *search* into a 404.

Both routes drop a stack with no live members through one shared rule. A member-less row can't name a cover at all; an all-trashed row hydrates fine but converts to `assets: []` with a `primaryAssetId` the client can't fetch — breaking both contracts clients rely on, since they read `assets.length` as the stack's size and `assets[0]` as the cover. Upstream never emits that DTO, because it deletes a stack once it falls below two assets, so nothing is written to survive one.

## Not in this PR: multi-library accounts

The Gumnut API makes `library_id` optional for an account with one library and required for an account with several, so an unqualified stack listing fails for the latter. That is real, but it is adapter-wide rather than a defect in this endpoint: every Gumnut listing call the adapter makes — timeline, search, trash, memories, map markers, the sync stream — goes out unqualified, and `/api/libraries` is a stub returning an empty list. A route that qualified its own calls in isolation would report full results while its neighbours still failed, which is harder to diagnose than a uniform failure. The constraint is now recorded in `docs/architecture/adapter-architecture.md` § Single-library assumption, where the fix is scoped as threading a library through shared request context rather than per route.

## Test plan

- `uv run pytest` — 1509 pass, 38 of them new in `tests/unit/api/test_stacks.py`: empty/one-page/multi-page lists, both bounds and their exact-boundary cases (a library landing exactly on the stack cap or exactly on the member budget must not report truncation), the oversized-stack warning and the silence of an ordinary library, the budgeted-vs-hydrated divergence on trashed members, full member hydration and UUID conversion, pinned vs. synthesized covers, the `primaryAssetId` filter across pinned/synthesized/non-primary/unstacked/unknown inputs, the three deleted-mid-request races, member-less and all-trashed drops on every path that decides them, detail success and 404s, and that both handlers declare their auth dependency.
- `uv run ruff check`, `uv run ruff format`, `uv run pyright` — clean.
- `python3 scripts/lint_docs.py` — clean.
- `tools/validate_api_compatibility.py --endpoints=stacks` against the Immich v3.0.3 spec — 0 errors; the 6 warnings are the pre-existing UUID `pattern` difference that every adapter path parameter carries, including the untouched stub routes.

## Known follow-up

`routers/utils/map_markers.py` sets its cap-hit flag *after* appending, so a library with exactly `MAP_MARKERS_CAP` markers reports a truncation that didn't happen (and `scan_cap_hit` is derived after the loop, with the same off-by-one). This PR adds the check-before-admit rule to `docs/references/code-practices.md` and applies it to stacks, but deliberately doesn't change map markers: the fix costs one lookahead item, which for that endpoint's evenly-dividing cap pulls in an extra upstream page. That tradeoff belongs to that endpoint rather than to this PR, so the doc scopes its citation instead.


routers/api/stacks.py
```diff
@@ -1,30 +1,172 @@
...
     """
-    Search stacks.
-    This is a stub implementation that returns an empty list.
-    """
-    return []
+    if primaryAssetId is not None:
+        return await _search_by_primary_asset(client, primaryAssetId, current_user)
+
+    stacks = []
+    async for stack in client.stacks.list_stacks(limit=GUMNUT_API_MAX_PAGE_SIZE):
```
taggartgorman, line 158: **[Must Fix] Enumerate stacks for multi-library users**

The Gumnut SDK requires `library_id` when an authenticated user owns multiple libraries. This unqualified call makes `GET /api/stacks` fail for those users even though individual stack lookups work.

Please discover the caller’s live libraries, enumerate stacks for each library, and apply the endpoint cap globally. A regression test should cover two libraries and require `library_id` on every stack-list call.

routers/api/stacks.py
```diff
@@ -1,30 +1,172 @@
...
+        stacks.append(stack)
+        if len(stacks) >= SEARCH_STACKS_CAP:
+            break
+    cap_hit = len(stacks) >= SEARCH_STACKS_CAP
+
+    hydrated = await hydrate_stacks(client, stacks)
```
taggartgorman, line 198: **[Should Fix] Bound total stack-member hydration work**

The 500-stack cap bounds stack rows, but every retained stack still triggers a complete paginated member walk. One unbounded user-created stack—or hundreds of moderately large stacks—can therefore cause excessive upstream requests and memory use.

Please establish a real total-work bound, preferably through an enforced stack-size limit or a bulk member-fetch operation. Individual member arrays should not be silently truncated because clients treat `assets.length` as the stack size.

routers/api/stacks.py
```diff
@@ -1,30 +1,172 @@
...
-    return []
+    if primaryAssetId is not None:
+        return await _search_by_primary_asset(client, primaryAssetId, current_user)
+
+    stacks = []
+    async for stack in client.stacks.list_stacks(limit=GUMNUT_API_MAX_PAGE_SIZE):
+        stacks.append(stack)
+        if len(stacks) >= SEARCH_STACKS_CAP:
+            break
+    cap_hit = len(stacks) >= SEARCH_STACKS_CAP
```
taggartgorman: **[Consider] Distinguish reaching the cap from truncating results**

With exactly 500 stacks this records `stack_cap_hit=True`, although no result was omitted. That conflicts with the documented meaning of the flag as the signal that the library has outgrown the endpoint.

Consider reading up to 501 rows, setting the flag only when the lookahead row exists, and hydrating the first 500. An exact-cap regression test would pin the boundary.

taggartgorman (commented):
The stack translation and effective-cover handling are thoughtfully implemented, and the new route tests cover the local behavior thoroughly. Before merging, the unfiltered route needs to support users with multiple Gumnut libraries, and the total member-hydration cost needs an explicit bound.

**[Should Fix] Update the stale stack sync comment**

`routers/api/sync/stream.py:147` still says “Gumnut has no stacks.” Stack REST reads now exist; what remains absent is a stack sync-stream source and per-asset stack fields. Please update the owning `_NOOP_REQUEST_TYPES` comment to match the revised architecture documentation.

I also noted an exact-cap observability issue below.

routers/api/stacks.py
taggartgorman: Skipping — the SDK requirement is real, but it is adapter-wide rather than a stacks defect. `assets.list` carries identical wording (`library_id` optional for one library, required above one), and every unqualified caller in the adapter passes none: `timeline.py:206`, `search.py:82/120/413/609`, `trash.py:148`, `assets.py:938/1044`, `sync/entity_fetch.py:61/90`, `memories.py:137`, `map_markers.py:87`. The only `library_id` in the tree is a placeholder in `oauth.py:194`, and `/api/libraries` is stubbed end to end.

Qualifying only this route would leave stacks enumerating correctly while the timeline, search, and sync still fail — the user still cannot use the app, and partial success is harder to diagnose than uniform failure. Recorded the constraint as `docs/architecture/adapter-architecture.md` § Single-library assumption instead, where the fix is scoped as threading a library through shared request context.

One thing I could not verify without the backend: whether `assets.list(stack_id=...)` also requires `library_id` above one library. If it does, `hydrate_stack` breaks too and even the scoped fix would not have left `GET /stacks/{id}` working.

routers/api/stacks.py
taggartgorman: Fixed — added `SEARCH_STACKS_MEMBER_BUDGET = 5000`, spent from each listing row's `asset_count` during the walk, so it stops on whichever bound binds first.

Both suggested remedies turned out not to be available: `assets.list` takes a singular `stack_id`, so there is no bulk member fetch in the SDK, and an enforced per-stack size limit would truncate member arrays — which you correctly ruled out. The lever that does work is the row's own `asset_count`: the size is known before anything is fetched, so the budget bounds work rather than measuring it afterwards.

Admission stays all-or-nothing per stack, so the real bound is the budget plus one stack, and a single stack outweighing the entire budget still hydrates whole — dropping a user's largest stack would be a worse answer than a slow one. That case now emits a warning rather than being absorbed silently. The budget is spent in live members while `fetch_stack_members` reads `state="all"`, so the summary also carries `stack_members_hydrated` next to `stack_members_budgeted`, making the gap measurable rather than assumed.

routers/api/stacks.py
taggartgorman: Fixed — both bounds are now tested before admitting a row rather than after appending it, so a library of exactly `SEARCH_STACKS_CAP` stacks exhausts the cursor and never sets the flag. Added the exact-cap regression test, plus the symmetric one for the member budget: self-review caught that the budget test passed identically under either ordering, leaving that half of the fix unpinned.

One addition beyond the suggestion: since two bounds can now stop the walk, `stack_cap_hit` would fire when the *member budget* truncated, sending an operator to the wrong constant. Renamed it `stack_search_truncated` and added `stack_truncated_by` (`stack_cap` / `member_budget` / null), since the remedies differ.

ternarybits (approved):

PR #325, opened 2026-07-30
Collapse burst stacks in the Immich time bucket

taggartgorman:
## What

- `GET /api/timeline/bucket` collapses each burst to one representative frame and badges it with Immich's `[stackId, liveCount]` tuple (both elements strings, as upstream emits them). Gated on `withStacked`, and the `stack` key is now omitted entirely when the flag is falsy rather than shipping `[None] * count`.
- New lean helpers in `routers/utils/stack_conversion.py` alongside the existing hydration path: chunked stack-row reads, a live-only member read, the timeline cover rule, and `resolve_timeline_stacks`. A fast path resolves the cover from assets already in the bucket, so the common case costs no extra upstream requests.
- Every way a stack can fail to resolve degrades to the same inert state — frames uncollapsed, tuple null — instead of hiding photos or failing the month.

## Why

Immich renders a burst as one tile with a frame-count badge, and it is the **server** that produces that shape: under `withStacked`, upstream's bucket query keeps an asset only when it is loose or is its stack's primary, then attaches the tuple to the survivor. The web client collapses nothing — it draws a badge and treats whichever asset carries the tuple as the primary.

That last part is why the tuple alone isn't enough. Emitting one per frame without collapsing would badge all five frames of a five-frame burst, each claiming to be the primary of the same stack — further from Immich's behavior than shipping no tuples at all. This PR was originally specified that way; reading the upstream source at the pinned version corrected it, along with two other premises about the client.

Both halves are gated on `withStacked` because upstream gates them together, and that turns out to be load-bearing rather than cosmetic: Immich web's album month view requests buckets *without* `withStacked` while still passing `showStackedIcon` to the thumbnail, so albums are badge-free upstream only because the field is absent. Collapsing unconditionally would drop burst frames out of album months.

**Two deliberate divergences from upstream**, both to avoid hiding photos:

- **Trash opts out of the stack path.** Upstream's predicate keys on the live primary, so a trashed non-primary frame gets filtered out of the trash view — the one place it has to stay visible, since trash is the only route back to restoring it. No Immich client requests `withStacked` on trash, so this costs nothing.
- **The timeline resolves its cover against live members only.** Both surfaces run the same rule, `resolve_effective_primary`; what differs is the member set. `/stacks` resolves against every member, so a trashed pin stays `StackResponseDto.primaryAssetId` and the response still lists the live frames beside it. Collapse is destructive in a way a listing isn't — it drops every frame that isn't the cover — so naming a trashed asset there erases the burst from the grid. Upstream has the same hole but promotes a new primary on permanent deletion, making it transient; the Gumnut API preserves a trashed pin until permanent deletion too, which stretches that window to the whole retention period.

**One divergence that isn't deliberate:** `GET /api/timeline/buckets` still counts every frame. Upstream applies the identical predicate to its count query so counts and contents agree by construction; `assets.counts` has no stack-aware filter and `assets.list` exposes only a single-stack `stack_id` filter, so computing the per-month deduction would mean paginating every stack in the library on a hot endpoint. The effect is cosmetic — the count feeds a pre-load height estimate the real layout replaces; the two things that still read the raw number are the scrubber, which snapshots it until a viewport resize, and `getRandomAsset()`, whose shuffle can pick an index past a collapsed month's contents and no-op for that press. Neither shows a wrong asset. The clean fix is a backend stack-collapse option on both calls.

## Degradation

Five per-stack cases (missing row, failed member read, past the read cap, undecodable ID, no live members) leave that stack uncollapsed with a null tuple. A failure of the stacks resource as a whole is caught at the route and degrades the entire bucket the same way — every frame kept, every tuple null, the `stack` key still present — rather than 5xx-ing the app's primary view. The route catches non-SDK exceptions too, at ERROR rather than WARNING, since reaching that clause means an adapter bug rather than an upstream outage.

Four of the five log one aggregate record per request rather than one per stack, each naming its own cause and carrying a sample to triage; the member-read record also names the exception, since a degraded assets resource that leaves `list_stacks` healthy never reaches the route-level guard and that record is the only signal there is. The no-live-members case is silent on purpose — in a live bucket it hides nothing.

## Test plan

`uv run pytest` (1614 passing), `uv run ruff check`, `uv run pyright` (0 errors), `uv run scripts/lint_docs.py`.

Coverage in `tests/unit/api/test_timeline.py` and `tests/unit/utils/test_stack_conversion.py`: the gate in both directions, collapse to pinned and unpinned covers, a pin naming no live frame, batching and chunking past the bulk-ID cap, the fast path vs. the fallback member read, badge count sourced from the row rather than the bucket *or* the resolved member set, column alignment after collapse, filter passthrough, both the total and in-flight bounds on the member-read fan-out, the exact-cap boundary, and each degradation path with its aggregate log record and severity.

Assertions were checked by reintroducing the defect each one names rather than by reading the assertion — that is how the badge-count and fan-out-concurrency gaps were found (both passed the full suite under the defect they claimed to pin).

Not manually exercised against a running Immich web client — the collapse and tuple shape are pinned against the upstream query and client code at the pinned tag rather than observed in a browser.


tests/unit/utils/test_stack_conversion.py
```diff
@@ -575,3 +602,481 @@ def test_trashed_member_fixture_is_actually_trashed():
...
+        routing the stack to the member read that resolves it correctly.
+        """
+        stack, members = _stack_with_members(
+            count=2, asset_count=2, trashed={0}, primary_asset_id=None
+        )
+        client = _timeline_client([stack], members_by_stack={stack.id: [members[1]]})
+
+        resolved = await resolve_timeline_stacks(client, members)
+
+        assert resolved.covers[stack.id] == members[1].id
```
taggartgorman, line 1147: **[Should Fix]** The badge count is pinned by no test — including the one named for it.

The tuple's count element is emitted as `str(row.asset_count)` (`routers/utils/stack_conversion.py:623`). On that line `len(members)` is also in scope. I checked every fixture that asserts a non-empty tuple:

| Test | `asset_count` | resolved live members |
|---|---|---|
| `test_complete_stack_resolves_without_a_member_read` | 3 | 3 |
| `test_count_comes_from_the_row_not_the_resolved_members` | 3 | 3 (read returns all) |
| `test_collapses_a_burst_to_its_cover` | 3 | 3 |
| `test_several_stacks_batch_and_preserve_asset_order` | 2 | 2 |
| `test_badge_counts_the_whole_stack_not_the_bucket` | 3 | 3 (read returns all) |

Every one varies the **bucket** member count against the row count — which is the defect the docstrings name — but never the **resolved** member count, which is the alternative actually available at line 623. So swapping `row.asset_count` for `len(members)` passes the entire suite.

This fixture is the only one where the two genuinely diverge: `asset_count=2` while the member read returns a single live member. It asserts only `covers`.

This is verbatim the trap the PR adds to `docs/references/code-practices.md` in the same diff — *"**Equal-by-fixture values**: if every fixture makes a source equal some other value in scope (a stack row's `asset_count` and the number of members resolved for it), swapping one for the other passes the suite — vary them in at least one test."*

One line here closes it:

```python
assert resolved.tuples[stack.id][1] == "2"
```

Under the `len(members)` defect that resolves to `"1"` and fails; `row.asset_count` gives `"2"`. Worth confirming by mutating line 623 rather than by reading the assertion.

routers/utils/stack_conversion.py
```diff
@@ -282,3 +296,348 @@ def build_stack_response(
...
+    """Pick the Gumnut asset ID of the frame that represents `stack` in a
+    collapsed timeline: the pinned cover when the pin is live, otherwise the
+    earliest live frame.
+
+    `live_members` must be the stack's **complete** live member set in ascending
+    capture order, since both rules read from it — an incomplete set can promote
+    a later frame, or miss a live pin and fall back. Returns `None` for a stack
+    with no live members, which has no frame to represent it.
+
+    Deliberately not `resolve_effective_primary`, which keeps a *trashed* pin so
```
taggartgorman: **[Should Fix]** This paragraph describes a rule divergence, but the divergence is entirely in the input.

Traced over a live-only member set — the contract this docstring itself imposes:

| Case | `select_timeline_cover` | `resolve_effective_primary` |
|---|---|---|
| pin present in the set | returns `pinned_id` | returns that member, whose `.id` **is** `pinned_id` |
| pin absent or `None` | returns `live_members[0].id` | clause 2 returns first `trashed_at is None` = `members[0]` |
| empty set | `None` | `None` |
| clause 3 (first trashed member) | — | unreachable on a live-only set |

Identical for every input the caller passes. And the only caller (`resolve_timeline_stacks`) always passes live-only members — either from `_bucket_members_by_stack`, which filters `trashed_at is None`, or from `fetch_live_stack_members`, which sends `state="live"`. **The liveness guarantee the timeline needs comes entirely from which member set is passed, not from a second selection rule.**

Two knock-ons:

1. "Deliberately not `resolve_effective_primary`, which keeps a *trashed* pin" implies this function does not keep one — but its pin loop has no `trashed_at` check either. Handed a trashed pin it returns it exactly as `resolve_effective_primary` would.
2. `test_trashed_pin_falls_back_instead_of_hiding_the_burst` calls the two with *different* inputs (`members` vs `members[1:]`) and its docstring says "Each function gets the input it really receives." Given the same live-only input both return `members[1]` — so the test pins no rule difference, despite that being its stated purpose.

Why this is worth changing rather than tolerating: the module docstring says the two steps live here "so the surfaces can't drift into disagreeing about which frame represents a burst." A second hand-written copy of the cover rule is precisely that drift risk — a later change to pin handling or the tie-break in `resolve_effective_primary` would leave `/stacks` and the timeline showing different frames for the same all-live burst, with nothing failing.

Suggested shape, keeping this function's real contribution (the input contract and the ID return type):

```python
def select_timeline_cover(stack, live_members) -> str | None:
    primary = resolve_effective_primary(stack, live_members)
    return primary.id if primary is not None else None
```

Then reframe this docstring and the architecture doc's "Timeline cover vs. effective primary" section: the divergence between the two surfaces is that the timeline resolves against the **live** member set while `/stacks` resolves against `state="all"` — the rule is shared. If the two rules genuinely must stay free to diverge later, say so explicitly and pin the difference with a test that gives both functions the *same* input.

routers/api/timeline.py
```diff
@@ -205,6 +209,30 @@ async def get_time_bucket(
...
+    # with a `[stackId, liveCount]` tuple; its client collapses nothing, so the
+    # tuple alone would badge every frame instead of showing one tile. Both
+    # halves are gated on `withStacked` and skipped for trash — see "Timeline
+    # stack collapse" in docs/architecture/adapter-architecture.md for why each
+    # of those is load-bearing.
+    stacks: TimelineStacks | None = None
+    if withStacked and not isTrashed:
+        try:
+            stacks = await resolve_timeline_stacks(client, filtered_assets)
+        except GumnutError:
```
taggartgorman, line 227: **[Consider]** This guard is one exception class narrower than the contract stated around it.

`resolve_timeline_stacks`, `TimelineStacks`, `_live_members_or_none`, and `docs/architecture/adapter-architecture.md` all say a stacks-resource failure "is caught at the route and degrades the entire bucket … rather than a 5xx on the app's primary view." That's implemented for SDK errors only.

No trigger exists today, and I checked carefully rather than assuming: `GumnutError` is the complete base for SDK failures (`APIStatusError`, `APIConnectionError`, `APITimeoutError`, `APIResponseValidationError` all derive from it), and `safe_uuid_from_stack_id`'s only failure mode is `ValueError`, already caught inside. So this is fragility, not a demonstrated bug.

What makes it worth the three lines: `resolve_timeline_stacks` is a new ~150-line function doing non-SDK work inside the same `try` scope — `zip(..., strict=True)`, and the `to_actual_utc(resolve_capture_datetime(a))` sort key — and its own comment concedes `strict=True` is there to name an invariant that could break. As written, a future line that can raise converts the designed degradation into a 500 on the main timeline, and no test would catch it.

The repo already has the two-clause shape for exactly this "degrade, never fail" pattern at `routers/api/auth.py:107-114`:

```python
except GumnutError:
    logger.warning("Failed to resolve timeline stacks; returning the bucket uncollapsed", exc_info=True)
    stacks = TimelineStacks(covers={}, tuples={})
except Exception:
    logger.error("Unexpected error resolving timeline stacks; returning the bucket uncollapsed", exc_info=True)
    stacks = TimelineStacks(covers={}, tuples={})
```

ERROR on the second clause also keeps a genuine adapter bug loud in Sentry while the expected upstream degradation stays at WARNING.

routers/utils/stack_conversion.py
```diff
@@ -282,3 +296,348 @@ def build_stack_response(
...
+
+
+def _bucket_members_by_stack(
+    assets: Iterable[AssetResponse],
+) -> dict[str, list[AssetResponse]]:
+    """Group a bucket's **live** stacked assets by stack ID, earliest first.
+
+    Trashed assets are skipped so the result is a live member set by
+    construction, whatever query produced `assets`. `select_timeline_cover`
+    requires that, and today it holds only because the bucket read inherits the
```
taggartgorman: **[Consider]** These two sentences contradict each other, and the weaker one is the risk.

Sentence 1 states the guarantee is unconditional — "Trashed assets are skipped so the result is a live member set by construction, whatever query produced `assets`." Sentence 2 says it "holds only because the bucket read inherits the Gumnut API's `state=\"live\"` default."

Sentence 1 is the accurate one. The loop guard `asset.stack_id is not None and asset.trashed_at is None` makes a trashed asset unreachable as a `bucket_members` entry regardless of what `state` the bucket read used. (Confirmed independently: `assets.list` defaults to `state="live"`, and the route's only `state` assignment is `"trashed"` under `isTrashed`, which the `withStacked and not isTrashed` gate excludes — so it's doubly protected today.)

The hazard sentence 2 creates: it invites a future maintainer to conclude the `trashed_at is None` guard is redundant ("`state=live` already handles it") and remove the only construction-level protection.

The real `state="all"` hazard also doesn't live in this function — it lives at the route, where trashed frames reaching `is_collapsed_away` would be dropped from the one view that must show them.

Suggest keeping sentence 1 and dropping sentence 2. If the hazard is worth documenting, move it next to the `withStacked and not isTrashed` gate in `routers/api/timeline.py`, phrased as what it actually is.

docs/architecture/adapter-architecture.md
```diff
@@ -329,6 +329,39 @@ The adapter applies this sort in memory before pagination slicing (see `_immich_
...
+#### Cost, and the bucket-count divergence
+
+Resolution usually costs nothing beyond the row read; `resolve_timeline_stacks` documents the fast path, the two shapes that reach the fallback member read, and the per-request cap on it.
+
+**`GET /api/timeline/buckets` counts are not collapsed.** Upstream applies the identical predicate to its count query so counts and contents agree by construction; the adapter cannot, because `assets.counts` has no stack-aware filter and `assets.list` exposes only a single-stack `stack_id` filter — there is no "belongs to any stack" filter to walk. Computing the per-month deduction would mean paginating every stack in the library plus a member read each, on a hot endpoint.
+
+Counts therefore overstate a month containing bursts. This is cosmetic: the client uses the count only for a pre-load month height estimate, which the real layout replaces once the bucket loads, and nothing compares the two numbers. The lasting artifact is the scrubber, which snapshots the count and refreshes only on a viewport resize. The clean fix is a backend stack-collapse option on both `assets.counts` and `assets.list`, which would return the adapter to a pass-through.
```
taggartgorman: **[Consider]** "nothing compares the two numbers" is stronger than what holds — there's a second consumer.

Verified at the pinned version (`v3.1.0`, matching `.immich-container-tag`; read from that tag directly, not the older local checkout):

`TimelineMonth.assetsCount` is `isLoaded ? <real contents> : #initialCount` — i.e. the server count for any month not yet loaded. `TimelineManager.getRandomAsset()` (`web/src/lib/managers/timeline-manager/timeline-manager.svelte.ts:430-460`) walks `month.assetsCount` to pick a global index, then indexes into the month's *loaded* days:

```ts
for (const month of this.months) {
  if (randomAssetIndex < accumulatedCount + month.assetsCount) { randomMonth = month; break; }
  accumulatedCount += month.assetsCount;
}
// … then over randomMonth.timelineDays, using day.viewerAssets.length
if (!randomDay) { return; }
```

With an inflated count a chosen index can fall past the collapsed month's real contents; the inner loop never matches, `randomDay` stays undefined, and `getRandomAsset()` returns `undefined`. `TimelineAssetViewer.svelte`'s `handleRandom` does `if (!randomAsset) return;` — so the shuffle button on the main timeline silently does nothing for that press.

No crash, no wrong asset, and albums are unaffected (album months don't send `withStacked`, so counts and contents still agree there). **The divergence itself is well-argued and I'm not asking for it to be fixed** — only that the doc not claim a stronger property, since this sentence is what a future reader will use to decide the gap is safe to keep ignoring.

Suggested wording: "…which the real layout replaces once the bucket loads. Two things still read the raw count: the scrubber, which snapshots it and refreshes only on a viewport resize, and `getRandomAsset()`, whose shuffle can land past a collapsed month's contents and no-op for that press. Neither shows a wrong asset."

No code change.

tests/unit/utils/test_stack_conversion.py
```diff
@@ -575,3 +602,481 @@ def test_trashed_member_fixture_is_actually_trashed():
...
+        assert broken.id not in resolved.tuples
+        # The degradation is invisible to the user by construction, so the
+        # record is the only signal an operator gets. Keyed on `extra` rather
+        # than message text so a reword doesn't break it.
+        (record,) = [r for r in caplog.records if hasattr(r, "failed_stack_count")]
+        assert getattr(record, "failed_stack_count") == 1
+        assert getattr(record, "attempted_stack_count") == 2
+
+    @pytest.mark.anyio
+    async def test_member_reads_are_capped_per_request(
```
taggartgorman, line 1018: **[Consider]** This pins the total read count but not the in-flight one.

`resolve_timeline_stacks` fans out up to `MAX_TIMELINE_STACK_MEMBER_READS` (50) member reads through `gather_with_concurrency`, whose semaphore holds in-flight calls to `BULK_FANOUT_CONCURRENCY_LIMIT` (10). This test asserts `call_count == 50` — the total — so swapping `gather_with_concurrency` for a plain `asyncio.gather` keeps every existing assertion green while opening 50 simultaneous upstream reads from a single request, on the timeline's hottest endpoint.

`hydrate_stacks` has exactly this guard (`test_bounds_concurrent_member_reads`), and its `_ConcurrencyTracker` / `_TrackedListing` helpers already live in this file — so the newer and hotter of the two fan-outs is the one left unguarded.

Reusing them in a test that forces enough partial stacks to exceed `BULK_FANOUT_CONCURRENCY_LIMIT`, asserting `tracker.peak > 1` and `tracker.peak <= BULK_FANOUT_CONCURRENCY_LIMIT`, mirrors the existing assertions.

tests/conftest.py
```diff
@@ -210,11 +221,32 @@ def make_gumnut_stack_members(
...
+    Filters by the `ids` each call asks for rather than replaying `rows`
+    wholesale, so chunked reads get their own chunk back — and so a test can
+    model a row that vanished between the asset read and the stack read simply
+    by leaving that stack out of `rows`.
+    """
+    rows_by_id = {row.id: row for row in rows}
+    return Mock(
+        side_effect=lambda **kwargs: MockSyncCursorPage(
+            [
+                rows_by_id[stack_id]
```
taggartgorman, line 270: **[Consider]** This ordering makes one of the production comments untestable.

`resolve_timeline_stacks` carries a load-bearing comment (`stack_conversion.py:527-532`) that classification walks `stack_ids` rather than `rows_by_id` precisely because "rows come back in the Gumnut API's own order" — and if that order were ever unstable, slicing the cap off a list built from them would resolve a *different* subset per request, making tiles appear and disappear across reloads. `test_member_reads_are_capped_per_request` asserts the resolved set is the front of the bucket to pin that.

But this comprehension builds its page in the order the caller *requested*, and `fetch_stack_rows` requests in `stack_ids` (bucket) order. So the production `rows_by_id` dict ends up in bucket insertion order under every test, and an implementation that built `partial_ids` by iterating `rows_by_id` would select the identical 50 stacks and pass. The mock makes the very reordering the comment defends against unrepresentable.

Returning the matched rows in a deliberately different order (e.g. `reversed(...)`) fixes it. Every consumer keys results by `row.id`, so no other assertion depends on the order.

docs/references/code-practices.md
```diff
@@ -468,6 +475,10 @@ For responses/errors from upstream Gumnut API calls, use status-based severity:
 
 When possible, use shared helpers in `routers/utils/error_mapping.py` (`upstream_status_log_level` / `log_upstream_response`) instead of ad-hoc `if/else` logging branches.
 
+### Per-item degradation on hot endpoints — log one aggregate record
+
+When a request degrades by skipping some of N items (a stack the timeline can't resolve, an entity a batch drops), log once per request with counts — `"%d of %d …"` plus `extra={"<thing>_count": n, "requested_<thing>_count": total, "sample_ids": ids[:10]}` — not once per item. Most causes of these are systemic rather than per-item, so a per-item record floods precisely when the signal matters, and on an endpoint a client calls repeatedly (timeline buckets, once per month scrolled) it multiplies again. The ratio is also what separates "all of them" from "one of them", which is the first thing triage needs. `resolve_timeline_stacks` carries the worked example across four such paths.
```
taggartgorman: **[Consider]** The prescribed `extra` template doesn't match the four records this cites as its worked example.

Checked all four:

| Record | keys |
|---|---|
| missing-row (`stack_conversion.py:518`) | `missing_stack_count`, `requested_stack_count`, `sample_stack_ids` |
| read-cap (`:552`) | `partial_stack_count`, `member_read_cap` — no denominator, **no sample** |
| failed-member-read (`:589`) | `failed_stack_count`, **`attempted_stack_count`**, `sample_stack_ids` |
| undecodable (`:636`) | `undecodable_stack_count`, `requested_stack_count`, `sample_stack_ids` |

Two mismatches against the template `{"<thing>_count": n, "requested_<thing>_count": total, "sample_ids": ids[:10]}`:

1. Every record uses `sample_stack_ids`, not `sample_ids`.
2. The `attempted_` vs `requested_` split is **meaningful**, not accidental — I verified it: `requested_stack_count` is `len(stack_ids)` (every stack in the bucket) while `attempted_stack_count` is `len(read_ids)` (only stacks a member read was issued for). Two genuine denominators warrant two names; the problem is the template has one slot, so the code *can't* follow it, and the "worked example across four such paths" claim holds for only one of the four.

Cheapest fix is doc-side: write the denominator slot generically (e.g. `"<denominator>_count": total`, with a clause noting it names whatever population the ratio is over — all items vs. only those attempted) and `"sample_<thing>_ids"` to match the code.

Separately, the read-cap record is the only one that drops the sample entirely, so an operator seeing it can't tell *which* stacks went uncollapsed — the exact triage affordance the other three provide. `sample_stack_ids: partial_ids[len(read_ids):][:10]` would make all four records genuinely the four-path example advertised here.

(Reserved-`extra`-keys rule checked and clean — none of the eight keys collide with a `LogRecord` attribute.)

docs/architecture/adapter-architecture.md
```diff
@@ -329,6 +329,39 @@ The adapter applies this sort in memory before pagination slicing (see `_immich_
...
+
+Every failure mode degrades the same way, because degrading to the pre-stack timeline always beats hiding photos behind a cover that may not exist. A stack is left unresolved — frames uncollapsed, tuple null — when its row is missing from the `list_stacks` response (dissolved between the two reads), when its member read fails, when it falls past the read cap, when its ID will not decode to an Immich UUID, or when it has no live members. A failure of the stacks resource as a whole is caught at the route and degrades the entire bucket the same way: every frame, no `stack` key contents, rather than a 5xx on the app's primary view.
```
taggartgorman: **[Consider]** "no `stack` key contents" points the wrong way for this path.

Line 343 establishes that "when the flag is falsy the `stack` key is absent from the response entirely." Having just read that, a client author reads "no `stack` key contents" here as the same omission.

But the whole-resource failure path sets `stacks = TimelineStacks(covers={}, tuples={})` (`routers/api/timeline.py:231`) — a non-`None` value — so `if stacks is not None: response["stack"] = stack_list` still emits the key, with one `None` per surviving asset. The key is present and full-length; only the *entries* are null.

The PR body states it correctly ("every frame, all tuples null") and `test_stack_read_failure_degrades_to_an_uncollapsed_bucket` pins the correct behavior (`result["stack"] == [None, None, None]`) — it's only this sentence that's ambiguous.

Suggest the PR body's wording: "…degrades the entire bucket the same way: every frame kept, every tuple null (the `stack` key is still present), rather than a 5xx on the app's primary view."

routers/api/timeline.py
```diff
@@ -205,6 +209,30 @@ async def get_time_bucket(
 
     filtered_assets = [a async for a in client.assets.list(**list_kwargs)]
 
+    # Immich collapses a burst to one tile server-side and badges the survivor
```
taggartgorman: **[Consider]** This comment paraphrases the doc it then points to.

The "server collapses, client only badges" explanation appears three times in this PR:

1. `docs/architecture/adapter-architecture.md:334` — canonical
2. here — paraphrases it, then links to the doc it just paraphrased
3. `tests/unit/api/test_timeline.py`, `TestTimeBucketStacks` docstring — paraphrases it again, then points to `get_time_bucket`

To be clear about what this is *not*: I measured the new section's prose-to-code ratio at 0.71, **below** this file's pre-existing 1.19 and in line with `asset_conversion.py` (0.56). The commentary volume is house style and isn't the issue. What survives is the specific defect — linking to a doc *and* restating its content is the same defect as restating it alone; keep the pointer, cut the restatement. `docs/references/documentation-conventions.md` says the same: "Explain a non-obvious rationale once and point to it elsewhere."

Three copies of an upstream-behavior claim will drift the next time the pinned Immich version changes what the client does — and this PR's own Why section is the argument for why that drift is expensive.

Suggest cutting to the facts needed at the call site plus the pointer, roughly: "Collapse and badge together, gated on `withStacked` and skipped for trash — see 'Timeline stack collapse' in `docs/architecture/adapter-architecture.md` for why each gate is load-bearing." Same for the test class docstring.

routers/utils/stack_conversion.py
```diff
@@ -282,3 +296,348 @@ def build_stack_response(
...
+        dict.fromkeys(asset.stack_id for asset in assets if asset.stack_id is not None)
+    )
+    if not stack_ids:
+        return TimelineStacks(covers={}, tuples={})
+
+    rows_by_id = {row.id: row for row in await fetch_stack_rows(client, stack_ids)}
+
+    # The asset read and the stack read are separate round-trips, so a stack
+    # dissolved in between leaves its former members carrying a stale
+    # `stack_id`. One aggregate record rather than one per stack: this endpoint
```
taggartgorman: **[Consider]** This rationale became a repo convention in the same PR; it doesn't need re-deriving here.

`docs/references/code-practices.md` now carries § "Per-item degradation on hot endpoints — log one aggregate record", stating the rule and its rationale (systemic causes, flooding when the signal matters, once-per-month-scrolled), and closing by naming `resolve_timeline_stacks` as the worked example. The same general rationale then appears five more times:

- here (lines 505-511) — near-verbatim the doc paragraph
- `:579-582` — "Aggregated for the same reason as the missing-row record above, and more urgently: …"
- `:626-629` — "Aggregated like the two records above, and this one needs it most: …"
- `_live_members_or_none`'s docstring closing line
- `docs/architecture/adapter-architecture.md:363`

The **site-specific deltas are load-bearing and should stay** — "a degraded assets resource that leaves `list_stacks` healthy never trips the route-level guard" and "unlike the member reads, nothing caps how many stacks reach this loop" are real and not derivable from the convention. It's the shared premise underneath them that's now in five places.

Suggest trimming this one to the fact unique to the site (the asset read and the stack read are separate round-trips, so a stack dissolved in between leaves stale `stack_id`s) and letting `code-practices.md` carry the aggregation rule. The two "for the same reason as above" comments then reduce cleanly to their deltas alone, and `adapter-architecture.md:363` can keep just the observable behavior.

taggartgorman (commented):
Reviewed across correctness/data-integrity/security/silent-failures, performance/test-coverage, and clarity/minimalism, with the upstream Immich source read at the pinned `v3.1.0`.

**The change is sound and the degradation design is the best part of it** — five distinct failure modes all converging on one inert state, each with an aggregate log record rather than a per-item flood, is the right shape for an endpoint hit once per month scrolled. The `withStacked` gate being load-bearing for album months, and trash opting out so a trashed non-primary frame stays restorable, are both correct calls that the upstream predicate would have gotten wrong.

Verified clean, so it doesn't need re-checking: `gather_with_concurrency` returns in input order (so `zip(..., strict=True)` can't mis-pair); `GumnutError` is the complete SDK exception base; `safe_uuid_from_stack_id` raises only `ValueError`; the `state="live"` default plus the `trashed_at is None` guard doubly protect the fast-path invariant; every `extra` key avoids `LogRecord` collisions; `#timeline-stack-collapse` resolves at both use sites; no PII or private-repo references in the diff or PR body.

**Two Should Fix**, both about the change describing itself more strongly than the code backs up — one test that can't fail for the defect it names, and one function whose docstring claims a rule divergence that is really an input divergence. Ten Consider items follow, mostly doc-accuracy and duplication.

---

**[Consider] `AGENTS.md` Documentation Map row** — reported here because it has no diff line to anchor to.

The PR adds a substantial new section ("Timeline stack collapse", ~30 lines, three subsections) to `adapter-architecture.md`, but its Map row still reads "Overall adapter design, auth and session handling, request observability…, trash/restore semantics, data translation, pagination, sync protocol, error handling, endpoint status." Someone grepping the Map for "stack" or "burst" finds nothing under Architecture and gets routed to `code-practices.md` (checksums) or the gap-analysis design doc instead.

`docs/references/documentation-conventions.md` says the Consult-when cell is "a routing trigger, not a summary — name the situations or concepts that should cause an agent to open the doc." Headroom exists: the cell is 210 characters against the `consult_cell_chars = 250` limit in `scripts/lint_docs.toml`, so appending e.g. `", timeline burst-stack collapse"` (~+34 chars) fits without a rewrite. CI's `map_cells` check passes because the row is unchanged — it does not detect a missing routing concept.

Verified clean elsewhere: `code-practices.md`'s cell already names "logging" and "project conventions," covering both of its new sections, and the gap-analysis row is unaffected.

tests/unit/utils/test_stack_conversion.py
taggartgorman: Fixed, and confirmed by mutation rather than by reading the assertion — swapping `row.asset_count` for `len(members)` passed all 131 tests, so the gap was real. `test_trashed_bucket_member_cannot_be_named_cover` now carries `assert resolved.tuples[stack.id][1] == "2"`; that mutation now fails exactly that test.

One thing your table understated: `test_count_comes_from_the_row_not_the_resolved_members` didn't just fail to pin the resolved count, its *name* claimed credit for it. Renamed to `..._not_the_bucket_members`, which is what it actually pins, with a docstring pointing at the fixture that carries the other half.

routers/utils/stack_conversion.py
taggartgorman: Fixed — you were right, and the fix went a step further than the suggestion.

Delegating to `resolve_effective_primary` as proposed, but liveness is now enforced *inside* `select_timeline_cover` rather than assumed of the caller. Trusting the docstring turned out to be worth removing: the shared rule keeps a trashed pin, and a cover the bucket cannot show collapses away every live frame of the burst — the one failure in this design that loses photos instead of degrading inertly.

Docstring, the architecture doc's "Timeline cover vs. effective primary" section, and `test_trashed_pin_falls_back_instead_of_hiding_the_burst` are all reframed around the divergence being the input.

Worth recording: my first attempt at that test asserted "same input, same answer" with a comment claiming it would catch a reintroduced hand-rolled rule. It doesn't and can't — on a live-only set the two are the same function, so no fixture distinguishes them. The comment now states what the test does pin and what nothing can.

routers/api/timeline.py
taggartgorman: Fixed. Added the second clause at ERROR, mirroring `routers/api/auth.py`.

Your framing held up under a second look: the promise is about the endpoint, not about which exception family broke it. Two additions beyond the suggestion —

- Self-review found the severity split was pinned in only *one* direction. Deleting the `except GumnutError` arm outright, or flipping it to ERROR, kept the whole suite green — so the two clauses, which differ only in log level, could have been collapsed by a future refactor with nothing objecting. Both directions are now asserted by logger name.
- Both handlers carry the month, bucket size, and filter shape. A burst of these records otherwise narrows to nothing.

On the general 5xx→ERROR rule in `code-practices.md`: the WARNING here is deliberate, and the comment now says so. That rule is for calls whose failure fails the request; this one is designed to fail, so the ERROR channel stays reserved for the adapter-bug clause.

routers/utils/stack_conversion.py
taggartgorman: Fixed — sentence 2 dropped, and nothing added at the route (the trash hazard is already covered in the architecture doc's "Trash opts out" paragraph, so moving it there would have been a third copy).

Self-review then caught that my replacement was also wrong, in the mirror-image way: it called the guard "load-bearing rather than a redundant echo of the bucket read's `state` filter", which after `select_timeline_cover` began filtering internally left the two functions each citing the *other* as the reason it filters, and neither stating the real one.

The real reason is the completeness check: `row.asset_count` is a live count, so counting a trashed bucket member toward `len(bucket_members) == row.asset_count` would let a partial stack classify as complete and resolve its cover from an incomplete set. A reviewer confirmed by deleting the guard — three tests fail. The docstring says that now, and notes the two guards protect different steps rather than duplicating each other.

docs/architecture/adapter-architecture.md
taggartgorman: Fixed, using close to your wording. I re-verified it independently at `v3.1.0` before editing rather than taking the trace on trust — and had to fetch the tag first, since the local sibling checkout was sitting on `v3.0.3`. `TimelineMonth.assetsCount` is `$derived(isLoaded ? … : #initialCount)`, and `getRandomAsset()` picks a global index off `month.assetsCount` then indexes into loaded days, so an inflated count can leave `randomDay` undefined and the press no-ops. Confirmed.

That near-miss became its own doc fix: `code-practices.md`'s upstream-verification section now says to read the pinned tag as `git show <tag>:<path>` rather than off the checkout's working tree, which answers for whatever release the clone was last left on without saying so.

Agreed on not fixing the divergence itself.

tests/unit/utils/test_stack_conversion.py
taggartgorman: Fixed, and the concern was right for the reason you gave: I confirmed by mutation that swapping `gather_with_concurrency` for a plain `asyncio.gather` kept all 131 tests green.

`test_bounds_concurrent_timeline_member_reads` reuses `_ConcurrencyTracker` / `_TrackedListing` as suggested, with `BULK_FANOUT_CONCURRENCY_LIMIT * 3` partial stacks (30, comfortably under the 50 cap so the two bounds stay independent), asserting `peak > 1` and `peak <= BULK_FANOUT_CONCURRENCY_LIMIT`. That mutation now fails exactly that test.

Self-review added the adjacent boundary you didn't ask for: `if len(partial_ids) > len(read_ids)` mutated to `>=` also passed the suite, and would have false-alarmed "leaving 0 stacks uncollapsed" with an empty sample on every exactly-50 bucket — on the endpoint this convention exists to keep quiet. Covered now too.

tests/conftest.py
taggartgorman: Fixed — `mock_list_stacks` now returns the matched rows reversed.

Checked your "no other assertion depends on the order" claim before applying it rather than after: `fetch_stack_rows`' own docstring states callers index by `row.id`, and the two consumers (`resolve_timeline_stacks`, `hydrate_stacks` via the `/stacks` routes) both do. Full suite passed unchanged.

The docstring was initially longer — it reproduced the cap-ordering rationale from `resolve_timeline_stacks` almost verbatim, which self-review flagged as the same duplication defect as the other threads here. It now states what the mock does and points at that comment for why it matters.

docs/references/code-practices.md
taggartgorman: Fixed on both sides — and the doc was wronger than this comment caught.

Doc: the sample key is now `sample_<thing>_ids`, and the denominator slot is generic. Your read that the `requested_`/`attempted_` split is meaningful rather than accidental was right, but a self-review pass found there are **three** live denominators, not two — the read-cap record's population is `partial_stack_count` (only the stacks classified partial, which is what the cap slices), which is neither of the two you named. The doc now names all three.

Code: the read-cap record gained `sample_stack_ids` as `partial_ids[len(read_ids):][:10]`, exactly as suggested. It also gained the numerator it was missing — the message reported "leaving %d uncollapsed" while `extra` carried no such field, so the one record whose ratio needs arithmetic was the one an operator couldn't query. Named `over_cap_stack_count` rather than `uncollapsed_stack_count`: all four records leave stacks uncollapsed, so the effect-shaped name would read as the total across every cause while reporting one.

Both new fields are asserted; mutating the numerator to `len(partial_ids)` fails the test.

docs/architecture/adapter-architecture.md
taggartgorman: Fixed with the PR body's wording, plus the contrast made explicit since line 343 is what sets up the misreading: "every frame kept, every tuple null (the `stack` key is still present, unlike the `withStacked`-falsy path where it is absent)".

routers/api/timeline.py
taggartgorman: Fixed at both sites — restatement cut, pointer kept, close to your suggested wording. The test class docstring got the same treatment.

The prose-to-code ratio measurement was a useful thing to include; it kept the discussion on the actual defect instead of on comment volume.

Self-review then caught me reintroducing the identical defect one thread over: my rewrite of `select_timeline_cover`'s docstring restated the architecture doc's "Timeline cover vs. effective primary" section *and* linked to it. Trimmed to the contract plus the pointer.

routers/utils/stack_conversion.py
taggartgorman: Fixed. The shared premise is gone from all five sites; `code-practices.md` carries the rule.

Kept exactly what you identified as load-bearing: "a degraded assets resource that leaves `list_stacks` healthy never trips the route-level guard" and "unlike the member reads, nothing caps how many stacks reach this loop" both survive as the deltas at their sites. `adapter-architecture.md:363` keeps the observable behavior and points at the convention.

One place the count went the other way: the member-read record needed a *new* reason, not fewer. Self-review flagged that swallowing the exception left `failed_stack_count` unable to separate an expired token from throttling from an outage — and since that path never reaches the route guard, as your comment notes, the record is the only signal there is. It now carries `exc_info` plus the error type and status, with the status read defensively since transport errors have none.

ternarybits (approved):

PR #328, opened 2026-07-31
Implement the Immich stack create and set-cover endpoints

taggartgorman:
## What

- `POST /stacks` groups assets into a new stack via the Gumnut API's `create_stack`, forwarding the requested IDs in order and pinning `assetIds[0]` as the cover.
- `PUT /stacks/{id}` pins a stack's cover via `set_cover`; an omitted or explicitly-null `primaryAssetId` degenerates to a plain read of the current stack.
- Both hydrate their response through the existing stack conversion layer, so a write reads back exactly like the matching `GET`.

## Why

The stack reads landed previously with the shared hydration and cover-resolution layer; these are the two mutations that return a full stack body, so they reuse that layer rather than growing a second notion of what a stack looks like on the wire.

Three decisions worth surfacing, all recorded in the handler docstrings:

- **The cover pin is the adapter's job, the rest is the backend's.** Immich specifies that the first member becomes primary and clients read `assets[0]` back as the cover, while the Gumnut API leaves a manually-created stack unpinned unless told otherwise — so the first ID is passed explicitly. Distinctness, the two-member minimum, library membership, and the already-stacked merge/repoint semantics are all forwarded untouched: both sides implement that merge, so re-deriving any of it here would only create a way for the two to disagree. The membership that comes back can exceed what was asked for, and the response is read off the returned row rather than off the request.
- **An over-cap create is refused, not chunked.** The backend caps the ID list, and unlike the adapter's other bulk-ID flows this one is deliberately not chunked. Creating a stack repoints its members out of whatever stacks they were in, so a failure partway can't be compensated by deleting the half-built stack — that dissolves it without restoring what was folded in. A clean failure that changed nothing beats a partial merge nobody can unwind.
- **The two routes differ on a stack with no live members.** Update 404s like `getStack`, since both of its paths end in a read. Create returns the stack anyway — the client just made it and needs the ID, and answering "not found" for a write that succeeded is the worse lie. That state needs a request naming nothing but trashed assets, which the backend permits and no shipped client can compose, so it now logs a warning rather than passing silently.

`PUT` is deprecated upstream but is still what the shipped clients call. Its replacement already exists in the pinned Immich source, excluded from the OpenAPI spec — which is the only reason generated clients still reach the `PUT`. That, and the fact that no spec diff will announce the switch, is now written down.

Delete, bulk-delete, and remove-asset remain stubs.

## Test plan

- `uv run pytest` — 1588 passed. New coverage in `tests/unit/api/test_stacks.py`: ID conversion and the first-as-cover pin, `library_id` never sent, the request forwarded once unmodified, the response following the backend's membership rather than the request (the fold-in case), backend rejections propagating with nothing read back, both wire shapes of an absent cover taking the read path, and the update 404 on both the read and set-cover paths.
- `uv run ruff check`, `uv run ruff format`, `uv run pyright` — clean.

Not exercised against a live backend; the SDK contract these routes depend on is pinned by the existing signature tests in `tests/unit/utils/test_stack_conversion.py`.


routers/api/stacks.py
```diff
@@ -242,13 +246,86 @@ async def delete_stacks(request: BulkIdsDto):
...
+    on a stack's cover appearing *anywhere* in `asset_ids`, not on its position:
+    worth knowing before reaching for a reordering the first ID doesn't need,
+    since its only special meaning is the cover pin above.
+
+    `library_id` is omitted, like every other Gumnut call the adapter makes —
+    see `docs/architecture/adapter-architecture.md` § "Single-library
+    assumption".
+
+    The one backend limit not left to fail late is worth naming: `asset_ids` is
+    capped at `GUMNUT_API_MAX_BULK_IDS`, so stacking a larger selection is a
```
taggartgorman: **[Consider]** The docstring asserts an upstream status the adapter doesn't itself enforce. `asset_ids` "is capped at `GUMNUT_API_MAX_BULK_IDS`, so stacking a larger selection is a 422" — but `GUMNUT_API_MAX_BULK_IDS` appears only here in the docstring; nothing adapter-side caps or length-checks the list, so the 422 depends entirely on the backend returning that exact code and the global handler forwarding it. That's consistent with how the adapter forwards upstream statuses, so the wording is plausibly correct — but it's an unverified cross-service claim, and citing an adapter constant the adapter doesn't reference in code invites a reader to grep for enforcement that isn't there. Worth confirming the backend returns 422 (not 400) for an over-cap `create_stack`; if that isn't guaranteed, describe the upstream rejection generically rather than naming the code.

routers/api/stacks.py
```diff
@@ -60,11 +60,15 @@ def _build_representable_response(
...
-    fetch — breaking both contracts `build_stack_response` documents, since
+    Every read-shaped route needs the same notion of "not representable" — the
+    list, the detail read, and the update route, whose two paths both end in a
+    read. `create_stack` deliberately opts out; its docstring says why.
+
+    It is wider than `hydrate_stack`'s member-less case: a stack whose members
+    are *all* trashed hydrates fine (`resolve_effective_primary` deliberately
+    keeps naming a cover) but converts to `assets: []` with a `primaryAssetId`
+    no client can fetch — breaking both contracts `build_stack_response`
+    documents, since
```
taggartgorman: **[Consider]** Ragged reflow introduced by this edit: the inserted paragraph left the surviving sentence unwrapped, so `since` dangles on its own short line — `...breaking both contracts \`build_stack_response\` documents, since` / `clients read \`assets.length\`...`. Purely cosmetic (the sentence still parses), but rewrapping so `documents, since clients read ...` flows normally restores the clean wrap that was on `main`.

taggartgorman (commented):
## Review: stack create + set-cover endpoints

Replaces the two fake-response stubs with real handlers backed by the Gumnut API's `create_stack` / `set_cover`, both hydrating their response through the same `stack_conversion.py` path as the reads so a write answers exactly like the matching GET. Membership rules, the merge/repoint semantics, and cover validation are left with the backend; the adapter's only added policy is the `assetIds[0]` cover pin, the null-cover read on `PUT`, and the create-vs-read divergence for an all-trashed stack. Tests are thorough and CI is fully green.

### Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (2 Consider items posted inline, both non-blocking)
- [x] **Codebase value**: After-state removes two stubs that returned fake stack bodies and reuses the existing hydration/cover-resolution layer instead of duplicating it. Alternatives considered: "do nothing" leaves the web timeline's multi-select Stack action collapsing a selection that vanishes on refresh; a smaller scope (create only) leaves the shipped "set as primary" action stubbed. The chosen scope closes both write paths that return a stack body while deferring the removal routes, which is the correct cut.
- [x] **Security/auth**: No authn/authz logic is defined or changed — both handlers wire the standard `get_authenticated_gumnut_client` + `get_current_user` dependencies, the same pattern the already-merged read routes use and pinned by `TestRouteDependencies` (now parametrized over `create_stack`/`update_stack`). Evidence: reviewed both handlers + the dependency test; the diff adds dependency wiring only, no permission/token handling.
- [x] **Data deletion**: No destructive operations. The diff touches only `create_stack` and `update_stack`; `delete_stack`, `delete_stacks`, and `remove_asset_from_stack` remain untouched no-op stubs.
- [x] **Confidence**: Correctness well understood — changes are confined to two handlers over a helper layer whose contract is pinned by `test_stack_conversion.py`; every new branch (500 on member-less read-back, 201-with-warning on all-trashed create, 404 on all-trashed update across both read and set-cover paths) has a direct test.
- [x] **Purpose clarity**: PR description and handler docstrings make the purpose and the three non-obvious decisions (cover-pin ownership, no-chunk-on-over-cap, create-vs-read divergence) explicit.

**Approve-equivalent** (posted as a comment because GitHub blocks approving your own PR). No blocking findings; the two inline notes are optional polish.

routers/api/stacks.py
taggartgorman: Fixed — reworded to describe the upstream rejection generically. Confirmed the adapter forwards the `asset_ids` list whole with no length check or chunking and never references `GUMNUT_API_MAX_BULK_IDS` here, so the specific 422/constant claim is gone.

routers/api/stacks.py
taggartgorman: Fixed — rewrapped so `documents, since clients read ...` flows normally; the dangling `since` line is gone.

ternarybits (approved):
Seems like a lot of doc edits, but they all look justified.

PR #329, opened 2026-07-31
Populate stack summary on asset detail

taggartgorman:
## Summary

Populate `AssetResponseDto.stack` on `GET /api/assets/{id}`, matching Immich v3.1.0 behavior.

- Resolve the asset's stack through the existing hydration and effective-cover policy.
- Return `{ id, primaryAssetId, assetCount }` for representable stacks.
- Return `stack=null` when the referenced stack disappears during resolution or has fewer than two live members.
- Leave update, search, memory, upload-event, and sync responses unchanged.

## Why

Immich's asset viewer uses the stack summary on asset detail to fetch the full stack and expose stack controls. Reusing the existing hydration path keeps asset detail consistent with `GET /api/stacks/{id}` without adding stack reads to unrelated response paths.

## Validation

- `uv run ruff format`
- `uv run ruff check`
- `uv run pyright`
- `uv run pytest` — 1621 passed
- `uv run scripts/lint_docs.py`


taggartgorman (commented):
Comprehensive review across correctness, data integrity, security, silent failures, performance, test coverage, clarity, and minimalism. No blocking issues — posting as **COMMENT** because two non-blocking Consider items are worth surfacing.

## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (2 Consider, inline below)
- [x] **Codebase value**: Leaves the codebase better than it found it. The `/stacks` read endpoints from #324 were unreachable because the nested `stack` block was null on every asset surface; this populates it consistently, so the burst badge renders and the viewer's stack controls appear. The batched-resolver / I/O-free-converter / degrade-posture split beats the alternatives — a per-route resolve would re-read shared stack rows and risk the same asset reading back stacked on one surface and loose on another; "do nothing" leaves the shipped read routes dead. Deferring the timeline tuples and `SyncAssetV1.stackId` keeps scope tight and is recorded in the gap analysis.
- [x] **Security/auth**: Does not touch authentication, authorization, or permission logic. Read-only enrichment of asset responses; no new trust boundary. Verified across `assets.py`, `search.py`, `memories.py`, `stack_conversion.py`, `asset_conversion.py`.
- [x] **Data deletion**: No destructive operations. The change only reads stack rows/members and joins a decorative field.
- [x] **Confidence**: Correctness well understood. `convert_gumnut_asset_to_immich` is the sole `AssetResponseDto` producer and every call site now passes a resolved lookup; the three null-stack paths are deliberate. Degrade catch is `GumnutError`-only (adapter bugs still 500), correctly scoped so a stacks 404 can't mask a real asset-not-found. Worst case bounded (row reads batched at 200/chunk; cover reads capped at 100 under concurrency 10). Extensive new tests, full suite green.
- [x] **Purpose clarity**: Purpose and approach are clear from the description, docstrings, and architecture doc.

routers/utils/stack_conversion.py
```diff
@@ -287,3 +324,378 @@ def build_stack_response(
...
+    criterion-less listing page clamp to the Gumnut per-page ceiling (200) and
+    memories tops out at 30 windows x 20 assets, but `/search/random` is capped
+    only by `RandomSearchDto.size` — up to **1000** assets, and so up to 1000
+    distinct unpinned bursts, in one request. `STACK_SUMMARY_COVER_READ_BUDGET`
+    bounds the reads that page can commit to; stacks past it ship `stack=None`,
+    the same degraded shape a dangling id already produces, and
+    `stack_summary_truncated` in the log below is the signal that a real library
+    is hitting it.
+
+    **A stack read that fails degrades the page rather than failing it.** The
```
taggartgorman: **[Consider]** The ~13-line failure-posture paragraph here ("A stack read that fails degrades the page rather than failing it..." through the `hydrate_stacks` contrast) is a verbatim copy of the same paragraph in `resolve_asset_stack_summaries`'s docstring above (the wrapper). Two things make the inner copy worth dropping:

1. **Drift hazard** — two word-for-word copies of one specific technical argument (global `GumnutError` handler forwarding status verbatim, the `GET /api/assets/{id}` "asset is gone" failure mode, `search_memories` tolerating a failed year) will diverge on the next edit to either.
2. **Mild self-contradiction** — the wrapper's docstring already says *"this wrapper owns the failure posture, which is the half worth explaining"*, and the inner function is the one that does **not** catch (it raises; the wrapper catches). So the degrade-posture explanation is misplaced here.

Suggest dropping this paragraph from `_resolve_asset_stack_summaries`, keeping only its resolving-mechanics content (dedup, chunked `list_stacks`, cover budget, representable filter), and leaving the single authoritative copy in the wrapper. Non-blocking.

tests/unit/utils/test_stack_conversion.py
```diff
@@ -483,6 +499,509 @@ async def test_all_trashed_stack_yields_empty_assets(self, mock_current_user):
...
+        become a silently missing block on every response instead of a 500."""
+        stack, members = make_gumnut_stack_with_members(count=2, primary_asset_id=None)
+        client = _summary_client([stack], {stack.id: members})
+        client.assets.list = Mock(side_effect=TypeError("adapter bug"))
+
+        with pytest.raises(TypeError):
+            await resolve_asset_stack_summaries(client, members)
+
+    @pytest.mark.anyio
+    async def test_cover_reads_are_capped_by_the_budget(
```
taggartgorman: **[Consider]** The over-budget side is well covered (this test uses budget+3, `test_budget_is_spent_in_page_order` uses budget+2), but nothing pins the exact-budget boundary: *exactly* `STACK_SUMMARY_COVER_READ_BUDGET` unpinned stacks admitting all summaries with **no** `stack_summary_truncated` warning. The budget guard is `cover_reads >= STACK_SUMMARY_COVER_READ_BUDGET` checked before increment — a `>` vs `>=` slip, or truncating the boundary row, would pass the whole suite. The code comment already flags this exact case ("why the flag would otherwise lie for a page of exactly the budget"), and the PR body notes it was verified only by temporarily restoring the old walk — a documented off-by-one trap with no committed regression guard.

Suggest a test with exactly `STACK_SUMMARY_COVER_READ_BUDGET` unpinned 2-frame stacks asserting `len(summaries) == budget`, `client.assets.list.call_count == budget`, and no WARNING carrying `stack_summary_truncated`. Non-blocking.

routers/utils/stack_conversion.py
taggartgorman: Fixed in 5cb7a6b — dropped the duplicated failure-posture paragraph from `_resolve_asset_stack_summaries`, leaving the wrapper `resolve_asset_stack_summaries` as the single owner of that explanation. Replaced it with a two-line pointer that also corrects the misplacement you noted: the inner function is where a stack-row lookup *propagates* (cover-read failures degrade per stack in `resolve_stack_cover`), so the pointer states that precisely rather than a blanket "raises on upstream failure".

tests/unit/utils/test_stack_conversion.py
taggartgorman: Added in 5cb7a6b as `test_exact_budget_admits_every_summary_without_truncating` — exactly `STACK_SUMMARY_COVER_READ_BUDGET` unpinned 2-frame stacks asserting `len(summaries) == budget`, `assets.list.call_count == budget`, and no record carrying `stack_summary_truncated`.

One correction from self-review: the boundary does not actually catch the `>` vs `>=` slip — at exactly the budget both comparators admit all rows and truncate none (they diverge only at the budget+1-th row), and the existing `budget+3` test already fails under a `>` slip via its `len==100`/`call_count==100` assertions. So the committed docstring credits the comparator coverage to the over-budget tests and states this test's real unique value: at the boundary every read is admitted and *no* truncation warning fires, which the always-truncating over-budget tests can't assert.

ternarybits (approved):

taggartgorman:
> Nice scope reduction.

Yeah, that was a "classic" Opus 5 explosion of scope. :-(

routers/api/assets.py
```diff
@@ -1192,7 +1193,26 @@ async def get_asset_info(
...
+                gumnut_asset.stack_id,
+                gumnut_asset.id,
+                extra={
+                    "stack_id": gumnut_asset.stack_id,
+                    "gumnut_id": gumnut_asset.id,
+                },
+            )
+    return convert_gumnut_asset_to_immich(
+        gumnut_asset, current_user, stack=stack_summary
+    )
```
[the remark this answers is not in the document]
taggartgorman: Fixed in `6e4705b` — the no-op/ignored-field PUT branch now retrieves and converts the asset directly, preserving `stack=null` and avoiding any stack lookup. The regression test now uses a stacked asset and asserts both behaviors.

PR #331, opened 2026-07-31
Implement the Immich stack removal endpoints

taggartgorman:
## What

- Replace the three stack-removal stubs in `routers/api/stacks.py` with real SDK forwarding:
  - `DELETE /api/stacks/{id}` dissolves one stack (204, no body).
  - `DELETE /api/stacks/{id}/assets/{assetId}` pulls a single frame out via a one-element `remove_assets` call (204).
  - `DELETE /api/stacks` bulk-dissolves many: dedupes ids preserving request order, fans out one single-stack delete per id under the shared `gather_with_concurrency` bound, and — since neither Immich nor the Gumnut SDK promises atomicity — lets every started call settle, logs how many of N deletes failed, then raises the first error in request order.
- Add tests covering single dissolve, single-asset removal, backend-triggered dissolution needing no second call, non-member silent success, and the bulk cases (empty no-op, dedupe/order, bounded fan-out, all-success, deterministic partial-failure with a failure-count log assertion, non-SDK abort).
- Update `docs/architecture/adapter-architecture.md` and `docs/design-docs/immich-adapter-gap-analysis.md` to mark all seven stack REST endpoints implemented.

## Why

This completes the seven-operation Immich REST stack surface (the read and create/set-cover endpoints already shipped). The removal operations dissolve only the grouping — the photos are untouched — and cover clearing, sub-two dissolution, and user-origin exemption stay owned by the backend, so the adapter forwards and returns 204. Missing or foreign stacks map to a 404 through the global error handler, matching the read routes. Bulk delete can't be atomic, so it surfaces a deterministic (request-order) error the client can reproduce rather than whichever failure happens to land first, while allowing earlier deletes to commit; a warning logs the aggregate failure count so partial failures leave a triage signal. REST completion is not end-to-end completion: the timeline count and sync-stream (`SyncAssetV1.stackId`) limitations are unchanged and remain called out in the docs.

## Test plan

- `uv run ruff format` / `uv run ruff check` — clean
- `uv run pyright` — 0 errors
- `uv run pytest` — passes (74 in `tests/unit/api/test_stacks.py`)
- `uv run scripts/lint_docs.py` — passes


routers/api/stacks.py
```diff
@@ -220,11 +221,49 @@ async def search_stacks(
...
+        try:
+            await client.stacks.delete(gumnut_stack_id)
+            return None
+        except GumnutError as exc:
+            return exc
+
+    errors = await gather_with_concurrency(
+        [_delete(gumnut_stack_id) for gumnut_stack_id in gumnut_stack_ids]
+    )
+    first_error = next((error for error in errors if error is not None), None)
```
taggartgorman: **[Consider]** Partial bulk-delete failures beyond the first are silently discarded.

Unlike `delete_people` (which never holds its errors — it lets `gather` propagate), this path catches every `GumnutError` and materializes them all in `errors`, then surfaces only the first-in-request-order one to the global handler. In a non-atomic batch where several stacks fail for different reasons, production triage sees exactly one error with no signal that, say, 7 of 20 deletes failed — and this is *not* the unavoidable silent-sibling discard `gather_with_concurrency` documents, since the errors are already in hand here.

Since you're already holding every error object, a summary log before `raise first_error` is nearly free:

```python
failures = [e for e in errors if e is not None]
if failures:
    logger.warning(
        "bulk stack dissolve: %d of %d deletes failed; surfacing the first in request order",
        len(failures), len(gumnut_stack_ids),
        extra={"failed": len(failures), "requested": len(gumnut_stack_ids)},
    )
    raise failures[0]
```

Non-blocking — the current behavior matches the established 204-contract convention.

taggartgorman (commented):
Reviewed the three stack-removal endpoints, the concurrency/bulk conventions they build on, the test matrix, and the doc updates.

This is clean, well-tested work that completes the REST stack surface. The approach reuses `gather_with_concurrency` correctly, routes errors through the global handler consistently with the sibling read/write routes, and the deliberate request-order-deterministic first-error selection is a nice touch (and thoroughly tested — the slow-early/fast-late ordering assertion is exactly right). The hand-rolled error loop correctly avoids `bulk.py`'s helpers, since this 204 endpoint needs the raw exception to re-raise rather than a flattened `BulkIdErrorReason`. Docs are accurate and the endpoint accounting checks out.

One non-blocking observability note below. Approving in spirit — nothing here blocks merge.

routers/api/stacks.py
taggartgorman: Fixed — `delete_stacks` now logs a `warning` summarizing how many of N deletes failed (with structured `extra`) before raising `failures[0]`, so partial non-atomic failures leave a triage signal. Added a caplog assertion (`2 of 4 deletes failed`) to the request-order test. Behavior otherwise unchanged (still first error in request order).

ternarybits (approved):

PR #334, opened 2026-08-03
Add Immich mobile stack snapshot to the sync stream

taggartgorman:
## What

- Sync asset rows now carry their `stackId` (Gumnut stack FK → Immich UUID) in both `SyncAssetV1` and `SyncAssetV2`; loose assets stay null. An undecodable stack_id degrades the asset to loose rather than aborting the stream.
- New `StacksV1` pass (`_stream_stacks`, before the asset loop): emits `StackV1` rows as **upsert-only incremental** delivery — every sync re-pages the stack table, sorts by `(updated_at, id)`, and emits rows after the checkpoint cursor. `primaryAssetId` is resolved by the same effective-primary helper the REST `/stacks` surfaces use; member-less stacks are skipped with a warning.
- `StacksV1` moved out of `_NOOP_REQUEST_TYPES` into the special current-state handling; `PartnerStacksV1` stays a no-op.

## Why

Immich mobile groups burst members under a stack via each asset's `stackId` and the `StacksV1` stream. Neither existed in the adapter — assets emitted `stackId=None` and `StacksV1` was a no-op, so bursts never collapsed on mobile.

Gumnut emits no stack lifecycle events yet, so stacks are delivered by re-paging the table each sync rather than as an event delta. The delivery is **upsert-only incremental**, not reset-only: asset rows carry `stackId` from current entity state unconditionally, so a burst created *after* the initial sync would otherwise leave its member assets pointing at a stack row the client never received — and the mobile timeline join (`... AND (rae.stack_id IS NULL OR rae.id = se.primary_asset_id)`) hides **every frame** of a burst whose stack row is missing. Emitting rows after the checkpoint cursor keeps new bursts visible and makes the pass resumable after a truncated stream. **Delete detection stays out of scope** (a separate event-support track): a dissolved stack's row goes stale but is benign under that same join, and the full table is never re-emitted to express a delete.

Because the pass runs before the asset loop, a Gumnut API error while hydrating one stack is caught at the call site and ends the pass gracefully — already-streamed rows are acked, the asset loop still runs, and the remaining stacks resume next sync — rather than truncating the whole sync. Stacks are emitted before assets for the asset→stack direction; the mobile client's `stackId`/`primaryAssetId` columns are unenforced indexes, so the circular reference isn't order-sensitive in practice.

Full rationale and scope boundary: the "Stack Snapshot (StacksV1)" section of `docs/architecture/sync-stream-architecture.md`.

## Test plan

- `tests/unit/api/sync/test_sync_stream_stacks.py`: V1/V2 asset stackId + loose null, undecodable-stack_id degrade, stack converter field mapping; snapshot with pinned vs synthesized primary, zero-member skip + warning; **incremental** emission (only stacks after the cursor), **caught-up** checkpoint emits nothing but still pages and pays no member fetches, **id tiebreak** at equal `updated_at`, undecodable-id skip without aborting (bad stack sorted first), **graceful degrade** on a hydration `GumnutError` (asset loop still runs, `SyncCompleteV1` intact); StackV1-before-stacked-asset ordering, `PartnerStacksV1` silent no-op.
- Updated the two no-op invariant tests (`StacksV1` is no longer a no-op) and conftest defaults.
- Full suite green: 1679 passed; ruff, pyright, and docs lint clean.

### DIFF
--- a/docs/architecture/sync-stream-architecture.md
+++ b/docs/architecture/sync-stream-architecture.md
@@ -73,56 +73,19 @@ delete from the album-user pass would duplicate `AlbumDeleteV1`. No
 
 ## Stack Snapshot (StacksV1)
 
-Stacks reach the mobile client two ways, both driven from the initial/full sync:
+Each sync asset carries its stack membership in `stackId`. Before streaming
+assets, `_stream_stacks` re-pages the stack table and emits rows after the
+client's checkpoint in `(updated_at, id)` order. This provides incremental
+upserts and resumes a truncated pass without a stack lifecycle event stream.
 
-1. **Asset membership.** Each sync asset row carries its `stackId` — the Gumnut
-   `stack_id` FK mapped to the Immich UUID (loose assets stay null). V2 inherits
-   it through the V1 delegation. This is what groups burst members under a stack
-   on the client.
-2. **Stack rows.** `StacksV1` streams `StackV1` rows, handled specially
-   (`_stream_stacks`) alongside the user types rather than via the events API —
-   the Gumnut backend emits no stack lifecycle events yet, so there is nothing to
-   feed the event stream. Each row's `primaryAssetId` is the **effective primary**
-   resolved by the shared `resolve_effective_primary`/`hydrate_stack` helpers (the
-   exact rule the REST `/stacks` surfaces use), so an auto-detected burst with no
-   pinned cover still gets a valid non-null primary. A member-less stack has no
-   honest primary and is skipped with a warning. `ownerId` is always the current
-   user (Gumnut is single-user; no stack sharing).
+Each row uses the same effective-primary resolution as the REST stack endpoints;
+member-less stacks are skipped because `primaryAssetId` is required. Incremental
+upserts prevent assets from referencing a missing stack, which would hide the
+burst in the mobile timeline.
 
-The stack rows are emitted **before** the phase-1 asset loop, so a stack row
-exists on the client before any asset naming it via `stackId` arrives. (The
-mobile client's `stackId` / `primaryAssetId` columns are unenforced indexes, not
-FKs, so the circular reference is not order-sensitive in practice — stacks-first
-is a deliberate, defensive choice for the asset→stack direction.)
-
-**Scope boundary — upserts only, no delete detection.** With no lifecycle event
-stream, `_stream_stacks` re-pages the whole stack table on **every** sync and
-emits the rows the client has not yet acked: it orders by `(updated_at, id)` and
-skips rows at or before the `StackV1` checkpoint cursor. This is upsert-only
-incremental delivery, and it exists for a concrete reason — asset rows carry
-`stackId` from current entity state unconditionally, so a burst created *after*
-the initial sync would otherwise leave its member assets pointing at a stack row
-the client never received. A dangling `stackId` is not benign: the mobile
-timeline join (`merged_asset.drift.dart`,
-`... AND (rae.stack_id IS NULL OR rae.id = se.primary_asset_id)`) hides **every
-frame** of a burst whose stack row is missing. Emitting the new stack keeps the
-burst visible and grouped.
-
-Ordering by the ack cursor also makes the pass **resumable**: if a snapshot
-truncates mid-stream, the checkpoint holds the last acked row's cursor and the
-next sync skips at-or-before it, so the un-streamed tail is not lost. This is
-also why a Gumnut API error while hydrating one stack is caught at the call site
-and ends the pass gracefully rather than propagating: the snapshot runs *before*
-the asset loop, so an unhandled error would suppress assets/albums/faces too —
-instead the already-streamed rows are acked, the asset loop still runs, and the
-remaining stacks resume next sync.
-
-**Delete detection stays out of scope**, tracked on a separate event-support
-track. A dissolved stack's row goes stale but is harmless: unstacking clears each
-member's `asset.stack_id` via asset events, and a stack row no asset references
-satisfies the `rae.stack_id IS NULL` branch of the join. `PartnerStacksV1` stays
-a no-op (partner sharing is unsupported), and `on_asset_stack_update` is **not**
-in the supported WebSocket table.
+Delete detection remains unsupported. A dissolved stack row is harmless once
+asset events clear its members' `stackId`. `PartnerStacksV1` remains a no-op
+because partner sharing is unsupported.
 
 ## Adding a New Sync Type Version
 
@@ -140,7 +103,10 @@ The invariant tests in `test_sync_v2.py` assert that every V2 type in `_SYNC_TYP
 
 Immich sync types that are accepted but have no Gumnut equivalent (e.g., `AssetEditsV1` — we don't support editing) go in `_NOOP_REQUEST_TYPES` in `stream.py`. This prevents "unsupported type" warnings while making the no-op explicit. Do not just add them to `_SUPPORTED_REQUEST_TYPES` without `_SYNC_TYPE_ORDER` — that silently drops them.
 
-The v3 mobile client requests a broad set of these on **every** sync (partner-*, `PartnerStacksV1`, memories-*, `AssetMetadataV1`, `AssetOcrV1`, `AlbumAssetExifsV1`, the V2 partner/album-asset variants) — all no-ops because the adapter has no sync-stream source for them, whether the underlying feature is absent (sharing, OCR) or reachable only through the REST surfaces (memories). They all belong in `_NOOP_REQUEST_TYPES` so the per-sync "unsupported types" warning stays quiet. Two requested types the adapter *does* handle are deliberately **not** no-ops, both handled specially alongside `UsersV1`/`AuthUsersV1` rather than via `_SYNC_TYPE_ORDER`: `UserMetadataV1` (synthesized preferences row — see "User Preferences" above) and `StacksV1` (current-state snapshot — see "Stack Snapshot" above). `PartnerStacksV1` remains a no-op regardless.
+The v3 mobile client requests these types on every sync, so keep unsupported
+ones in `_NOOP_REQUEST_TYPES` to avoid repeated warnings. `UserMetadataV1` and
+`StacksV1` are handled specially outside `_SYNC_TYPE_ORDER` and must not be
+listed as no-ops.
 
 ## Contract with the Gumnut API
 
--- a/docs/architecture/sync-stream-architecture.md
+++ b/docs/architecture/sync-stream-architecture.md
@@ -71,6 +71,42 @@ pass: an `album_deleted` event emits `AlbumDeleteV1`, and the client's
 delete from the album-user pass would duplicate `AlbumDeleteV1`. No
 `AlbumUserDeleteV1` is emitted (Gumnut has no unshare operation).
 
+## Stack Snapshot (StacksV1)
+
+Stacks reach the mobile client two ways, both driven from the initial/full sync:
+
+1. **Asset membership.** Each sync asset row carries its `stackId` — the Gumnut
+   `stack_id` FK mapped to the Immich UUID (loose assets stay null). V2 inherits
+   it through the V1 delegation. This is what groups burst members under a stack
+   on the client.
+2. **Stack rows.** `StacksV1` streams a **current-state snapshot** of `StackV1`
+   rows, handled specially (`_stream_stacks`) alongside the user types rather
+   than via the events API — the Gumnut backend emits no stack lifecycle events
+   yet, so there is nothing to feed the event stream. Each row's
+   `primaryAssetId` is the **effective primary** resolved by the shared
+   `resolve_effective_primary`/`hydrate_stack` helpers (the exact rule the REST
+   `/stacks` surfaces use), so an auto-detected burst with no pinned cover still
+   gets a valid non-null primary. A member-less stack has no honest primary and
+   is skipped with a warning. `ownerId` is always the current user (Gumnut is
+   single-user; no stack sharing).
+
+The snapshot is emitted **before** the phase-1 asset loop, so a stack row exists
+on the client before any asset naming it via `stackId` arrives. (The mobile
+client's `stackId` / `primaryAssetId` columns are unenforced indexes, not FKs,
+so the circular reference is not order-sensitive in practice — stacks-first is a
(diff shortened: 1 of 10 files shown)


routers/api/sync/stream.py
```diff
@@ -503,6 +504,75 @@ def _yield_buffered_deletes(
...
+      library ever makes reset latency matter.
+    - A client with no stacks (or only member-less ones) emits no StackV1 row, so
+      never acks one, so never stores a checkpoint — it re-pages ``list_stacks``
+      (one empty page) on every sync until it has a real stack to checkpoint.
+      A consequence: the first stack a stackless client gains after its initial
+      sync does propagate on the next delta (there is no checkpoint yet to
+      suppress it) — strictly more than the "snapshot only on reset" guarantee,
+      never less, so it is harmless.
+    """
+    if checkpoint is not None:
```
taggartgorman: **[Must Fix]** Checkpointed clients get dangling asset `stackId`s that hide new bursts from the mobile timeline

Once a client holds a StackV1 checkpoint, this gate suppresses stack rows forever — but the asset converters now unconditionally emit `stackId` from current entity state. Any stack created after the initial sync (a newly uploaded burst, or existing assets stacked by the burst detector and later re-synced by any asset event, since the adapter batch-fetches current state) produces asset rows referencing a stack row the client will never receive.

Verified against the mobile timeline query (`mobile/lib/infrastructure/entities/merged_asset.drift.dart`): `LEFT JOIN stack_entity se ON rae.stack_id = se.id ... AND (rae.stack_id IS NULL OR rae.id = se.primary_asset_id)` — with no stack row, the predicate is false for every member, so **all frames of the burst disappear from the timeline**. That's a regression vs. pre-PR behavior (stackId always null → all frames visible). The PR body's "unenforced indexes, not order-sensitive" point covers insert ordering, but not this join semantics.

Suggested fix: don't make checkpoint *presence* suppress the pass. The ack cursor already encodes `updated_at` — on checkpoint present, emit only stacks with `updated_at` after the checkpointed value (upsert-only incremental). Deletes can stay out of scope: unstacking clears `asset.stack_id` via asset events, and a stale stack row no asset references is benign under the same join.

routers/api/sync/stream.py
```diff
@@ -503,6 +504,75 @@ def _yield_buffered_deletes(
...
+        # An existing checkpoint means the client already holds the snapshot;
+        # emit no speculative updates. See the docstring's scope boundary.
+        return
+
+    async for stack in gumnut_client.stacks.list_stacks(limit=GUMNUT_API_MAX_PAGE_SIZE):
```
taggartgorman: **[Should Fix]** Interrupted snapshot + batch acks can permanently lose the un-streamed tail

The StackV1 checkpoint's only semantics is presence — its cursor value is never consulted. If the stream truncates mid-snapshot (e.g. a transient 500 from `hydrate_stack`'s member fetch reaches the post-200 broad except, ending the stream without `SyncCompleteV1`) after the client has acked any StackV1 row — the mobile client acks in 5,000-line batches, and this pass does one member-fetch round trip per stack, so the window is long on a burst-heavy library — the checkpoint then exists and every remaining stack is suppressed until a manual reset. Unlike every other multi-row type, there's no resume.

Suggested fix: make the checkpoint a resume cursor — the `updated_at`+id ack was chosen to be stable; order the snapshot by `(updated_at, id)` and skip rows at-or-before the checkpointed cursor. This composes naturally with the incremental fix suggested above. (Per-stack catch-and-skip alone is *not* a safe fix here — it acks past silently dropped rows.) If instead this is an accepted gap, document it in the trade-off list and pin the truncation behavior with a test.

taggartgorman: **[Should Fix]** An undecodable stack row ID aborts the whole sync stream — the exact failure mode this PR guards on the asset side

`_immich_stack_id` exists (with a docstring, a code-practices section, and a dedicated test) precisely because a `ValueError` inside the streaming generator truncates the stream so every subsequent sync fails identically. But this loop hits the same decoder unguarded, twice: `hydrate_stack` and `gumnut_stack_to_sync_stack_v1` both call `safe_uuid_from_stack_id(stack.id)`. Same systemic trigger (a backend prefix change), worse blast radius: this pass runs before the phase-1 asset loop, so on a first sync the entire library — assets included — never syncs, repeatably.

Suggested fix: wrap the per-stack body in `try/except ValueError` (scoped to `ValueError` only, not `GumnutError` — see the resume-cursor comment above), skip the stack, and aggregate one warning after the loop, mirroring `resolve_timeline_stacks`' `undecodable_ids` handling. This is also self-consistent: if the prefix breaks, `_immich_stack_id` already degrades every asset to loose, so emitting no stack rows matches.

taggartgorman (commented):
Solid, well-tested implementation of the stack snapshot, and the docs/tests are thorough. One blocking-severity data-integrity issue in how the checkpoint gate interacts with post-sync stack creation, plus two Should Fix items in the same area of `_stream_stacks`, one doc-structure Should Fix, and three minor Considers.

The headline issue was verified against the Immich mobile client's actual timeline query, so it's not theoretical: a dangling `stackId` doesn't just fail to group — it hides every frame of the burst from the timeline.

Summary: 1 Must Fix · 3 Should Fix · 3 Consider (inline below).

routers/api/sync/stream.py
```diff
@@ -503,6 +504,75 @@ def _yield_buffered_deletes(
             yield json_line, delete_type
 
 
+async def _stream_stacks(
```
taggartgorman, line 505: **[Should Fix]** Scope-boundary rationale is written into four homes that will drift

The same argument (no stack lifecycle events yet → reset-only snapshot; checkpoint present emits nothing; full-table re-emit can't express deletes; separate event-support track; stacks-before-assets; member-less skip; `updated_at`+id acks) appears near-verbatim in the arch doc's Stack Snapshot section, this ~45-line docstring, the test module docstring, and the sync-communication doc note. The copies already diverge — only this docstring documents the stackless-client consequence, while the arch doc states absolutely "streams only on reset / first sync". Per `documentation-conventions.md`, explain a non-obvious rationale once and point to it.

Suggested: make the arch doc section canonical (move the stackless-client nuance there); trim this docstring to the behavioral contract plus the two code-local trade-offs with a pointer; cut the test module docstring to one sentence + pointer; in `immich-sync-communication.md` keep only the table-level delta (StackV1 upserts only, never StackDeleteV1, reset-only) + the pointer.

tests/unit/api/sync/test_sync_stream_stacks.py
```diff
@@ -0,0 +1,336 @@
...
+def _members_by_stack(mapping):
+    """A ``client.assets.list`` mock keyed by the ``stack_id`` hydrate_stack asks for."""
+    return Mock(
+        side_effect=lambda **kwargs: MockSyncCursorPage(
+            mapping.get(kwargs.get("stack_id"), [])
+        )
+    )
+
+
+async def _collect(gen):
```
taggartgorman: **[Consider]** `_collect` duplicates `collect_stream`

This helper is functionally identical to `collect_stream` from the sync conftest, which this file already imports and uses. Suggest deleting `_collect` and calling `collect_stream(_stream_stacks(...))`; the one test needing raw lines already iterates the generator directly.

docs/architecture/sync-stream-architecture.md
```diff
@@ -71,6 +71,42 @@ pass: an `album_deleted` event emits `AlbumDeleteV1`, and the client's
 delete from the album-user pass would duplicate `AlbumDeleteV1`. No
 `AlbumUserDeleteV1` is emitted (Gumnut has no unshare operation).
 
+## Stack Snapshot (StacksV1)
```
taggartgorman, line 74: **[Consider]** `last-updated` not bumped on two meaningfully-edited docs

This doc (+35 lines) and `code-practices.md` (new exception paragraph) keep `last-updated` dates from earlier PRs; only `immich-sync-communication.md` was bumped. The docs linter passes only because both dates fall inside the 7-day freshness window — but that window exists to avoid churn re-bumps on the same branch, and these dates came from prior PRs on main. Suggest setting both to 2026-08-03 manually (`lint_docs.py --fix` won't, since it considers them fresh).

tests/unit/api/sync/test_sync_stream_stacks.py
```diff
@@ -0,0 +1,336 @@
...
+    async def test_partner_stacks_v1_is_silent_noop(self, caplog):
+        """PartnerStacksV1 stays an intentional no-op: no rows, no 'unsupported'
+        warning, and it never triggers the stack snapshot."""
+        updated_at = datetime(2025, 1, 15, 10, 0, 0, tzinfo=timezone.utc)
+        user = create_mock_user(updated_at)
+        client = create_mock_gumnut_client(user)
+        client.stacks.list_stacks = _stacks_page(make_gumnut_stack())
+
+        request = SyncStreamDto(types=[SyncRequestType.PartnerStacksV1])
+        import logging
```
taggartgorman: **[Consider]** Inline `import logging` vs. the file's string-level convention

Every other `caplog.at_level` in this file passes a string ("DEBUG", "WARNING"); this test imports `logging` mid-body just for `logging.INFO`. Suggest `caplog.at_level("INFO", logger="routers.api.sync.stream")` and dropping the import.

routers/api/sync/stream.py
taggartgorman: Fixed in `7bd3e95` — `_stream_stacks` is now upsert-only incremental: it sorts by `(updated_at, id)` and emits rows after the checkpoint cursor instead of suppressing the pass entirely. A burst created after the initial sync now reaches the client, so its assets no longer carry a dangling `stackId`. Deletes remain out of scope.

routers/api/sync/stream.py
taggartgorman: Fixed in `7bd3e95` — the checkpoint cursor is now consulted (not just its presence). Because rows emit in `(updated_at, id)` order, the ack advances monotonically, so a truncated snapshot resumes from the last ack via skip-at-or-before. Composes with the incremental fix above.

taggartgorman: Fixed in `7bd3e95` — `stack.id` is decoded up front and guarded with a narrow `except ValueError`; hydration/conversion sit outside the guard. Skips the stack + aggregates one warning, mirroring `resolve_timeline_stacks`. (Self-review follow-up: keeping the guard off the hydration call also matters because pydantic `ValidationError` subclasses `ValueError` — a broad guard would have swallowed a real member validation error.)

routers/api/sync/stream.py
taggartgorman: Fixed in `7bd3e95` — the arch doc "Stack Snapshot (StacksV1)" section is now canonical (and rewritten for the new incremental behavior). The `_stream_stacks` docstring is trimmed to the behavioral contract + two code-local trade-offs + a pointer; the test-module docstring is one sentence + pointer; `immich-sync-communication.md` keeps only the table-level delta (StackV1 upserts, never StackDeleteV1) + pointer.

tests/unit/api/sync/test_sync_stream_stacks.py
taggartgorman: Fixed in `7bd3e95` — deleted `_collect`; all call sites use conftest `collect_stream`.

docs/architecture/sync-stream-architecture.md
taggartgorman: Fixed in `7bd3e95` — set `last-updated: 2026-08-03` on both `sync-stream-architecture.md` and `code-practices.md`.

tests/unit/api/sync/test_sync_stream_stacks.py
taggartgorman: Fixed in `7bd3e95` — `caplog.at_level("INFO", logger="routers.api.sync.stream")`, inline `import logging` removed.

routers/api/sync/stream.py
```diff
@@ -623,6 +681,26 @@ async def generate_sync_stream(
...
+                    yield stack_line
+                    event_counts[SyncEntityType.StackV1.value] = (
+                        event_counts.get(SyncEntityType.StackV1.value, 0) + 1
+                    )
+                    total_events += 1
+            except GumnutError:
+                logger.warning(
+                    "StackV1 snapshot cut short by a Gumnut API error; already "
+                    "streamed rows are acked and the rest resume next sync",
+                    extra={"user_id": owner_id},
```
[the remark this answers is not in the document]
taggartgorman: Verified this is real — the asset pass emits `stackId` unconditionally (`converters.py:234`), so if this pass truncates mid-hydration, assets after the failure point reference stacks the client never received and the timeline join hides those frames.

Skipping it here, though, because the snapshot runs *before* the asset loop: re-raising to leave the stream incomplete suppresses every other entity type too, so one flaky stack read would block all sync progress indefinitely rather than just hiding some burst frames. Round 1 weighed exactly this and deliberately chose partial delivery over a full-sync block (see `stream.py:684` and the `except GumnutError` at `:696`). The third option — nulling out membership in the asset pass for undelivered stacks — is a real fix but needs delivered-stack state threaded across both passes, and this whole pass is pending replacement by API-side stack lifecycle events rather than reshaping. Resolving; will revisit if that direction changes.

routers/api/sync/stream.py
```diff
@@ -503,6 +502,65 @@ def _yield_buffered_deletes(
...
+            continue
+
+        # Keep the ValueError guard separate from hydration failures.
+        try:
+            safe_uuid_from_stack_id(stack.id)
+        except ValueError:
+            undecodable_ids.append(stack.id)
+            continue
+
+        hydrated = await hydrate_stack(gumnut_client, stack)
```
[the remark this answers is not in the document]
taggartgorman: Confirmed — `hydrate_stack` does a full `ASSET_INCLUDE` member read per stack purely to resolve `primary_asset_id`, and this pass awaits them serially, so a large first sync is an N+1 serial chain.

Skipping in this pass. Switching to the bounded-concurrency `hydrate_stacks` helper would hydrate the whole library before yielding anything, which loses the incremental per-stack streaming and monotonic cursor-resume that round 1 deliberately built (a truncated snapshot resumes from the last ack instead of restarting). The over-fetch/serial cost is the same one the sync-load review raises, and its conclusion is that the fix belongs API-side — an `updated_after`/`order` filter on the stacks list plus a lean member read — which removes this serial loop and the client-side sort altogether. Deferring to that direction. Resolving.

PR #345, opened 2026-08-11
Event-driven Immich stack sync; retire the full-table snapshot

taggartgorman:
## Summary

Replace the per-sync full-table stack snapshot with cursor-driven `StacksV1` events. Immich clients now receive stack upserts and deletes as deltas, while synced assets carry their stack membership in both V1 and V2 formats.

## Behavior

- Stream stack upserts before member assets so the mobile timeline never sees an asset reference an unknown stack. Buffer `StackDeleteV1` until after asset deletes and membership-clearing updates.
- Apply the event payload `stack_id` to asset updates, preserving membership at the sync cutoff when an asset moves again during hydration.
- Resolve the required `primaryAssetId` with lean, short-circuit member queries: pinned member, first live member, then first all-state member for an all-trashed stack. Reads are concurrency-bounded and queued siblings are cancelled after a failure.
- Treat an empty all-state member read as retryable even when `asset_count` is zero, because the count excludes trashed members. The stream ends without `SyncCompleteV1`, leaving the cursor unacknowledged for retry.
- Degrade undecodable stack IDs to loose assets without truncating the stream.

The Gumnut API contract supplies stack lifecycle events and emits `asset_updated` events before `stack_deleted` when a stack dissolves.

## Documentation

The sync architecture now records the visibility, retry, primary-selection, cutoff, and dissolve contracts. The active gap analysis marks mobile stack sync closed, and streaming-error guidance is cursor- and dependency-aware.

## Test plan

- `uv run ruff format`
- `uv run ruff check`
- `uv run pyright`
- `uv run pytest` — 1,689 passed
- `uv run scripts/lint_docs.py`

### DIFF
--- a/docs/architecture/sync-stream-architecture.md
+++ b/docs/architecture/sync-stream-architecture.md
@@ -73,56 +73,19 @@ delete from the album-user pass would duplicate `AlbumDeleteV1`. No
 
 ## Stack Snapshot (StacksV1)
 
-Stacks reach the mobile client two ways, both driven from the initial/full sync:
+Each sync asset carries its stack membership in `stackId`. Before streaming
+assets, `_stream_stacks` re-pages the stack table and emits rows after the
+client's checkpoint in `(updated_at, id)` order. This provides incremental
+upserts and resumes a truncated pass without a stack lifecycle event stream.
 
-1. **Asset membership.** Each sync asset row carries its `stackId` — the Gumnut
-   `stack_id` FK mapped to the Immich UUID (loose assets stay null). V2 inherits
-   it through the V1 delegation. This is what groups burst members under a stack
-   on the client.
-2. **Stack rows.** `StacksV1` streams `StackV1` rows, handled specially
-   (`_stream_stacks`) alongside the user types rather than via the events API —
-   the Gumnut backend emits no stack lifecycle events yet, so there is nothing to
-   feed the event stream. Each row's `primaryAssetId` is the **effective primary**
-   resolved by the shared `resolve_effective_primary`/`hydrate_stack` helpers (the
-   exact rule the REST `/stacks` surfaces use), so an auto-detected burst with no
-   pinned cover still gets a valid non-null primary. A member-less stack has no
-   honest primary and is skipped with a warning. `ownerId` is always the current
-   user (Gumnut is single-user; no stack sharing).
+Each row uses the same effective-primary resolution as the REST stack endpoints;
+member-less stacks are skipped because `primaryAssetId` is required. Incremental
+upserts prevent assets from referencing a missing stack, which would hide the
+burst in the mobile timeline.
 
-The stack rows are emitted **before** the phase-1 asset loop, so a stack row
-exists on the client before any asset naming it via `stackId` arrives. (The
-mobile client's `stackId` / `primaryAssetId` columns are unenforced indexes, not
-FKs, so the circular reference is not order-sensitive in practice — stacks-first
-is a deliberate, defensive choice for the asset→stack direction.)
-
-**Scope boundary — upserts only, no delete detection.** With no lifecycle event
-stream, `_stream_stacks` re-pages the whole stack table on **every** sync and
-emits the rows the client has not yet acked: it orders by `(updated_at, id)` and
-skips rows at or before the `StackV1` checkpoint cursor. This is upsert-only
-incremental delivery, and it exists for a concrete reason — asset rows carry
-`stackId` from current entity state unconditionally, so a burst created *after*
-the initial sync would otherwise leave its member assets pointing at a stack row
-the client never received. A dangling `stackId` is not benign: the mobile
-timeline join (`merged_asset.drift.dart`,
-`... AND (rae.stack_id IS NULL OR rae.id = se.primary_asset_id)`) hides **every
-frame** of a burst whose stack row is missing. Emitting the new stack keeps the
-burst visible and grouped.
-
-Ordering by the ack cursor also makes the pass **resumable**: if a snapshot
-truncates mid-stream, the checkpoint holds the last acked row's cursor and the
-next sync skips at-or-before it, so the un-streamed tail is not lost. This is
-also why a Gumnut API error while hydrating one stack is caught at the call site
-and ends the pass gracefully rather than propagating: the snapshot runs *before*
-the asset loop, so an unhandled error would suppress assets/albums/faces too —
-instead the already-streamed rows are acked, the asset loop still runs, and the
-remaining stacks resume next sync.
-
-**Delete detection stays out of scope**, tracked on a separate event-support
-track. A dissolved stack's row goes stale but is harmless: unstacking clears each
-member's `asset.stack_id` via asset events, and a stack row no asset references
-satisfies the `rae.stack_id IS NULL` branch of the join. `PartnerStacksV1` stays
-a no-op (partner sharing is unsupported), and `on_asset_stack_update` is **not**
-in the supported WebSocket table.
+Delete detection remains unsupported. A dissolved stack row is harmless once
+asset events clear its members' `stackId`. `PartnerStacksV1` remains a no-op
+because partner sharing is unsupported.
 
 ## Adding a New Sync Type Version
 
@@ -140,7 +103,10 @@ The invariant tests in `test_sync_v2.py` assert that every V2 type in `_SYNC_TYP
 
 Immich sync types that are accepted but have no Gumnut equivalent (e.g., `AssetEditsV1` — we don't support editing) go in `_NOOP_REQUEST_TYPES` in `stream.py`. This prevents "unsupported type" warnings while making the no-op explicit. Do not just add them to `_SUPPORTED_REQUEST_TYPES` without `_SYNC_TYPE_ORDER` — that silently drops them.
 
-The v3 mobile client requests a broad set of these on **every** sync (partner-*, `PartnerStacksV1`, memories-*, `AssetMetadataV1`, `AssetOcrV1`, `AlbumAssetExifsV1`, the V2 partner/album-asset variants) — all no-ops because the adapter has no sync-stream source for them, whether the underlying feature is absent (sharing, OCR) or reachable only through the REST surfaces (memories). They all belong in `_NOOP_REQUEST_TYPES` so the per-sync "unsupported types" warning stays quiet. Two requested types the adapter *does* handle are deliberately **not** no-ops, both handled specially alongside `UsersV1`/`AuthUsersV1` rather than via `_SYNC_TYPE_ORDER`: `UserMetadataV1` (synthesized preferences row — see "User Preferences" above) and `StacksV1` (current-state snapshot — see "Stack Snapshot" above). `PartnerStacksV1` remains a no-op regardless.
+The v3 mobile client requests these types on every sync, so keep unsupported
+ones in `_NOOP_REQUEST_TYPES` to avoid repeated warnings. `UserMetadataV1` and
+`StacksV1` are handled specially outside `_SYNC_TYPE_ORDER` and must not be
+listed as no-ops.
 
 ## Contract with the Gumnut API
 
--- a/docs/architecture/sync-stream-architecture.md
+++ b/docs/architecture/sync-stream-architecture.md
@@ -71,6 +71,42 @@ pass: an `album_deleted` event emits `AlbumDeleteV1`, and the client's
 delete from the album-user pass would duplicate `AlbumDeleteV1`. No
 `AlbumUserDeleteV1` is emitted (Gumnut has no unshare operation).
 
+## Stack Snapshot (StacksV1)
+
+Stacks reach the mobile client two ways, both driven from the initial/full sync:
+
+1. **Asset membership.** Each sync asset row carries its `stackId` — the Gumnut
+   `stack_id` FK mapped to the Immich UUID (loose assets stay null). V2 inherits
+   it through the V1 delegation. This is what groups burst members under a stack
+   on the client.
+2. **Stack rows.** `StacksV1` streams a **current-state snapshot** of `StackV1`
+   rows, handled specially (`_stream_stacks`) alongside the user types rather
+   than via the events API — the Gumnut backend emits no stack lifecycle events
+   yet, so there is nothing to feed the event stream. Each row's
+   `primaryAssetId` is the **effective primary** resolved by the shared
+   `resolve_effective_primary`/`hydrate_stack` helpers (the exact rule the REST
+   `/stacks` surfaces use), so an auto-detected burst with no pinned cover still
+   gets a valid non-null primary. A member-less stack has no honest primary and
+   is skipped with a warning. `ownerId` is always the current user (Gumnut is
+   single-user; no stack sharing).
+
+The snapshot is emitted **before** the phase-1 asset loop, so a stack row exists
+on the client before any asset naming it via `stackId` arrives. (The mobile
+client's `stackId` / `primaryAssetId` columns are unenforced indexes, not FKs,
+so the circular reference is not order-sensitive in practice — stacks-first is a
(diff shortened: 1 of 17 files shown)


routers/api/sync/entity_fetch.py
```diff
@@ -16,6 +20,45 @@ def _batched(items: list[str], size: int) -> list[list[str]]:
...
+    try:
+        return await hydrate_stack(client, stack_row)
+    except GumnutError:
+        logger.warning(
+            "Stack member read failed during sync; syncing its assets as loose "
+            "and skipping the stack row",
+            extra={"stack_id": stack_row.id},
+            exc_info=True,
+        )
+        return None
```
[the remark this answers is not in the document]
taggartgorman: Good catch — fixed in 88d69bc. You're right that skipping a retriable failure permanently drops the stack: `_stream_entity_type` advances the events cursor past the skipped event, so its `StackV1` never emits while the asset pass still stamps `stackId` on its members, hiding the burst.

The per-stack `GumnutError` catch is removed. Hydration now skips to `missing_ids` only for *permanently* non-emittable stacks that strand nothing downstream (member-less, or an undecodable id whose members degrade to loose too); a transient failure propagates, truncating the sync with the cursor preserved so the stack retries next cycle. A persistent failure then wedges the pass loudly rather than losing data silently — bounded per-stack retry is noted as a future refinement.

routers/api/sync/stream.py
```diff
@@ -72,6 +73,11 @@
 # Order matters - assets before metadata, albums before album_assets, etc.
 # This ordering ensures FK parents are streamed before children during upserts.
 _SYNC_TYPE_ORDER: list[tuple[SyncRequestType, str, SyncEntityType]] = [
+    # Stacks stream before assets: the mobile timeline hides an asset whose
+    # stackId names a stack row the client doesn't have, so the stack must land
+    # before the member asset pointing at it. (The client enforces no FK
+    # constraint either way — this is about visibility, not insert order.)
+    (SyncRequestType.StacksV1, "stack", SyncEntityType.StackV1),
```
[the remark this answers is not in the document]
taggartgorman: Valid mechanism, but this is a pre-existing property of the per-`SyncEntityType` checkpoint model rather than something specific to stacks, so I'm not fixing it in this PR. Every pass resumes from its upsert-type checkpoint (`checkpoint_map[StackV1]`, `[AssetV1]`, …) while deletes ack under the delete type, so a delete-only window replays from the stale upsert cursor for assets, albums, and faces in exactly the same way. Reconciling the upsert/delete checkpoints is a change to the shared cursor machinery affecting every entity type — out of scope here, where stacks deliberately follow the established pattern. It's worth addressing separately given stacks can be more delete-heavy / create-sparse than assets, so I've tracked it as a follow-up.

routers/api/sync/entity_fetch.py
```diff
@@ -101,4 +148,31 @@ async def fetch_entities_map(
...
+            stack_page = await gumnut_client.stacks.list_stacks(
+                ids=chunk, limit=len(chunk)
+            )
+            rows = stack_page.data
+            hydrated_stacks = await gather_with_concurrency(
+                [_hydrate_stack_for_sync(gumnut_client, row) for row in rows]
+            )
+            for stack_row, hydrated in zip(rows, hydrated_stacks, strict=True):
+                if hydrated is None:
+                    missing_ids.add(stack_row.id)
```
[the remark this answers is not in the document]
taggartgorman: Good catch — fixed in 18d4bb5. You're right that the exception fix left the non-exception path open: `hydrate_stack` also returns `None` when `asset_count > 0` but the member read comes back empty, and skipping that would strand the members.

`_hydrate_stack_for_sync` now skips to `missing_ids` only when `asset_count == 0` (genuinely member-less — the `state="all"` read would otherwise see even trashed members). A positive `asset_count` with an empty read raises `StackMemberReadInconsistent`, truncating the sync so the cursor is preserved and the stack retries next cycle — same treatment as a member-read `GumnutError`.

routers/api/sync/entity_fetch.py
```diff
@@ -101,4 +148,31 @@ async def fetch_entities_map(
...
+            # lacking metadata; a transient member-read failure propagates
+            # instead (see _hydrate_stack_for_sync for why skipping would hide
+            # the burst).
+            stack_page = await gumnut_client.stacks.list_stacks(
+                ids=chunk, limit=len(chunk)
+            )
+            rows = stack_page.data
+            hydrated_stacks = await gather_with_concurrency(
+                [_hydrate_stack_for_sync(gumnut_client, row) for row in rows]
+            )
```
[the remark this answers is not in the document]
taggartgorman: Real, but I'm treating it as a scoped follow-up rather than fixing it here. For auto-burst stacks — the dominant case, a handful of frames — the retained data is modest and per-chunk/transient; the pathological case is large user-created stacks. The clean fix is a *lean* member read (`state="all"`, no `ASSET_INCLUDE`) feeding `resolve_effective_primary`, which is a divergent fetch path that still has to preserve the trashed-pin correctness `hydrate_stack` relies on — and this PR was scoped to reuse that helper unchanged. Tracked as a follow-up so it gets its own change + review; the `asset_count`-contradiction handling added here carries into it.

routers/api/sync/entity_fetch.py
```diff
@@ -101,4 +169,31 @@ async def fetch_entities_map(
...
+            # blocking round-trip per stack. A permanently non-emittable stack
+            # (member-less, or undecodable id per _hydrate_stack_for_sync) yields
+            # None and goes to missing_ids — the same inert skip as an asset
+            # lacking metadata; a transient member-read failure propagates
+            # instead (see _hydrate_stack_for_sync for why skipping would hide
+            # the burst).
+            stack_page = await gumnut_client.stacks.list_stacks(
+                ids=chunk, limit=len(chunk)
+            )
+            rows = stack_page.data
```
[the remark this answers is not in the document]
taggartgorman: Confirmed valid — and it's the same "a transiently-missing stack strands its members" class as the two paths fixed here, but this one is the *generic* skip-missing-row behavior every entity type shares (`not_returned` in `_stream_entity_type`): benign for assets (merely absent), data-hiding for stacks.

The robust fix can't naively raise on a missing row: a stack created-and-deleted within one sync window is legitimately absent from `list_stacks` (its create is an upsert, the row is gone), so raising there would wedge the sync forever. Distinguishing a lagging row from a deleted one needs the window's delete events — a change to the shared not-returned/cursor machinery.

Given that, plus the low probability (read-replica lag + a specific page ordering), this was a deliberate decision to land the PR with the two in-scope data-loss bugs fixed and track the shared-machinery fix as a separate follow-up, rather than reshape the generic path mid-review. Not silently accepted — tracked to be fixed.

taggartgorman (commented):
The event-driven approach is a clear improvement over full-table snapshots, and CI is green. I found correctness, performance, coverage, and documentation issues around stack hydration and retry behavior.

The PR description also repeats its explanation across Summary/What/Why and still claims per-stack failure isolation, which contradicts the current propagation behavior. Please consolidate it and remove those stale claims.

Findings: 0 Must Fix, 5 Should Fix, 3 Consider.

routers/api/sync/entity_fetch.py
```diff
@@ -16,6 +20,62 @@ def _batched(items: list[str], size: int) -> list[list[str]]:
...
+        safe_uuid_from_stack_id(stack_row.id)
+    except ValueError:
+        logger.warning(
+            "Undecodable stack id during sync; skipping the stack row and "
+            "syncing its assets as loose",
+            extra={"stack_id": stack_row.id},
+        )
+        return None
+    hydrated = await hydrate_stack(client, stack_row)
+    if hydrated is None and stack_row.asset_count > 0:
```
taggartgorman: **[Should Fix] Don’t treat zero live assets as proof that the stack has no members**

`asset_count` excludes trashed members, so zero can describe an all-trashed stack. If this `state="all"` read is transiently empty, returning `None` advances the event cursor past a stack the client never received. Treat emptiness as retryable unless the API provides a signal proving permanent absence, and add an all-trashed regression case.

routers/api/sync/entity_fetch.py
```diff
@@ -101,4 +169,31 @@ async def fetch_entities_map(
...
+            # (member-less, or undecodable id per _hydrate_stack_for_sync) yields
+            # None and goes to missing_ids — the same inert skip as an asset
+            # lacking metadata; a transient member-read failure propagates
+            # instead (see _hydrate_stack_for_sync for why skipping would hide
+            # the burst).
+            stack_page = await gumnut_client.stacks.list_stacks(
+                ids=chunk, limit=len(chunk)
+            )
+            rows = stack_page.data
+            hydrated_stacks = await gather_with_concurrency(
```
taggartgorman: **[Should Fix] Avoid full asset hydration when sync only needs the primary ID**

`hydrate_stack` fetches every member with metadata, people, and file-data fields, but this path retains only `primary_asset_id`. Initial sync can therefore perform many unnecessarily heavy member reads. Use a sync-specific lean fetch, or parameterize hydration to omit fields unused by primary selection.

tests/unit/api/sync/test_sync_stream_stacks.py
```diff
@@ -0,0 +1,526 @@
...
+        assert missing == set()
+        fetched_a = result[stack_a.id]
+        fetched_b = result[stack_b.id]
+        assert isinstance(fetched_a, FetchedStack)
+        assert isinstance(fetched_b, FetchedStack)
+        assert fetched_a.primary_asset_id == safe_uuid_from_asset_id(members_a[0].id)
+        assert fetched_b.primary_asset_id == safe_uuid_from_asset_id(members_b[2].id)
+
+    @pytest.mark.anyio
+    async def test_transient_member_read_failure_propagates_not_skipped(self):
```
taggartgorman, line 263: **[Should Fix] Test retry behavior at the stream boundary**

This proves only that `fetch_entities_map` raises. Add a `generate_sync_stream` regression test with stack and asset passes, asserting that hydration failure prevents later asset events and `SyncCompleteV1`. That is the behavior preserving the unacknowledged cursor for retry.

docs/references/code-practices.md
```diff
@@ -259,6 +259,14 @@ for asset_uuid in request.ids:
 
 Use `map_gumnut_error(e, context, extra=..., exc_info=True)` only when the call site needs to enrich the upstream log record with context the global handler can't see — most commonly the upload paths logging filename / device ids / tracebacks.
 
+**Streaming responses.** Once a `StreamingResponse` commits its headers,
```
taggartgorman, line 262: **[Should Fix] Make this guidance cursor- and dependency-aware**

The blanket instruction to handle expected per-item failures contradicts this PR’s required behavior: transient stack failures must propagate because skipping advances the cursor past referenced data. State that per-item degradation is safe only when skipping cannot strand downstream references; otherwise propagate for retry. Keep the detailed stack rule in the architecture document.

routers/api/sync/stream.py
```diff
@@ -122,6 +128,9 @@
     SyncEntityType.PersonDeleteV1,
     SyncEntityType.AlbumDeleteV1,
     SyncEntityType.AssetDeleteV1,
+    # After asset deletes: a stack is the FK parent of its member assets, so the
```
taggartgorman: **[Should Fix] Describe the actual ordering constraint**

The architecture document correctly says Immich enforces no stack FK, so calling the stack an ‘FK parent’ is misleading.

```suggestion
# Delete member assets before their stack; dissolve updates ran in phase 1.
SyncEntityType.StackDeleteV1,
```

routers/api/sync/entity_fetch.py
```diff
@@ -16,6 +20,62 @@ def _batched(items: list[str], size: int) -> list[list[str]]:
...
+    A transient contradiction (e.g. read replication lag), not a member-less
+    stack — surfaced so the sync truncates and retries rather than skipping the
+    stack and stranding its members as a hidden burst.
+    """
+
+
+async def _hydrate_stack_for_sync(
+    client: AsyncGumnut, stack_row: StackListStacksResponse
+) -> HydratedStack | None:
+    """Hydrate one stack, skipping only *permanent* non-emittable conditions.
```
taggartgorman: **[Consider] Consolidate the retry rationale**

This invariant is repeated in the architecture doc, this docstring, the fetch branch, `FetchedStack`, tests, and the PR body. Keep the durable explanation in the architecture doc and reduce this to the return-versus-propagate contract. Retain one local note about catching UUID decoding narrowly because `ValidationError` is also a `ValueError`; remove the road-not-taken discussion.

docs/architecture/sync-stream-architecture.md
```diff
@@ -71,6 +73,40 @@ pass: an `album_deleted` event emits `AlbumDeleteV1`, and the client's
 delete from the album-user pass would duplicate `AlbumDeleteV1`. No
 `AlbumUserDeleteV1` is emitted (Gumnut has no unshare operation).
 
+## Stacks (StacksV1)
+
+Stacks ride the same event-cursor path as every other entity type. `StacksV1`
```
taggartgorman: **[Consider] Condense this section**

The section can preserve its contracts in roughly half the space: event types and visibility ordering; primary selection and bounded hydration; then the Gumnut API’s dissolve-ordering contract. Details about converter purity, `FetchedStack`, manufactured covers, and repeated timeline consequences belong with their owning symbols or can be removed.

tests/unit/api/sync/test_sync_stream_stacks.py
```diff
@@ -0,0 +1,526 @@
+"""Tests for stacks in the event-driven sync stream.
```
taggartgorman: **[Consider] Trim test narration and duplicate cases**

Make the module/helper docstrings one line, delete comments that repeat test names or assertions, and parameterize the duplicate mapped/null V1/V2 converter tests. Keep short comments for the non-obvious regression mechanisms: crossed covers, cursor advancement, inconsistent reads, and prefix-drift blast radius.

routers/api/sync/entity_fetch.py
taggartgorman: Fixed — sync resolves only the primary UUID through short-circuit lean queries with no metadata, people, or file-data includes.

docs/references/code-practices.md
taggartgorman: Fixed — streaming guidance now allows degradation only when cursor advancement cannot strand dependent data; otherwise failures propagate for retry.

routers/api/sync/entity_fetch.py
taggartgorman: Fixed — reduced the helper and carrier documentation to local contracts and kept the durable retry rationale in the architecture doc.

docs/architecture/sync-stream-architecture.md
taggartgorman: Fixed — condensed the stack architecture section around visibility ordering, lean primary resolution, retry behavior, and dissolve ordering.

routers/api/sync/stream.py
taggartgorman: Fixed — the comment now describes member-asset deletion and phase-1 dissolve updates without claiming a client FK.

tests/unit/api/sync/test_sync_stream_stacks.py
taggartgorman: Fixed — shortened module/helper docstrings, removed assertion-narrating comments, and parameterized the duplicate V1/V2 converter cases while retaining regression rationale.

tests/unit/api/sync/test_sync_stream_stacks.py
taggartgorman: Fixed — added a generate_sync_stream boundary test proving stack hydration failure prevents later AssetV2 and SyncCompleteV1 events.

ternarybits (approved):

PR #348, opened 2026-08-12
Propagate stack membership to the upload-ready payload

taggartgorman:
## What

- Move the private `_immich_stack_id` mapping helper out of the sync package into the shared `routers/utils/gumnut_id_conversion.py` as `immich_stack_id`, and use it on **both** the sync asset converter and the upload-ready WebSocket payload (`build_asset_upload_ready_payload`), which previously hardcoded `stackId=None`. An undecodable stack ID degrades to a loose asset (`stackId=None`) rather than raising.
- Add regression tests for the incremental asset-membership path: a move between stacks emits one asset upsert carrying the new `stackId`; a mid-window dissolve delivers the freed member with `stackId=None` before `StackDeleteV1`; the upload-ready payload maps / loose-nulls `stackId`; plus a direct unit test for the shared helper.
- Document the asset-membership mapping and note why stacks are deliberately kept out of per-event FK verification.

## Why

The event-driven Immich stack sync (#345) mapped `stackId` in the sync converter and established the two-phase dissolve ordering, but left the upload-ready WebSocket payload silently dropping stack membership and the mapping helper private to the sync package. This finishes those loose ends and locks the behavior in with tests.

Stacks stay out of `FK_REFERENCES`: the Immich mobile client has no `stackId` foreign key (the column is indexed only), so a transient reference to an already-removed stack is a self-healing timeline-visibility gap — and the dissolve ordering already clears each member before the stack delete — not the unacknowledgeable insert failure that per-event FK verification exists to prevent. The one non-obvious test trap this surfaced (a `Mock` fixture fails the `isinstance` guard the override sits behind, so it must be a real SDK response) is recorded in the testing reference.

## Test plan

- `uv run pytest` — full suite green (1697 passed), including the new stack-membership, dissolve-ordering, and upload-ready tests.
- `uv run ruff format` / `uv run ruff check` / `uv run pyright` — clean.


taggartgorman (commented):
## Review: Approve ✅ (posted as a comment — GitHub blocks approving your own PR)

Clean, well-scoped change. It extracts the previously-private `_immich_stack_id` helper into the shared `routers/utils/gumnut_id_conversion.py` as public `immich_stack_id`, and wires it into `build_asset_upload_ready_payload`, which previously hardcoded `stackId=None` — so the upload-ready WebSocket payload now reflects burst-stack membership identically to the sync converter. Behavior on the sync path is byte-for-byte preserved.

Reviewed across correctness, data integrity, security, silent failures, performance, test coverage, clarity, and minimalism (3 parallel reviewers + orchestrator). Findings: **0 Must Fix / 0 Should Fix / 1 Consider** (non-blocking, inline).

### Verified
- **`stack_id` is always available at both call sites.** `AssetResponse.stack_id` is a top-level scalar field (not an `include`-gated heavy field), so both `build_asset_upload_ready_payload` callers populate it. No “reports loose when actually stacked” hazard; at fresh-upload time it is `None` (bursts are detected afterward), and the populated case is exactly the re-emit/refresh path this PR fixes.
- **Behavior preserved on the sync path** — `immich_stack_id` is a byte-for-byte extraction; `import logging` / `logger` / the old private helper are fully removed with no dangling references, and `safe_uuid_from_stack_id` is still used by `gumnut_stack_to_sync_stack_v1`.
- **Degrade-to-loose on an undecodable ID** is the pre-existing, faithfully-moved posture — raising would drop the whole asset’s realtime event over a display-only grouping, so degrading is the better trade-off; it mirrors `resolve_timeline_stacks`’s documented fallback and logs the offending id.
- **Tests are non-vacuous** — the dissolve test genuinely exercises the `isinstance`-gated event-time override (real `AssetResponse`, payload clears `stack_id` to `None`); all three `immich_stack_id` branches and both payload cases (stacked / loose) are covered.
- **Single-user model** — no security / cross-tenant concern; **CI green** (lint, type, 1697 tests, docs, compatibility).

## Pre-approve checklist
- [x] **Findings**: 0 Must Fix / 0 Should Fix (1 Consider, inline, non-blocking)
- [x] **Codebase value**: Removes a latent inconsistency (the upload-ready payload silently dropped stack membership) at the smallest scope — extracts an already-existing helper and calls it in the second site, making both call sites identical. Beats “do nothing” (fixes the dropped-membership bug) and a narrower fix (invents no new abstraction).
- [x] **Security/auth**: No authentication, authorization, or permission logic touched (ID conversion + payload construction only). Checked `gumnut_id_conversion.py`, `asset_conversion.py`, `converters.py`; no permission/token surfaces involved.
- [x] **Data deletion**: No destructive operations. The dissolve *test* asserts stack-delete ordering, but the change itself adds no deletion code.
- [x] **Confidence**: Correctness well understood — a byte-for-byte helper extraction plus one hardcoded `None` replaced, covered by new and existing tests, CI green.
- [x] **Purpose clarity**: Purpose and approach are clear from the description and the code.

docs/architecture/sync-stream-architecture.md
```diff
@@ -93,12 +93,21 @@ converter. Reads are concurrency-bounded. An empty all-state member read is
...
+
 Stack references stay out of `FK_REFERENCES`. When a stack dissolves, the
 Gumnut API emits an `asset_updated` clearing each member's `stack_id` before
 `stack_deleted`. Those upserts run in phase 1 and the stack delete in phase 2,
 so members stop referencing the stack before the client removes it. Asset
 updates use the payload's event-time `stack_id`, preventing a later move outside
-the sync window from leaking into the current event.
+the sync window from leaking into the current event. Unlike the face→person and
+album→cover payload references, stacks are deliberately not verified per event:
+the client has no `stackId` foreign key (the column is indexed only), so an
```
taggartgorman: **[Consider]** The “no client `stackId` FK ⇒ visibility, not integrity” premise is now derived twice in this section. Line 86 already establishes it for emit ordering (“The client has no stack foreign key; this is a visibility constraint.”), and these added lines re-derive it from scratch for the FK_REFERENCES decision. Both are correct and each supports a distinct decision, so this is entirely optional — if you want to trim, you could let one occurrence carry the canonical statement (keeping the new “the column is indexed only” precision). Fine to leave as-is.

taggartgorman: Done in de19cb4 — consolidated the "no client `stackId` FK" fact onto the emit-ordering line (kept the "indexed only" precision there) and made the `FK_REFERENCES` paragraph back-reference it instead of re-deriving. Thanks!

ternarybits (approved):

PR #349, opened 2026-08-12
Emit on_asset_stack_update on stack mutations

taggartgorman:
## What

- Add `STACK_UPDATE = "on_asset_stack_update"` to `WebSocketEvent` and emit it (payload `None`) from the five mutating stack routes: `create_stack`, `update_stack` (cover-change branch only), `remove_asset_from_stack`, `delete_stack`, and `delete_stacks` (one emit per clean bulk dissolve). Each is scoped to the owning user's room, mirroring upstream Immich's emit sites in `stack.service.ts`.
- Add a lightweight `get_current_user_id` dependency to the three delete/remove routes so they can resolve the room key.
- Update `docs/references/websocket-events-reference.md` and add emit tests in `tests/unit/api/test_stacks.py`.

## Why

E4 moved stack changes onto the event-driven sync stream (the pull path). This is the push half: when one client mutates a stack, the adapter now sends the same realtime hint Immich's own server does, so it behaves like upstream rather than silently dropping the event.

**Payload is empty, deliberately.** Upstream calls `clientSend('on_asset_stack_update', userId)` with no data argument — its repository type `on_asset_stack_update: string[]` is a rest-param that spreads to zero args, not an id array. So the adapter emits with `payload=None` via `emit_user_event`, never `emit_user_event_per_id`.

**No current Immich client subscribes to this event.** The web client declares `on_asset_stack_update` in its `Events` type map but registers no `.on()` handler for it; the mobile client has zero references. So this emission is upstream-parity / forward-compat, not an immediate UI trigger — the durable convergence path for stack changes remains the event-driven sync stream. This is why the reference doc's prior "the adapter omits it because no handler subscribes" note is replaced rather than kept.

**Scope notes:**
- No push-side trigger exists outside these HTTP routes (there is no add-asset-to-existing-stack route, and the sync stream is pull-based), so no polling loop is invented — emission rides the existing stack CRUD routes.
- The cover-less PUT path is a pure read and stays silent, diverging from upstream's unconditional `StackUpdate` emit — signalling "updated" when nothing changed would be misleading noise.
- A partial bulk dissolve raises before emitting; the client reconciles through the sync stream.
- `emit_user_event` already swallows `SocketIOError`, so a transport failure can't fail the originating mutation (no new try/except at call sites).

## Test plan

- `uv run pytest` — full suite green (1702 passed). New tests in `tests/unit/api/test_stacks.py` cover: emit on create / cover change / single dissolve / clean bulk dissolve / member removal; silence on the cover-less read, a partial-failure batch, and an empty batch; owner-scoping of the room id; and the `get_current_user_id` dependency wiring on the three delete/remove routes.
- `uv run ruff check`, `uv run ruff format --check`, `uv run pyright` — clean.

### DIFF
--- a/docs/references/websocket-events-reference.md
+++ b/docs/references/websocket-events-reference.md
@@ -1,6 +1,6 @@
 ---
 title: "Immich WebSocket Events Reference"
-last-updated: 2026-07-31
+last-updated: 2026-08-12
 ---
 
 # Immich WebSocket Events Reference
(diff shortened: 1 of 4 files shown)


taggartgorman (commented):
Reviewed — no blocking findings. Clean, well-scoped push-side counterpart to the stack sync work: emits `on_asset_stack_update` from the five mutating stack routes with an empty payload, faithfully mirroring upstream's zero-data-arg `clientSend`.

Verified:
- Room scoping (`str(user_id)` normalization) matches the `connect()` room key and every existing emit site (`assets.py`, `trash.py`).
- Emit ordering is post-mutation; partial bulk dissolves raise before emitting, so no hint is sent for a half-applied batch.
- `emit_user_event` swallows `SocketIOError`, so a WS failure can't fail the originating mutation — no new error handling needed at call sites.
- Read paths (cover-less PUT, empty/partial bulk) correctly stay silent; the reference doc and architecture doc are consistent with the new behavior.
- Test coverage pins emit-once, owner-scoping, read-path silence, and the new dependency wiring.

Non-blocking note: the three delete/remove routes now resolve `get_current_user_id`, which triggers one cached `users.me()` call — expected, matching the `trash.py` restore-route convention.

Note CI's `Lint and type check` / `Test` were still pending at review time; the PR body reports the full suite green locally.

ternarybits (approved):

PR #350, opened 2026-08-12
Retry a stack row missing from the sync bulk read unless a delete explains it

taggartgorman:
## What

- Add a guard to the sync stream's stack pass: a stack upsert whose row is absent from the bulk `list_stacks` read now truncates the stream (cursor preserved, next sync retries) unless a `stack_deleted` event in the sync window explains the absence — checked against the current events page first, then a look-ahead scan of the remaining stack events, seeded by the triggering page's `has_more`.
- Exclude undecodable-id rows from the guard: the read returned them, and they are deliberately degraded rather than evidence of read lag.
- Seven tests cover the guard's branches (no-delete truncation with scan-skip pin, same-page delete, later-page delete, mixed explained/unexplained, second-scan-page delete pinning cursor chaining, empty-page-despite-has_more, undecodable exclusion); the sync-stream architecture doc gains the matching paragraph.

## Why

A transiently-invisible row (e.g. read lag between the events endpoint and `list_stacks`) was previously treated as permanently missing and skipped, advancing the events cursor past it — while the asset pass still stamped `stackId` on the stack's members. The Immich mobile timeline hides any asset whose `stackId` names no stack row, so the whole burst disappeared with no self-heal until some later stack event re-emitted it. This was the third variant of the "transiently-unavailable stack strands its members" class surfaced in the review of #345; the two row-present variants were fixed there, and this row-absent one was deferred because a naive "raise on any missing row" would wedge forever on a stack created and deleted within one sync window — the delete-event excusal here is what makes the raise safe. Both transient shapes self-heal on retry: a lagging row becomes visible, and a stack deleted after the window bound gets its delete event inside the next window.

## Test plan

- `uv run pytest` — full suite passes (1712 tests), including 36 in `tests/unit/api/sync/test_sync_stream_stacks.py`
- `uv run ruff check`, `uv run ruff format --check`, `uv run pyright`, `uv run scripts/lint_docs.py` — all clean

### DIFF
--- a/routers/api/sync/stream.py
+++ b/routers/api/sync/stream.py
@@ -207,12 +207,16 @@ _SUPPORTED_REQUEST_TYPES: frozenset[SyncRequestType] = frozenset(
 
 async def _require_missing_stacks_deleted(
     gumnut_client: AsyncGumnut,
-    missing_ids: set[str],
+    absent_ids: set[str],
     page_events: Sequence[EventData],
     page_has_more: bool,
     sync_started_at: datetime,
 ) -> None:
-    """Raise unless every missing stack row is explained by an in-window delete.
+    """Raise unless every absent stack row is explained by an in-window delete.
+
+    ``absent_ids`` are ids the bulk read did not return at all — distinct from
+    ``fetch_entities_map``'s ``missing_ids``, which the read returned but
+    deliberately degraded.
 
     A stack **created and deleted within one sync window** is legitimately
     absent from ``list_stacks`` — its create is an upsert event, but the row is
@@ -228,14 +232,16 @@ async def _require_missing_stacks_deleted(
     for why the transient shapes self-heal and a persistent failure wedges the
     pass, deliberately).
     """
-    unexplained = missing_ids - {
+    unexplained = absent_ids - {
         event.entity_id for event in page_events if event.event_type == "stack_deleted"
     }
     last_cursor = page_events[-1].cursor
     has_more = page_has_more
     while unexplained and has_more:
-        # Params dict matches _stream_entity_type's call shape, including the
-        # plain-string entity_types the SDK accepts as a comma-delimited value.
+        # The dict is deliberate: the SDK types entity_types as a string
+        # sequence, but the API also accepts the plain comma-delimited string
+        # _stream_entity_type sends — the untyped dict is what lets the same
+        # wire shape pass the type checker.
         params: dict[str, Any] = {
             "created_at_lt": sync_started_at,
             "entity_types": "stack",
--- a/routers/api/sync/stream.py
+++ b/routers/api/sync/stream.py
@@ -239,7 +239,7 @@ async def _require_missing_stacks_deleted(
     has_more = page_has_more
     while unexplained and has_more:
         # The dict is deliberate: the SDK types entity_types as a string
-        # sequence, but the API also accepts the plain comma-delimited string
+        # sequence, but the API also accepts the plain string
         # _stream_entity_type sends — the untyped dict is what lets the same
         # wire shape pass the type checker.
         params: dict[str, Any] = {
--- a/routers/api/sync/stream.py
+++ b/routers/api/sync/stream.py
@@ -66,6 +66,13 @@ class StackRowReadIncomplete(Exception):
     members, and the mobile timeline hides any asset whose ``stackId`` names no
     stack row — a hidden burst that would not self-heal until some later stack
     event re-emitted the row.
+
+    Both transient shapes resolve on the retry: a lagging row becomes visible,
+    and a stack deleted *after* the window bound gets its delete event inside
+    the next window. A stack that fails *persistently* wedges the pass until it
+    recovers — loud in the logs, and the same trade-off the member-read
+    failures in ``entity_fetch`` already make; never stranding a member is the
+    priority.
     """
 
 
@@ -202,6 +209,7 @@ async def _require_missing_stacks_deleted(
     gumnut_client: AsyncGumnut,
     missing_ids: set[str],
     page_events: Sequence[EventData],
+    page_has_more: bool,
     sync_started_at: datetime,
 ) -> None:
     """Raise unless every missing stack row is explained by an in-window delete.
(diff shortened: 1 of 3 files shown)


taggartgorman (commented):
Reviewed via `/review-pr` — three parallel dimension reviewers (correctness/data-integrity/security/silent-failures; performance/test-coverage; clarity/minimalism) plus an orchestrator Value Check. This is a COMMENT review rather than an approval only because GitHub won't accept a self-approval on your own PR; there are **no blocking findings**.

**Verdict: 0 Must Fix · 0 Should Fix · 3 Consider (all non-blocking, inline below).**

Correctness is well-confirmed with concrete traces: the cursor never advances past an unexplained missing row (the raise fires before any current-page stack event is yielded and before the asset pass runs, so no member is ever stamped with a dangling `stackId`); every in-window `stack_deleted` is reachable by the scan (a delete is always chronologically after the create, and the create is on the current page); the scan terminates (window bound + `has_more` + empty-page `break` + monotonically shrinking `unexplained`); and `not_returned - missing_ids` correctly isolates truly-absent rows from deliberately-degraded undecodable ones. The change extends the existing `StackMemberReadInconsistent` truncate-and-retry pattern rather than inventing new machinery, and the Value Check passes — it fixes a real, demonstrated bug with the minimal addition (the delete-excusal) that keeps the raise from wedging on a created-and-deleted-in-window stack.

Filtered out as low-signal: an O(P²) event-page refetch tail (only under contrived heavy dissolve churn, and already framed as acceptable in the docstring); a multi-page-truncation edge test (the raise branch is already covered by six tests, and the delta is guaranteed by generator semantics); and the guard being entered with an empty absent-set in the undecodable-only path (a deliberately safe no-op).

routers/api/sync/stream.py
```diff
@@ -182,6 +205,67 @@
...
+    unexplained = absent_ids - {
+        event.entity_id for event in page_events if event.event_type == "stack_deleted"
+    }
+    last_cursor = page_events[-1].cursor
+    has_more = page_has_more
+    while unexplained and has_more:
+        # The dict is deliberate: the SDK types entity_types as a string
+        # sequence, but the API also accepts the plain comma-delimited string
+        # _stream_entity_type sends — the untyped dict is what lets the same
+        # wire shape pass the type checker.
```
taggartgorman, line 244: **[Consider]** Wording: `_stream_entity_type` (and this guard) only ever send a single token (`"stack"`/`"asset"`), never a comma-delimited list — so "comma-delimited string" points a reader at a wire shape that isn't actually produced here. The type-checker rationale itself is correct (`entity_types` is typed `Optional[SequenceNotStr[str]]`, which rejects a bare `str`, so the untyped `dict` is genuinely needed). Just drop "comma-delimited":

```suggestion
        # The dict is deliberate: the SDK types entity_types as a string
        # sequence, but the API also accepts the plain string
        # _stream_entity_type sends — the untyped dict is what lets the same
        # wire shape pass the type checker.
```

routers/api/sync/stream.py
```diff
@@ -262,6 +346,18 @@ async def _stream_entity_type(
         # Track entity IDs that were requested but not returned (deleted/404)
         not_returned = set(upsert_ids) - entities_map.keys()
         if not_returned:
+            # A missing stack row is only safely skippable when a delete
+            # explains it — see StackRowReadIncomplete. Rows the fetch returned
+            # but deliberately degraded (undecodable id, in missing_ids) are
+            # excluded: they are not evidence of read lag.
+            if gumnut_entity_type == "stack":
```
taggartgorman, line 353: **[Consider]** (verification, non-blocking) Because stacks are first in `_SYNC_TYPE_ORDER` and the raise aborts the whole generator, a *persistently* absent stack row — a genuine inconsistency with no in-window delete, recurring every cycle — freezes the **entire** sync (assets, albums, people, faces), not just the stacks pass. That matches the "never stranding a member is the priority" intent and the pre-existing `StackMemberReadInconsistent` blast radius, so it's very likely what you want. Worth a conscious confirmation — and note there's currently no signal that separates a one-cycle self-heal from a stuck-forever wedge (only repeated identical ERROR logs); a consecutive-failure counter could later distinguish them if that ever matters.

routers/api/sync/stream.py
```diff
@@ -182,6 +205,67 @@
...
+        if not events:
+            break
+        unexplained -= {
+            event.entity_id for event in events if event.event_type == "stack_deleted"
+        }
+        last_cursor = events[-1].cursor
+        has_more = events_response.has_more
+
+    if unexplained:
+        raise StackRowReadIncomplete(
```
taggartgorman, line 262: **[Consider]** (verification, non-blocking) This expected, self-healing transient (read lag between the events endpoint and `list_stacks`) flows into `generate_sync_stream`'s broad `except Exception`, which logs `logger.error(..., exc_info=True)` — so in Sentry a routine read-lag retry looks identical to a genuine generator failure. That's consistent with the existing transient-truncation logging and the docstring's deliberate "loud in the logs," so it may be exactly as intended. If the Sentry signal matters, a distinct `except StackRowReadIncomplete` logging at WARNING would stay loud without emitting a traceback-bearing ERROR.

routers/api/sync/stream.py
taggartgorman: Fixed in a71b6e8 — dropped "comma-delimited"; only the single token `"stack"` is ever sent. Grepped the sync router, tests, and architecture doc for other instances of the phrase — none.

routers/api/sync/stream.py
taggartgorman: Confirmed intended: with an unexplained absent row, any progress would stamp members with a dangling `stackId`, so freezing the whole sync until the read heals (or the delete event arrives) is the chosen trade-off — same blast radius as the pre-existing `StackMemberReadInconsistent`, and documented in the `StackRowReadIncomplete` docstring. Skipping the consecutive-failure counter as speculative until a stuck-forever wedge is actually observed.

routers/api/sync/stream.py
taggartgorman: Keeping ERROR for consistency: the sibling transient (`StackMemberReadInconsistent`) flows through the same top-level handler at ERROR, and the docstring deliberately chose loud logging. A WARNING carve-out for only one of the two retriable truncations would diverge from that pattern; if routine read-lag ever produces real Sentry noise, both should get the distinct-handler treatment together.

ternarybits (approved):

PR #361, opened 2026-08-19
Map isEdited from asset kind and honor the edited download selector

taggartgorman:
## Summary (plain English)

Gumnut lets photo-editing tools save an edited version of a picture on top of the original you uploaded — without destroying that original. This adapter translates between Immich apps (the photo apps many people use) and the Gumnut service, and until now it always told those apps "this photo has never been edited" and always handed them whatever the service considered the current picture.

This change fixes both halves of that story:

- **Immich apps now see which photos are edited.** Everywhere the adapter describes a photo — browsing, syncing to your phone, live updates — it now reports the photo as edited whenever the current picture isn't the original upload.
- **Downloads let you choose which one you get.** Asking for the edited picture gives you what the photo currently looks like. Asking for the original (which is what apps do by default) gives you back the exact file you uploaded, byte for byte — important because apps compare file fingerprints to decide whether a photo is already backed up. If the service's records about a photo's versions ever look inconsistent, the download fails with a clear error rather than quietly handing over the wrong file.
- **Interrupted downloads can resume.** The adapter now passes along the "give me just this part of the file" request that apps send when resuming a partial download.

For photos that were never edited, nothing changes — both choices return the same file. Bulk ZIP downloads still always contain the current picture; giving them the same original-vs-edited choice is left for a follow-up.

## What

- Derive Immich's `isEdited` flag from the Gumnut asset's top-level `kind` through a shared predicate (`kind != "original"`) on REST, WebSocket, and sync responses.
- Add Immich's `edited` selector to `GET /api/assets/{id}/original`: `edited=true` streams the current rendering, while the default `edited=false` streams the exact uploaded bytes from the unique position-zero version.
- Forward Range and content-disposition headers on both paths. Invalid version roots fail with 502; a missing original variant returns 404.
- Document the cross-file contract once and keep production comments, docstrings, fixtures, and tests focused on local constraints.

Batch archives deliberately continue to stream the current rendering regardless of `edited`; exact-original selection there needs a separate fan-out design.

## Why

Edits created by Gumnut clients can make an asset's current rendering differ from its uploaded bytes. Immich clients need both accurate edited state and a byte-exact original for checksum reconciliation. The top-level `kind` supplies edited state without another query, while the version-chain root is the stable source of the uploaded bytes.

## Test plan

- `uv run pytest` — 1,781 tests passed.
- `uv run ruff check` — passed.
- `uv run ruff format` — clean.
- `uv run pyright` — passed.
- `uv run scripts/lint_docs.py` — passed.


routers/api/assets.py
```diff
@@ -1262,22 +1326,39 @@ async def view_asset(
 )
 async def download_asset(
     id: UUID,
+    request: Request,
+    edited: bool = Query(default=False, description="Return edited asset if available"),
```
taggartgorman: **[Should Fix]** Exercise the omitted `edited` selector through the handler

This test only inspects `Query.default`; it never exercises the default branch. Direct handler calls that omit this argument receive the truthy `Query` object and therefore select the current rendering, so the test can pass without proving the claimed exact-original default.

Please use `edited: Annotated[bool, Query(description="Return edited asset if available")] = False`, then test a call that omits `edited` and assert it lists versions rather than retrieving the current rendering.

tests/unit/api/test_assets.py
```diff
@@ -3272,18 +3272,76 @@ async def test_view_asset_wide_landscape_preview_request_unaffected(
         )
 
 
+def _make_mock_version(
+    position: int,
+    *,
+    kind: str = "original",
```
taggartgorman: **[Consider]** Remove the unused version `kind` fixture field

`_stream_exact_original` selects the root exclusively through `position == 0`; it never reads the version's `kind`. Keeping this parameter and passing `kind="edit"` implies that kind participates in selection even though it cannot affect an assertion.

Consider removing the parameter, assignment, and explicit `kind="edit"` arguments so the fixture expresses the actual contract.

taggartgorman (commented):
The implementation matches the verified Immich v3.1.0 and Gumnut API contracts: edited state is consistently derived across emit surfaces, while exact-original downloads fail closed on invalid version chains. I found one testability/convention gap in the default selector path and one minor fixture cleanup.

taggartgorman:
Addressed this round's review feedback (both items auto-fixed, two-way doors):

- **Exercise the omitted `edited` selector through the handler** (`routers/api/assets.py`) — Fixed. The param now uses the `Annotated[bool, Query(...)] = False` shape per the repo's route-parameter convention, and the signature-introspection test was replaced with a handler-level test proving an omitted selector lists versions (exact-original path) and never retrieves the current rendering.
- **Remove the unused version `kind` fixture field** (`tests/unit/api/test_assets.py`) — Fixed. `_make_mock_version` no longer takes `kind`; root selection reads only `position`, and the fixture docstring now says so.

Verification: 1530 tests, ruff, and pyright all pass.

routers/api/assets.py
taggartgorman: Fixed — the param now uses the Annotated[bool, Query(...)] = False shape, and the default is exercised through the handler: a call omitting edited asserts versions are listed and the current rendering is never retrieved.

tests/unit/api/test_assets.py
taggartgorman: Fixed — removed the kind parameter, assignment, and kind="edit" call args; the fixture docstring now states that root selection reads only position.

taggartgorman:
@codex review

routers/api/assets.py
```diff
@@ -219,6 +219,70 @@ async def _retrieve_and_stream_variant(
...
+    The Gumnut API guarantees every asset's chain contains exactly one root.
+    If the response violates that (no rows, no root, or duplicate roots), fail
+    closed with a 502 rather than silently substituting another rendering —
+    clients use this route for byte-exact checksum reconciliation, and a wrong
+    variant looks valid but never matches.
+    """
+    gumnut_asset_id = uuid_to_gumnut_asset_id(asset_uuid)
+    versions = await client.assets.versions.list(gumnut_asset_id, include=["variants"])
+
+    roots = [version for version in versions if version.position == 0]
```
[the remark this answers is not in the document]
taggartgorman: Skipping — the premise does not hold: `assets.versions.list` is not paginated. The SDK method returns `VersionListResponse`, a plain `List[VersionListResponseItem]` (not a cursor page/paginator); it accepts no limit or cursor parameters, and its docstring states the contract explicitly: "Returns every retained rendering of one asset, ordered by `position` ascending … Not paginated." A single await therefore materializes the complete chain, so the position-0 root can never sit on a later page and the 502 fail-closed path only fires on a genuinely malformed chain. Iterating with `async for` over a plain list would imply pages that cannot exist. If the Gumnut API ever paginates this endpoint, that lands as an SDK-visible contract change and is the right moment to revisit.

ternarybits (approved):

PR #362, opened 2026-08-19
Add a pure Immich edit-recipe codec

taggartgorman:
## Summary

When you crop, rotate, or mirror a photo in Immich, the app sends the server a list of editing steps and expects the server to apply them. Gumnut instead stores a single compact "recipe" alongside the photo that describes the finished result — for example, "crop to this rectangle, rotate 90°, then mirror." This PR adds the translator between those two formats: it turns any valid Immich edit list into that one normalized recipe, and turns a stored recipe back into the edit list the Immich editor expects to see when the photo is reopened. It handles the tricky part — different orders of rotations and mirrors that end up looking the same are recognized and reduced to one canonical form — and it strictly rejects anything malformed, with clear error codes. Nothing user-facing changes yet: this is groundwork logic plus a large test suite, and the features that will use it (actually rendering the edited photo, and the editing endpoints) come in follow-up work.

## What

- New `routers/utils/asset_edit_conversion.py`: a pure, deterministic codec between Immich's ordered crop/rotate/mirror edit-action lists and the consolidated v1 edit recipe stored as an edit version's `params` (`{"version": 1, "crop"?, "angle", "mirror"}`), plus adapter-domain exceptions (`AssetEditValidationError`, `UnsupportedEditRecipeError`) with stable machine-readable codes.
- Reverse translation synthesizes canonical Immich operation rows (`crop`, `mirror`, `rotate`) with deterministic IDs keyed on asset/version identity plus action kind, so repeated reads return identical rows. IDs are SHA-256-derived with RFC 4122 version-4 bits stamped: upstream Immich validates edit-row IDs as UUIDv4 specifically, and adapter-generated IDs are v4 everywhere else in this repo, so name-based uuid5 values would fail strict clients. Golden-ID and version-bit tests pin the derivation across deployments.
- New `tests/unit/utils/test_asset_edit_conversion.py`: 147 table-driven tests covering all eight composed orientation states, round-trips of every supported recipe, rotation normalization, crop-bounds and duplicate/mismatch/malformed rejections, `model_construct`-smuggled value rejection (bools, strings, huge ints), version-first error precedence, golden row IDs, and a fresh-interpreter check proving the module imports no SDK/HTTP/imaging dependencies.
- Documented the codegen gotcha this surfaced in `docs/references/routes-dtos-and-upstream-compatibility.md`: model generation drops upstream zod refinements, so fields typed bare `UUID`/`float` may be stricter upstream.

## Why

Immich clients send edit *operations* and expect the server to produce pixels, while the Gumnut API stores one consolidated recipe on the edit version and treats `params` as opaque producer-defined JSON — so the adapter owns the recipe schema. Landing the codec as a standalone pure module gives the planned server-side bake pipeline and the planned `/assets/{id}/edits` routes a single tested semantic contract (GUM-1484); nothing calls it yet.

The composition semantics replicate upstream Immich exactly, verified against its source: the server applies crop first regardless of list position (`server/src/repositories/media.repository.ts`) and composes rotate/mirror into one affine in list order with the last element applied to pixels first (`web/src/lib/utils/editor.ts`); duplicate rejection matches upstream's per-axis mirror uniqueness (`server/src/dtos/editing.dto.ts`), so horizontal + vertical mirrors together remain legal. Every duplicate-free sequence composes to one of the eight display-orientation states, uniquely representable as `(mirror, angle)`. One deliberate divergence, noted in the code: any finite multiple of 90° normalizes modulo 360 (upstream 400s on values outside the literal `{0, 90, 180, 270}`), since the generated pydantic DTO drops upstream's zod refinement and the recipe contract calls for normalization.

## Test plan

- `uv run pytest tests/unit/utils/test_asset_edit_conversion.py` — 147 passed
- `uv run pytest` — full suite, 1917 passed
- `uv run ruff format --check`, `uv run ruff check`, `uv run pyright`, `uv run scripts/lint_docs.py` — clean


taggartgorman:
@codex review

routers/utils/asset_edit_conversion.py
```diff
@@ -0,0 +1,405 @@
...
+    return EditRecipe(crop=crop, angle=angle, mirror=mirror)
+
+
+def _synthesized_row_id(
+    asset_id: str, version_id: str, action: AssetEditAction
+) -> uuid.UUID:
+    # Newline-joined to keep the key unambiguous under concatenation. The IDs
+    # identify the synthesized current state deterministically; they do not
+    # claim to preserve user-action history.
+    return uuid.uuid5(
```
taggartgorman: **[Must Fix]** Do not introduce UUIDv5 edit-row IDs

This is the repository’s only `uuid5()` call. Ordinary adapter-owned UUIDs are generated with `uuid4()`—17 production call sites—and the remaining non-v4 values are either reproduced from Gumnut identities or explicitly encoded virtual IDs. There is no existing UUIDv5 convention to extend.

Immich is also unambiguous: `asset_edit.id` defaults to `uuid_generate_v4()`, and `AssetEditActionItemResponseSchema.id` requires `z.uuidv4()`. These synthesized UUIDv5 values therefore conflict with both the adapter’s established generation pattern and the upstream contract.

Please use a UUIDv4-compatible design. Because the IDs must also remain stable across reads, simply generating a fresh `uuid4()` on every response is insufficient: either persist stable UUIDv4 row IDs or introduce a narrowly scoped deterministic UUIDv4-formatted derivation with explicit rationale. Pin both requirements in tests: `row.id.version == 4` and fixed golden IDs for the known asset/version/actions.

tests/unit/utils/test_asset_edit_conversion.py
```diff
@@ -0,0 +1,589 @@
...
+
+class TestSynthesizedRowIds:
+    PARAMS = EditRecipe(
+        crop=CropBox(x=1, y=2, width=30, height=40), angle=90, mirror=True
+    ).to_params()
+
+    def rows(self, asset_id: str = ASSET_ID, version_id: str = VERSION_ID):
+        return recipe_to_immich_edits(asset_id, version_id, self.PARAMS)
+
+    def test_ids_are_stable_across_repeated_reads(self):
```
taggartgorman, line 529: **[Should Fix]** Pin deterministic row IDs to golden UUIDs

Calling the implementation twice proves repeatability only within the current build. Changing the namespace or key encoding would keep this test green while changing every synthesized ID across deployments.

After correcting the UUID version, assert fixed UUID literals for this known asset/version fixture. The existing differentiation tests can remain.

routers/utils/asset_edit_conversion.py
```diff
@@ -0,0 +1,405 @@
+"""Pure codec between Immich edit-action lists and the Gumnut v1 edit recipe.
+
+Immich's editor submits an ordered list of crop / rotate / mirror actions and
+expects the server to produce the pixels. Gumnut stores one consolidated,
+normalized recipe on the `edit` version's `params` — the Gumnut API treats
+`params` as opaque JSON whose schema is defined by the producer, so this module
+*is* the schema definition for `kind="edit"`. Both directions live here so the
```
taggartgorman: **[Should Fix]** Describe this as a future integration boundary

This PR has no production callers yet, but the module and exception documentation say the rendering pipeline and routes already share the codec and surface its failures. Editing remains unsupported in the current architecture.

Please phrase these as intended future integrations. No broader architecture-document update is needed until the runtime paths land.

tests/unit/utils/test_asset_edit_conversion.py
```diff
@@ -0,0 +1,589 @@
...
+            "angle": 0,
+            "mirror": False,
+        }
+        recipe = parse_recipe_params(params)
+        assert recipe.crop == CropBox(x=0, y=0, width=bound, height=bound)
+        rows = recipe_to_immich_edits(ASSET_ID, VERSION_ID, params)
+        assert isinstance(rows[0].parameters, CropParameters)
+        assert rows[0].parameters.width == bound
+
+    @pytest.mark.parametrize(
```
taggartgorman, line 362: **[Consider]** Pin version-first error precedence

The parser promises to validate the version before every other field, but unsupported versions and other malformed fields are tested only separately. Reordering the checks could therefore change the stable error code without failing the suite.

Consider adding one case with an unsupported version plus an unknown or malformed field and asserting `unsupported_recipe_version`.

taggartgorman (commented):
The affine composition, crop handling, validation boundaries, and reverse translation are otherwise well designed and thoroughly tested. The main concern is that the new synthesized edit-row IDs introduce a UUID convention that conflicts with both this repository and pinned Immich behavior; the additional comments pin the associated stability contract and correct future-facing documentation.

taggartgorman:
Addressed all four review threads in d1c7020 (all auto-fixed two-way doors; none skipped):

- **UUIDv5 row IDs (Must Fix)** — synthesized row IDs now hash the identity key with SHA-256 and stamp RFC 4122 version-4 bits, keeping determinism while satisfying upstream's `z.uuidv4()` validation and the repo's v4 convention. Rationale lives in `_synthesized_row_id`.
- **Golden-ID pinning (Should Fix)** — added `test_ids_match_golden_values` (fixed UUID literals for the known fixture) and `test_ids_are_uuid_version_4` (version/variant bits).
- **Future-tense docs (Should Fix)** — module and exception docstrings now describe the rendering pipeline and edit routes as planned callers; the PR body says the module is currently uncalled.
- **Version-first precedence (Consider)** — added a combined unsupported-version + malformed-fields case asserting `unsupported_recipe_version` wins.

A 3-iteration self-review of the fixes surfaced and fixed five more small items: `math.isfinite` is now applied only to floats (a `model_construct`-smuggled `10**400` angle previously raised `OverflowError` instead of the stable `invalid_angle` code), the uuid5-era namespace constant became a plain string key prefix, the ID-stability rationale was consolidated to one home, the params-write-once assumption behind row-ID identity is documented, and coverage was added for smuggled bool/string/huge-int angles and non-int source dims. Also documented the codegen-drops-zod-refinements gotcha in `docs/references/routes-dtos-and-upstream-compatibility.md`.

Full suite: 1917 passed; ruff/pyright/doc-linter clean.

routers/utils/asset_edit_conversion.py
taggartgorman: Fixed in d1c7020 — row IDs are now SHA-256-derived from the same identity key with RFC 4122 version-4 bits stamped: deterministic across reads/deployments, valid under upstream Immich's z.uuidv4(), and consistent with the repo's v4 convention. Rationale documented in _synthesized_row_id; pinned by row.id.version == 4 and golden-ID tests.

tests/unit/utils/test_asset_edit_conversion.py
taggartgorman: Fixed in d1c7020 — added test_ids_match_golden_values asserting fixed UUID literals for the known asset/version fixture, plus test_ids_are_uuid_version_4 for the version/variant bits. The differentiation tests remain.

routers/utils/asset_edit_conversion.py
taggartgorman: Fixed in d1c7020 — the module docstring now says nothing calls the codec yet and names the pixel-rendering pipeline and edit routes as planned callers; the UnsupportedEditRecipeError docstring says future edit routes are expected to surface it. PR body updated to match.

tests/unit/utils/test_asset_edit_conversion.py
taggartgorman: Fixed in d1c7020 — added a combined case (unsupported version + invalid angle + invalid mirror + unknown field) asserting unsupported_recipe_version wins, pinning the documented version-first check order.

taggartgorman:
@codex review

ternarybits (approved):

PR #363, opened 2026-08-20
Suppress Immich face geometry while an edited rendering is current

taggartgorman:
## What

- Add `should_expose_face_geometry` (any non-original `kind` gates, fail-safe) as the single home for the face-geometry rule, and gate every emit/write site on it: `GET /api/faces` returns `[]` for edited assets before spending face/person calls, `POST /api/faces` rejects with 409 before coordinate math, and the face sync pass gates `SyncAssetFaceV1/V2` rows.
- Sync gating is a **state transition, never an omission**: a gated face event re-emits the row with `isVisible=false` on `AssetFaceV2` and retracts it inline via `AssetFaceDeleteV1` on `AssetFaceV1` (inline rather than buffered, so a same-stream restore upsert cannot be reordered behind the retraction), and the next face event after the original is restored re-emits the visible row — each under its own acked cursor, so checkpointed clients resume past the transition. A `suppressed_face_geometry` counter surfaces gated rows in the sync summary log.
- Owning assets are resolved via `fetch_suppressed_face_ids`: one bulk `assets.list` per face event page (deduplicated ids, chunked at the bulk-id limit, `state="all"`, no `include` requested — the gate reads only lean-core `kind`), never one retrieve per face. A face whose owning asset can't be fetched is gated fail-safe with a warning.
- Document the rule canonically in `docs/references/asset-and-media-handling.md` (code docstrings point there), and record the general gating-is-a-state-transition principle in `docs/architecture/sync-stream-architecture.md`.

## Why

Gumnut stores face bounding boxes in the original upload's pixel space and keeps them there across edits. While an edited rendering is current, emitting those boxes would draw them in the wrong frame over derived pixels. A client may also already hold a face row synced before the edit, so the sync gate must actively transition that row to a hidden/retracted state rather than silently skipping events — otherwise the stale original-space geometry stays cached against the edited rendering. Suppression hides geometry only: stored faces, people associations, person thumbnails, `AssetResponseDto.people`, and face delete events are untouched, and reverting to the original restores the same geometry with no reconstruction. The backend enforces the same invariant on writes; the adapter guard gives Immich clients a stable 409 instead of a doomed upstream call.

Accepted interim-contract limitations: an edit or restore with **no subsequent face event** reaches checkpointed clients only on the next face event or full resync — closing that gap needs upstream face events emitted on rendering changes. `GET /api/faces` for a nonexistent asset now surfaces the upstream 404 rather than `[]` because the owning asset is resolved first by design.

## Test plan

- `uv run pytest` — 1945 passed; `uv run ruff check`, `uv run pyright`, and `uv run scripts/lint_docs.py` clean.
- New coverage: edited/unknown-kind gating on both REST routes (409 before any coordinate math, no `faces.create` call, no wasted face/person calls); V1 retraction and V2 hidden re-emission for edited assets, including the mixed-page partition (edited asset's face retracted while the original asset's face on the same page upserts); a checkpointed edit→restore transition test (hidden then visible, each under its own acked cursor); an ordering guard proving the V1 retraction emits inline before a same-stream restore upsert; the `suppressed_face_geometry` summary counter asserted via caplog; direct `fetch_suppressed_face_ids` tests for empty input, per-asset partitioning + fail-safe, and chunking at the bulk-id boundary; people/person sync and face delete events unaffected for edited assets; direct `should_expose_face_geometry` unit tests pinning the open kind namespace.

### DIFF
--- a/docs/references/asset-and-media-handling.md
+++ b/docs/references/asset-and-media-handling.md
@@ -62,6 +62,22 @@ archive route currently streams the current rendering regardless of `edited`.
 Mock assets must set `kind` explicitly; `make_gumnut_asset` defaults it to
 `"original"`.
 
+**Face geometry is suppressed while an edited rendering is current.** Gumnut
+stores face bounding boxes in the *original* upload's pixel space and keeps
+them there across edits, so emitting them alongside derived pixels would draw
+boxes in the wrong frame. `should_expose_face_geometry`
+(`routers/utils/asset_conversion.py`) is the single home for the rule — gate
+any new face-geometry emit or write site on it instead of re-deriving from
+`kind`. Today that gates three sites: `GET /api/faces` returns `[]` for edited
+assets (before spending face/person calls), `POST /api/faces` rejects with 409
+before any coordinate math, and the face sync pass skips
+`SyncAssetFaceV1/V2` rows via `fetch_suppressed_face_ids` — one lean bulk
+asset read per event page, never one retrieve per face, suppressing
+fail-safe when the owning asset can't be fetched. Suppression hides rows; it
+never mutates stored faces, and people associations, person thumbnails, and
+`AssetResponseDto.people` stay intact. Reverting to the original restores the
+same geometry with no reconstruction.
+
 ## Thumbnail variant selection by aspect ratio
 
 `GET /api/assets/{id}/thumbnail?size=thumbnail` normally streams the 360px `thumbnail` variant, but `_retrieve_and_stream_variant` (`routers/api/assets.py`) upgrades it to the 720px `small` variant for **wide-landscape** assets — `width > height` AND aspect ratio above `_LANDSCAPE_SMALL_ASPECT_THRESHOLD` (see the constant for the value). The Immich web timeline is a justified-rows grid that renders every row at a fixed height; a 360px-longest-edge thumbnail of a landscape asset has a height of only `360/aspect`, so the wider the asset the shorter the tile and the more visibly it softens when upscaled to fill the row. Only assets past the threshold — where the upscale is visible — get bumped. The 720px `small` keeps those panorama/ultrawide cells crisp at roughly a quarter of the pixels of the 1440px `preview`, which is far more resolution than a timeline tile needs. Portrait assets are deliberately excluded — 360px lands on their height, which already meets the row height, so they stay crisp without the extra bandwidth. `small`/`preview`/`fullsize`/`original` requests and missing/zero dims pass through unchanged (the `width`/`height` `0`-means-unknown guard from *Asset dimensions and orientation* applies here too). Video upgrades resolve to `small_image` via the existing `_image`-suffix logic — so any variant the bump can target must be a member of both the `AssetVariant` type and `_VIDEO_IMAGE_VARIANTS`, or a video upgrade resolves to a bare (non-`_image`) key that isn't in `asset_urls` and 404s. The threshold is tunable.
--- a/docs/references/asset-and-media-handling.md
+++ b/docs/references/asset-and-media-handling.md
@@ -62,25 +62,16 @@ archive route currently streams the current rendering regardless of `edited`.
 Mock assets must set `kind` explicitly; `make_gumnut_asset` defaults it to
 `"original"`.
 
-**Face geometry is suppressed while an edited rendering is current.** Gumnut
-stores face bounding boxes in the *original* upload's pixel space and keeps
-them there across edits, so emitting them alongside derived pixels would draw
-boxes in the wrong frame. `should_expose_face_geometry`
-(`routers/utils/asset_conversion.py`) is the single home for the rule — gate
-every face-geometry emit or write site on it instead of re-deriving from
-`kind` (search its callers for the gated surfaces). REST reads return no
-geometry and writes are rejected while gated. The sync stream represents the
-gate as a **state transition, never an omission**, because a client may
-already hold a face row synced before the edit: a gated face event re-emits
-the row with `isVisible=false` on `AssetFaceV2` and retracts it
-(`AssetFaceDeleteV1`) on `AssetFaceV1`, and the next face event after the
-original is restored re-emits the visible row. An edit or restore with no
-subsequent face event reaches checkpointed clients only on the next face
-event or full resync — closing that gap needs upstream face events emitted
-on rendering changes. Suppression never mutates stored state — faces, people
-associations, person thumbnails, and `AssetResponseDto.people` stay intact,
-and reverting to the original restores the same geometry with no
-reconstruction.
+**Suppress face geometry while an edited rendering is current.** Gumnut stores
+face boxes in the original upload's pixel space, so they are invalid over
+derived pixels. Gate every REST and sync emit or write site through
+`should_expose_face_geometry`: REST reads return no geometry and writes return
+409. Sync must transition existing rows rather than omit them — `AssetFaceV2`
+emits `isVisible=false`, while `AssetFaceV1` emits `AssetFaceDeleteV1`; a later
+face event after restoration re-emits the visible row. Edits and restores do
+not themselves emit face events, so checkpointed clients converge only after
+the next face event or full resync. Suppression does not mutate faces, people
+associations, person thumbnails, or `AssetResponseDto.people`.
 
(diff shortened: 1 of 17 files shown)


routers/api/sync/stream.py
```diff
@@ -428,6 +440,12 @@ async def _stream_entity_type(
                 else:
                     stats.delete_event_skips += 1
             else:
+                # Geometry-gated face row — advance the cursor without
```
taggartgorman: **[Must Fix]** Reconcile existing face rows during suppression

Omitting this upsert does not remove or hide a face row the client already holds. If face F synced while its asset was original, a later edit followed by face_updated reaches this branch and is discarded, leaving the original-space geometry cached while the edited rendering is current.

Please represent suppression as a state transition—for example, V2 isVisible=false/true and V1 delete/upsert—and add a checkpointed edit→restore test. This may require an edit/restoration event contract so both transitions are observable.

routers/api/sync/stream.py
```diff
@@ -815,6 +833,8 @@ async def generate_sync_stream(
             summary_extra["buffered_deletes"] = stats.buffered_deletes
         if stats.fk_warnings > 0:
             summary_extra["fk_reference_warnings"] = stats.fk_warnings
+        if stats.suppressed_face_geometry > 0:
```
taggartgorman, line 860: **[Should Fix]** Test the suppression summary counter

The new counter is the operator-visible evidence that rows were intentionally gated, but the behavioral tests do not inspect it. A regression in incrementing or publishing suppressed_face_geometry would leave them green.

Please capture the summary log in the two-face suppression test and assert suppressed_face_geometry == 2.

docs/references/asset-and-media-handling.md
```diff
@@ -62,6 +62,19 @@ archive route currently streams the current rendering regardless of `edited`.
 Mock assets must set `kind` explicitly; `make_gumnut_asset` defaults it to
 `"original"`.
 
+**Face geometry is suppressed while an edited rendering is current.** Gumnut
+stores face bounding boxes in the *original* upload's pixel space and keeps
+them there across edits, so emitting them alongside derived pixels would draw
+boxes in the wrong frame. `should_expose_face_geometry`
+(`routers/utils/asset_conversion.py`) is the single home for the rule — gate
```
taggartgorman: **[Should Fix]** Keep this rationale in one canonical home

The same pixel-space rationale and restoration behavior are repeated in the predicate, sync helper, and test docstrings. This paragraph also inventories “three sites,” contrary to the documentation convention against restating call-site counts.

Please retain the cross-file rationale here, trim the other copies to contracts or pointers, and replace the site count with a non-enumerated requirement or symbol-search instruction.

tests/unit/api/test_faces.py
```diff
@@ -430,6 +474,53 @@ async def test_creates_face_via_sdk(self):
         assert result.person.name == "Calvin"
         assert result.person.id == safe_uuid_from_person_id(gumnut_person_id)
 
+    @pytest.mark.anyio
+    async def test_edited_asset_rejected_with_409_before_coordinate_math(self):
+        """Drawing a face while an edited rendering is current is rejected with
+        a 409 before any scaling, and the doomed upstream create is never sent
+        — the box would be captured in the derived image's pixel space while
+        Gumnut stores boxes in the original's."""
+        from fastapi import HTTPException
```
taggartgorman: **[Should Fix]** Move this import to module scope

Both new rejection tests import HTTPException inside their bodies, but no circular dependency requires that exception to the repository’s module-scope import convention. Please import it once at the top and remove both inline imports.

taggartgorman (commented):
The shared predicate and bounded bulk lookup are sensible, and the REST behavior is well covered. The incremental sync path still needs a state transition for clients that already hold face rows; the remaining comments cover the suppression diagnostic and repository conventions.

taggartgorman:
## Review round summary

All four review items were addressed (auto-fixed, two-way door):

- **Reconcile existing face rows during suppression (Must Fix)** — Suppression is now a state transition instead of an omission: a gated face event re-emits the row with `isVisible=false` on `AssetFaceV2` and retracts it inline via `AssetFaceDeleteV1` on `AssetFaceV1`; the next face event after restore re-emits the visible row. Added a checkpointed edit→restore test plus an ordering guard proving the V1 retraction emits inline before a same-stream restore upsert. Remaining gap (noted in the doc and PR body): an edit/restore with no subsequent face event still reaches checkpointed clients only on the next face event or full resync — making those transitions observable needs upstream face events emitted on rendering changes, which is out of this repo's scope.
- **Test the suppression summary counter (Should Fix)** — The two-face suppression test now captures the summary log via caplog and asserts `suppressed_face_geometry == 2`.
- **Keep the rationale in one canonical home (Should Fix)** — The pixel-space rationale now lives only in `docs/references/asset-and-media-handling.md`; the predicate, sync-helper, conftest, and test docstrings are trimmed to contracts/pointers, and the "three sites" enumeration is replaced with a caller-search instruction.
- **Move HTTPException import to module scope (Should Fix)** — Done; both inline imports removed.

Self-review of the round (2 iterations, 3 reviewers each) additionally added the ordering guard test, corrected the "lean bulk read" docstring claim (lean once the API's lean-include default lands), extended the AGENTS.md doc-map row to mention edited-state/face-geometry suppression, and recorded the gating-is-a-state-transition principle in `docs/architecture/sync-stream-architecture.md`. Two reviewer verification notes were evaluated and deliberately not acted on: V1 checkpoint non-advancement in fully-gated windows (idempotent retractions; pre-existing delete-only-window property) and durable retraction on a transient owning-asset read miss (deliberate, logged fail-safe trade).

`uv run pytest` — 1945 passed; ruff, pyright, and the docs linter are clean.

routers/api/sync/stream.py
taggartgorman: Fixed — suppression is now a state transition: AssetFaceV2 re-emits the gated row with isVisible=false, AssetFaceV1 retracts it inline via AssetFaceDeleteV1, and the next face event after restore re-emits the visible row under its own acked cursor. Added the checkpointed edit→restore test plus an inline-ordering guard. The edit/restoration event contract (making transitions with no subsequent face event observable) is noted in the doc and PR body as an upstream follow-up.

routers/api/sync/stream.py
taggartgorman: Fixed — the two-face suppression test now captures the summary log via caplog and asserts suppressed_face_geometry == 2.

docs/references/asset-and-media-handling.md
taggartgorman: Fixed — this doc paragraph is now the canonical home (updated for the state-transition semantics); the predicate, sync-helper, conftest, and test docstrings are trimmed to contracts/pointers, and the site enumeration is replaced with a caller-search instruction.

tests/unit/api/test_faces.py
taggartgorman: Fixed — HTTPException is imported once at module scope; both inline imports removed.

taggartgorman:
@codex review

routers/api/sync/stream.py
```diff
@@ -428,6 +444,35 @@ async def _stream_entity_type(
...
+                        # this retraction after a same-stream restore upsert
+                        # of the same face id.
+                        delete_data = SyncAssetFaceDeleteV1(
+                            assetFaceId=safe_uuid_from_face_id(event.entity_id)
+                        )
+                        yield (
+                            make_sync_event(
+                                SyncEntityType.AssetFaceDeleteV1,
+                                delete_data.model_dump(mode="json"),
+                                event.cursor,
```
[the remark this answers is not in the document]
taggartgorman: Accurate that the retraction acks under `AssetFaceDeleteV1` and the pass resumes from `AssetFaceV1`, so a gated event at the tail of the stream is re-fetched each sync until a visible face upsert follows. That's the same behavior every delete in the adapter has today (see the "deletes could be lost if client resumes from the upsert checkpoint" log in phase 2) — the replay is bounded to the gated tail, not the whole face history, and is idempotent on the client. Fixing it only for the retraction would introduce a payload-type/ack-type mismatch for one event; the general delete-checkpoint gap deserves one fix covering all delete types, tracked separately.

ternarybits (approved):

PR #364, opened 2026-08-20
Render Immich edit recipes server-side from position-0 bytes

taggartgorman:
## What

- Adds `services/asset_edit_renderer.py`, which materializes normalized Immich crop, clockwise-rotate, and horizontal-mirror recipes into JPEG/PNG bytes from the asset's edit base. `pillow-heif` enables HEIC/HEIF decoding.
- Adds `routers/utils/asset_version_chain.py` with two selectors: `select_edit_base` (the highest-position version whose `kind` is not `edit`/`edit:*`), used by the renderer, and `select_root` (the position-0 upload), which `GET /api/assets/{id}/original?edited=false` keeps using so *Download original* and backups always receive the uploaded bytes.
- Adds `edit_render_*` limits for input bytes, decoded pixels, dimensions, output bytes, timeout, concurrency, and spooling.
- Adds 69 unit tests covering reference pixel transforms, EXIF orientations 2–8, format selection, real iPhone HEIC files, cap boundaries, concurrency, failures, and cleanup.

## Why

Immich's editor sends operations rather than rendered pixels. Rendering each adjustment from the edit base's exact bytes — never from a prior edit — keeps edits non-cumulative and avoids generational re-encoding loss before version creation through the Gumnut API. Preferring the latest non-edit version over position 0 preserves an `external:*` rendering layered on the upload; external renderings derive from the full chain below them, so an edit beneath one is already part of it. Until an external rendering exists, the base is the position-0 original.

Design points worth reviewer attention:

- **Resource bounds**: A dedicated bounded executor limits CPU and decoded-pixel memory without starving other asyncio work. Admission occurs before download, so queued requests retain no source spools; timed-out workers keep their slots until they exit.
- **Cleanup**: Temporary files close on success, validation or upstream failure, caller failure, timeout, and cancellation, including submit/start races and orphaned worker results.
- **Safety**: Input caps apply to both `Content-Length` and streamed bytes; image caps apply before decode; invalid crops and animated sources are rejected; and the capped writer withholds `fileno()` so Pillow cannot bypass the output limit.
- **Contract**: This change creates no version, websocket event, or HTTP route. Failures use stable `AssetEditError` subclasses, the Gumnut API remains authoritative for finalized metadata, and logs exclude signed URLs and image bytes.

## Test plan

- `uv run pytest` — 2037 passed (72 in the renderer suite, 18 in the version-chain suite)
- `uv run ruff check`, `uv run ruff format`, `uv run pyright`, `uv run scripts/lint_docs.py` — clean


services/asset_edit_renderer.py
```diff
@@ -0,0 +1,509 @@
...
+    """Decode the source into display orientation, exactly once.
+
+    Dimension and pixel caps are checked from the header before any pixel
+    data is decoded. EXIF orientation is applied to pixels here and nowhere
+    else; stale embedded thumbnails are irrelevant to the transform because
+    only the primary image's pixels are read, and the Gumnut API performs
+    authoritative metadata finalization after upload.
+    """
+    try:
+        image = Image.open(input_file)
```
taggartgorman, line 317: **[Must Fix]** Plain Pillow cannot bake supported HEIC originals

The adapter accepts HEIC/HEIF image assets and explicitly supports position-0 HEIC originals, but this passes their HEIF/HEVC bytes to plain Pillow without a registered HEIF decoder. A typical iPhone HEIC therefore raises `UnidentifiedImageError` and returns `unsupported_image`.

Add and register a HEIF decoder such as `pillow-heif`, then cover the existing real HEIC fixture with orientation and basic-transform tests.

services/asset_edit_baker.py
```diff
@@ -0,0 +1,509 @@
...
+            raise
+        with state_lock:
+            if state["abandoned"]:
+                result.body.close()
+                return None
+            state["result"] = result
+        return result
+
+    try:
+        result = await asyncio.get_running_loop().run_in_executor(
```
taggartgorman: **[Should Fix]** Bound admission before downloading each source

The executor bounds active workers but retains an unbounded queue, and each request downloads its source into a spool before submission. Under saturation, request concurrency can therefore retain an arbitrary number of downloaded inputs—up to 16 MiB memory and 100 MiB total storage each—despite `edit_bake_max_concurrency`.

Bound admission before `_download_source`. A timed-out running bake should retain its slot until its worker exits, while requests waiting for admission should time out without downloading or allocating a spool.

services/asset_edit_renderer.py
```diff
@@ -0,0 +1,509 @@
...
+                if received > max_bytes:
+                    raise EditBakeLimitError(
+                        "input_too_large", "Source image exceeds the input byte cap"
+                    )
+                destination.write(chunk)
+        except httpx.HTTPError as exc:
+            # open_cdn_response classifies only the initial send; a failure
+            # mid-stream (read timeout, dropped connection) surfaces here
+            # and needs the same stable classification.
+            logger.warning(
```
taggartgorman, line 286: **[Should Fix]** Include the asset identifier in this warning

Recording only `error_type` leaves no safe way to correlate a mid-stream failure with the affected asset. Include `asset_id` from `_bake`, either by passing it into this helper or logging after wrapping the exception. A sanitized CDN hostname would also be safe; continue excluding the signed URL.

tests/unit/services/test_asset_edit_baker.py
```diff
@@ -0,0 +1,893 @@
...
+ASSET_ID = "asset_editbaketest"
+ORIGINAL_URL = "https://cdn.example.test/signed/original"
+
+_EXIF_ORIENTATION_TAG = 0x0112
+
+
+# --- reference transforms (pure Python, independent of the baker) ---
+
+
+def rotate_cw(grid):
```
taggartgorman: **[Should Fix]** Annotate the new test helpers

The repository requires annotations on every function parameter and return type, but many helpers and fixtures in this new file are unannotated. Please annotate all new signatures; small pixel/grid aliases and protocols for the fake response objects can keep them readable.

taggartgorman (commented):
The transform pipeline, cleanup handshake, resource caps, and tests are thoughtfully implemented. Before merging, the baker needs to support the HEIC position-0 originals already accepted by the adapter, and admission must be bounded before source downloads so queued requests cannot bypass the intended resource limit.

I’ve also noted two repository-convention gaps concerning diagnostic logging and type annotations.

taggartgorman:
## Review round summary — all four items addressed (auto-fixed, two-way door)

| Feedback | What changed |
|---|---|
| **[Must Fix]** Plain Pillow cannot bake supported HEIC originals | Added `pillow-heif` and registered its opener at module import in `services/asset_edit_baker.py`. New `TestHeicSources` covers the real iPhone fixture (`tests/fixtures/livephoto/IMG_1309.HEIC`): identity bake asserting display-space dimensions/JPEG output/no orientation tag, and a crop+rotate+mirror bake pixel-compared (with JPEG tolerance) against a display-space reference decode of the same bytes. Note: pillow-heif applies the container's display transforms during decode (opened image reports orientation 1), so HEIC orientation is asserted via the reference-decode comparison; the exhaustive orientation matrix stays on the lossless PNG tests. |
| **[Should Fix]** Bound admission before downloading each source | Added an admission `asyncio.Semaphore` sized by `edit_bake_max_concurrency`, acquired before the spool/download. A timed-out running bake retains its slot until its worker exits (released thread-safely from the runner's `finally`); requests waiting for admission time out under the existing bake budget without downloading or spooling; a bake failing before its worker starts releases from the awaiting side — exactly-once in every interleaving. New tests: waiting bake downloads nothing until admitted; timed-out bakes hold then free their slot; failed download releases the slot; the queued-cancel test now widens admission explicitly to keep pinning the submit-to-start cancellation window. |
| **[Should Fix]** Include the asset identifier in the mid-stream warning | The warning now carries `asset_id` and the bare CDN hostname (`httpx.URL(...).host`), still never the signed URL; the mid-stream test asserts both fields and the URL's absence. |
| **[Should Fix]** Annotate the new test helpers | All helpers, fixtures, fakes, and inner test functions are fully annotated, with `Pixel`/`Grid` aliases keeping the reference transforms readable. |

Self-review of the round additionally: updated the stale `config/settings.py` comments (timeout now covers admission waiting; `edit_bake_max_concurrency` sizes both the pool and admission, with the decoded-memory sizing arithmetic recorded), Event-gated the two abandoned-timeout tests that previously raced wall-clock sleeps, deduplicated four copy-pasted concurrency test helpers, and added grayscale `L`/`LA` encode-preservation tests. Reference docs gained one-line notes on pillow-heif registration, signed-URL log hygiene, and admission-before-staging.

`uv run pytest` — 1995 passed (69 in the baker suite); ruff, pyright, and lint_docs clean.

services/asset_edit_renderer.py
taggartgorman: Fixed — added pillow-heif and registered its opener at module import. New TestHeicSources bakes the real iPhone fixture: identity (display-space dims, JPEG, no orientation tag) and crop+rotate+mirror pixel-compared against a display-space reference decode. pillow-heif applies the container display transforms at decode (orientation reads 1), so HEIC orientation is pinned via the reference comparison; the lossless orientation matrix stays on PNG.

services/asset_edit_baker.py
taggartgorman: Fixed — an admission asyncio.Semaphore sized by edit_bake_max_concurrency is acquired before the spool/download. A timed-out running bake retains its slot until its worker exits (released thread-safely from the runner finally); requests waiting for admission time out under the bake budget without downloading or spooling; a bake failing before its worker starts releases from the awaiting side. Each path has a dedicated test, and the queued-cancel test now widens admission explicitly to keep covering the submit-to-start window.

services/asset_edit_renderer.py
taggartgorman: Fixed — the warning now carries asset_id and the bare CDN hostname (httpx.URL(...).host), still never the signed URL. The mid-stream test asserts both fields and that the signed URL is absent from the record.

tests/unit/services/test_asset_edit_baker.py
taggartgorman: Fixed — every helper, fixture, fake, and inner test function is now fully annotated, with Pixel/Grid aliases keeping the reference transforms readable. Also deduplicated the four copy-pasted blocking-worker/run-one helpers into make_blocking_bake_sync/run_stub_bake while annotating.

taggartgorman:
@codex review

services/asset_edit_renderer.py
```diff
@@ -0,0 +1,561 @@
...
+            )
+        received = 0
+        try:
+            async for chunk in response.aiter_bytes(chunk_size=_DOWNLOAD_CHUNK_BYTES):
+                received += len(chunk)
+                if received > max_bytes:
+                    raise EditBakeLimitError(
+                        "input_too_large", "Source image exceeds the input byte cap"
+                    )
+                destination.write(chunk)
```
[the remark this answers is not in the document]
taggartgorman: Accurate observation, but leaving as-is. The writes are 64 KiB slices interleaved with `await`s on `aiter_bytes`, so no single blocking call exceeds one chunk write and the bake timeout fires at the next chunk boundary rather than being held off for the whole download. The one-time rollover copy is bounded at ~16 MiB (`edit_bake_spool_max_bytes`), and the output spool — where the heavy I/O actually happens — is already written inside the bake worker thread. Offloading each chunk write via `to_thread` would add ~1.6k thread hops per 100 MB bake for no measurable gain on local disk. Will revisit if loop-lag metrics ever correlate with bakes.

ternarybits (commented):
If there are non-edit renders after the original, why would we base the edit off of the original?  Shouldn't it be based on the most recent non-edit version of the asset?

taggartgorman:
> If there are non-edit renders after the original, why would we base the edit off of the original? Shouldn't it be based on the most recent non-edit version of the asset?

Good catch — you're right, and the fix is in 30962e5.

The renderer was hard-wired to position 0. That happened to be correct because every chain today is `original → edit*`, but it would silently throw away an `external:*` rendering layered on the upload. The base is now the highest-position version whose `kind` is not `edit` (or `edit:*`), so:

- `original → edit` → renders from the original (unchanged behaviour, edits stay non-cumulative)
- `original → external → edit` → renders from the external
- `original → edit → external` → renders from the external, which already contains the earlier edit since externals derive from the whole chain below them

Unknown kinds are treated as valid bases per the open-namespace contract.

The selection is one helper, `select_edit_base` in `routers/utils/asset_version_chain.py`, and it's shared with the `edited=false` branch of `GET /api/assets/{id}/original` so "the original" means the same version everywhere the adapter exposes it. Error codes and statuses are unchanged; tests cover all three chain shapes at the helper, renderer, and route level.

Two things this surfaced, tracked separately:

1. Once an external rendering exists, Immich's *Download original* (`edited=false`) returns that rendering rather than the upload, and the adapter has no route left that returns the uploaded bytes. That's the contract your original-download issue is asking for; I've added a note there.
2. Immich's editor previews `thumbnail?size=preview&edited=false`, and our thumbnail route doesn't honour `edited` yet — so on the *second* edit of an asset the editor shows the already-cropped render while the server crops the base. Pre-existing, not introduced here, but it's a real v1 bug; filed as High.

ternarybits:
@codex review

ternarybits:
@CharlieHelps review

ternarybits (approved):

routers/api/assets.py
```diff
@@ -230,30 +234,29 @@ async def _stream_exact_original(
...
     versions = await client.assets.versions.list(gumnut_asset_id, include=["variants"])
 
-    roots = [version for version in versions if version.position == 0]
-    if len(roots) != 1:
-        logger.error(
-            "Asset version chain has no unique root",
-            extra={"asset_id": gumnut_asset_id, "root_count": len(roots)},
-        )
+    try:
+        base = select_edit_base(versions, asset_id=gumnut_asset_id)
```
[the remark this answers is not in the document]
taggartgorman: Agreed — fixed in 960c126. `edited=false` is back on the position-0 root via a new `select_root`, and only the edit renderer uses `select_edit_base`. The two selectors now sit side by side in `routers/utils/asset_version_chain.py` with their contracts documented, plus a route test asserting *Download original* ignores an `external:*` rendering.

taggartgorman:
Correction to my earlier reply: point 1 no longer applies. A review finding made the case that `GET /api/assets/{id}/original?edited=false` is only consumed by *Download original* and backup tools — not by the editor, which previews the thumbnail route — so it should stay pinned to the uploaded root. 960c126 does that (`select_root`), while the edit renderer keeps basing on the latest non-edit version per this thread. Point 2 (the thumbnail preview route) still stands as the follow-up.

services/asset_edit_renderer.py
```diff
@@ -0,0 +1,482 @@
...
+            pass
+
+    def runner() -> RenderdEdit | None:
+        with state_lock:
+            if state["abandoned"]:
+                # Cancellation won the submit-to-start race; the awaiter owns cleanup.
+                return None
+            state["started"] = True
+        try:
+            assert input_file is not None
```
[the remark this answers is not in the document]
taggartgorman: Fixed in 1f5df97. The worker now takes the spooled source as a parameter (`runner(source)` via `run_in_executor`), so there is no captured Optional to narrow, and the `exif_transpose` assert was dead — Pillow's `@overload` already types the default call as `Image.Image`. No asserts remain in the module; pyright is clean without them.

tests/unit/utils/test_asset_version_chain.py
```diff
@@ -0,0 +1,90 @@
...
+    [
+        ("edit", True),
+        ("edit:immich", True),
+        ("original", False),
+        ("external:enhancer", False),
+        ("editorial", False),
+        ("something-new", False),
+    ],
+)
+def test_is_edit_version(kind: str, expected: bool) -> None:
```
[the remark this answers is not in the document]
taggartgorman: Stale requirement - nearly half the tests in this repo are not marked as async.

PR #365, opened 2026-08-21
Implement Immich edit routes and AssetEditReadyV2

taggartgorman:
## What

- Add `GET`/`PUT`/`DELETE /api/assets/{id}/edits`, adapting the unmodified Immich web editor to Gumnut version-chain operations: GET decodes the current `edit` version's recipe into Immich action rows; PUT validates against the edit base's display dimensions, renders server-side, and commits via `versions.append` (original current) or `versions.replace` (edit current); DELETE removes the current edit and restores the predecessor.
- Emit `AssetEditReadyV2` (`{asset: SyncAssetV2, edit: SyncAssetEditV1[]}`) plus `on_asset_update` after every committed write — including the idempotent root-current DELETE — and never on a failed or CAS-losing write. Emission is suppressed when the post-commit retrieve shows another write already superseded this one (guard keyed on `current_version_id`), so an event never pairs a newer asset row with a stale edit list; the winning write emits the current state, which also resolves the waiting client (the wait filters on `asset.id` alone).
- Add `select_current` to the version-chain selectors, bump `gumnut-sdk` to 0.159.0 (adds `versions.append`/`replace`), and document the contract in the asset-handling and websocket-events references.

## Why

- The targeted Immich web (v3.1.0) arms a 10-second wait for `AssetEditReadyV2` (filtered on `asset.id`) before sending both the PUT and the empty-edits DELETE, and treats a timeout as a failed apply — so every committed write must emit exactly one, and the editor UI was blocked on `GET /edits` existing at all (the editor panel calls it unconditionally on mount).
- Concurrency is compare-and-swap by construction: the Gumnut API accepts an append only while the original is current, and replace/delete name the snapshotted tip — a tip moved by a concurrent writer surfaces as the upstream 409 and is never retried against a refetched chain.
- Opaque (`external:*`) tips are never presented as adjustable Immich edits: GET reads them as no edits, PUT/DELETE conflict with 409. Identity-reducing edit lists are rejected with 400 rather than committing a lossy re-encode that would read back as zero edits.
- No feature flag was needed: the web editor entry is gated purely client-side, and the `editor` job-queue stub already existed.
- A slow render can still commit after the client's fixed 10-second wait has expired; this is deliberate. The commit is durable and correct, `on_asset_update` still refreshes the timeline, and a retry re-applies the same wholesale recipe idempotently — whereas a hard pre-commit deadline would make large or slow-to-fetch images permanently uneditable while only narrowing (not closing) the late-commit window.

## Test plan

- 26 unit tests in `tests/unit/api/test_assets.py::TestAssetEdits`: GET root-only/decode/malformed-recipe/opaque-tip/invalid-chain; PUT append vs replace routing, single chain snapshot shared with the renderer, opaque-tip 409, invalid crop, missing dims, identity 400, CAS-loss propagation without retry or emit, quota/storage passthrough to the global mapping, render-error status matrix, CDN-open 502, no-edit-base 502, emit ordering/payload shape, websocket-failure tolerance, superseded-write suppression (PUT and DELETE); DELETE idempotent-root emit, exact-tip delete, buried-external-survivor restore, opaque-tip 409, CAS-loss.
- Renderer-level test that a supplied chain snapshot skips the second `versions.list` call and sources bytes from the snapshot.
- 5 new `select_current` unit tests in `tests/unit/utils/test_asset_version_chain.py`.
- Full suite: 2068 passed; ruff, pyright, and the doc linter clean. `tools/dump_openapi_json.py` imports the app cleanly (its isolated env now carries the imaging deps), fixing the check-compatibility job.


routers/api/assets.py
```diff
@@ -1339,6 +1362,244 @@ async def download_asset(
...
+        logger.warning(
+            "Edit render source fetch failed",
+            extra={"asset_id": gumnut_asset_id, "error_type": type(exc).__name__},
+        )
+        raise HTTPException(
+            status.HTTP_502_BAD_GATEWAY,
+            detail="Failed to fetch source image bytes",
+        )
+
+    gumnut_asset = await client.assets.retrieve(gumnut_asset_id, include=ASSET_INCLUDE)
```
taggartgorman: **[Must Fix] Guard the event against a newer current version**

Another edit can win between this request's commit and this retrieval. In that case, this emits the newer asset row with edit rows derived from `new_version`, producing an internally inconsistent `AssetEditReadyV2`.

For a v1 guard, compare the retrieved `current_version_id` with `new_version.id`. If another write has already won, suppress this request's stale event and let the winning edit request emit the current state. Apply the equivalent predecessor check after DELETE and add focused coverage.

routers/api/assets.py
```diff
@@ -1339,6 +1362,244 @@ async def download_asset(
...
+        # "Remove all edits" is DELETE on this route (that is what Immich web
+        # sends); a non-empty list that reduces to no transformation would
+        # commit a lossy re-encode that reads back as zero edits.
+        raise HTTPException(
+            status.HTTP_400_BAD_REQUEST,
+            detail="Edit actions perform no transformation",
+        )
+
+    try:
+        async with render_asset_edit(
```
taggartgorman, line 1731: **[Must Fix] Do not commit after the client's event deadline**

Immich starts a fixed 10-second `AssetEditReadyV2` wait before issuing this request, while rendering alone can wait/run for 60 seconds. Upload, retrieval, and emission happen afterward. A saturated or large request can therefore commit after Immich has removed its listener and reported failure.

Give pre-commit work a conservative deadline below 10 seconds and ensure append/replace never starts after that budget expires. Add a deterministic test asserting that an exhausted request does not mutate the version chain.

tests/unit/api/test_assets.py
```diff
@@ -3811,3 +3811,658 @@ def test_file_modified_at_defaults_to_created(self):
...
+        self, sample_uuid, mock_current_user
+    ):
+        from routers.api.assets import apply_asset_edits
+        from routers.immich_models import AssetEditsCreateDto
+
+        versions = [self._version(position=0)]
+        client = self._client(versions)
+        fake_render, render_calls = self._fake_render()
+        with (
+            patch("routers.api.assets.render_asset_edit", fake_render),
```
taggartgorman, line 5102: **[Should Fix] Exercise the real snapshot branch**

This proves that the route passes `versions` to a fake renderer, but it cannot detect the real renderer ignoring that argument and listing the chain again.

Add a renderer-level test that calls `render_asset_edit(..., versions=snapshot)`, asserts `client.assets.versions.list` was not called, and verifies that the source URL came from the supplied snapshot.

docs/references/websocket-events-reference.md
```diff
@@ -66,6 +67,23 @@ last-updated: 2026-08-12
 
 ---
 
+### `AssetEditReadyV2`
+
+**Upstream trigger**: Emitted when the `AssetEditThumbnailGeneration` job finishes re-rendering an edited asset (`job.service.ts`), for both edit apply and edit removal.
+
+**Adapter trigger**: Emitted from `_emit_edit_committed_events` in `routers/api/assets.py` after an edits-route write commits — a successful `PUT /api/assets/{id}/edits` (append or in-place replace) or `DELETE /api/assets/{id}/edits` (including the idempotent root-current delete, which changes nothing but must still emit). Never emitted on a failed or CAS-losing write. The adapter bakes synchronously inside the request, so the event follows the HTTP response immediately rather than after a background job.
```
taggartgorman: **[Should Fix] Correct the event/response ordering**

Both edit handlers await `_emit_edit_committed_events` before returning, so emission is attempted after the version write but before the HTTP response completes—not after the response.

Please document that ordering and describe transport failure as non-fatal or best-effort. The existing 'fire-and-forget' wording in the asset reference should be updated too, because these calls are awaited.

taggartgorman (commented):
The edit-route implementation is thoughtfully structured, but two behaviors need attention: events can describe a different version than the request committed, and work can commit after Immich's fixed event wait has already expired. I also found one missing guard test and a documentation timing error.

CI currently has a separate compatibility failure while importing Pillow during OpenAPI generation; the other checks pass.

routers/api/assets.py
taggartgorman: Fixed — `_emit_edit_committed_events` now takes `committed_version_id` and suppresses both events when the retrieved `current_version_id` differs (a newer write won). PUT passes the new version id; DELETE passes the restored survivor computed from its snapshot (root for the idempotent case). Suppression is safe for the waiting client: the winner emits the current state and the web wait filters on `asset.id` alone. Covered by PUT/DELETE suppression tests plus a buried-external-survivor test.

routers/api/assets.py
taggartgorman: Skipping (approved) — the late commit is benign: it is durable and correct, `on_asset_update` still refreshes the timeline, and a retry re-applies the same wholesale recipe idempotently, so the cost is a misleading failure toast rather than corruption. A hard <10s pre-commit budget would make large or slow-to-fetch images permanently uneditable, and cannot bound the append/replace upload once started — it narrows the window without closing it. The trade-off is now recorded in the PR body.

tests/unit/api/test_assets.py
taggartgorman: Fixed — added `test_supplied_snapshot_skips_listing` at the renderer level: `render_asset_edit(..., versions=snapshot)` with a client whose own list would resolve a different chain; asserts `versions.list` is never called and the CDN open uses the snapshot version's URL.

docs/references/websocket-events-reference.md
taggartgorman: Fixed — the event reference now says the emission is awaited after the version write and immediately before the HTTP response returns, best-effort (transport failures swallowed, never fails the committed write), and documents the superseded-write suppression. Swept the stale phrasing everywhere: "fire-and-forget" reworded to awaited-but-best-effort in the asset reference, the websocket architecture doc, and both emit-helper docstrings, with the deferred video-upload path carved out explicitly.

routers/api/assets.py
```diff
@@ -1339,6 +1359,277 @@ async def download_asset(
...
+    if gumnut_asset.current_version_id != committed_version_id:
+        logger.info(
+            "Suppressing edit-committed events superseded by a newer write",
+            extra={
+                "asset_id": gumnut_asset.id,
+                "committed_version_id": committed_version_id,
+                "current_version_id": gumnut_asset.current_version_id,
+            },
+        )
+        return
```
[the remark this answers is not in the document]
taggartgorman: Fixed in d559466 — valid catch: these routes are the event's only emitter, so "the winner emits" only holds for an edit-route winner. The guard now splits on the refreshed asset's `kind`: an `edit`/`edit:*` winner still suppresses (that request emits its own consistent event, which resolves the asset-ID-filtered wait); any other winner (`original`, `external:*`) never emits, so this request emits the refreshed asset row with an empty edit list — consistent with how GET reads root and opaque tips — so the editor's wait resolves instead of timing out on a successful write. Covered by `test_put_superseded_by_external_writer_emits_refreshed_state`; the edit-winner suppression tests still pin the other side.

ternarybits:
@codex review

ternarybits:
@CharlieHelps review

routers/api/assets.py
```diff
@@ -1339,6 +1414,354 @@ async def download_asset(
...
+            extra={"asset_id": gumnut_asset_id, "error_type": type(exc).__name__},
+        )
+        raise HTTPException(
+            status.HTTP_502_BAD_GATEWAY,
+            detail="Failed to fetch source image bytes",
+        )
+
+    gumnut_asset = await _retrieve_asset_for_edit_event(
+        client, gumnut_asset_id, previous_thumbhash=previous_thumbhash
+    )
```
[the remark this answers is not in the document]
taggartgorman: Fixed — `_retrieve_asset_for_edit_event` now catches a failed post-commit read (logs with `exc_info`, returns `None`) and `_emit_edit_committed_events` skips emission and wraps payload conversion, so a durable append/replace/delete always returns success. Covered by `test_put_refresh_failure_returns_success_without_events`, `test_delete_refresh_failure_returns_success_without_events`, and `test_put_payload_conversion_failure_returns_success_without_events`; the degradation is documented under the event's Failure semantics.

routers/utils/asset_edit_conversion.py
```diff
@@ -225,7 +238,8 @@ def immich_edits_to_recipe(
                     "duplicate_action", "Duplicate rotate action"
                 )
             seen.add("rotate")
-            angle = (angle + _normalize_angle(parameters.angle)) % 360
+            rotation = _normalize_angle(parameters.angle)
```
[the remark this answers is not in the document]
taggartgorman: Fixed — `immich_edits_to_recipe` now raises `crop_not_first` (→ 400) for any crop that is not `edits[0]`, matching upstream's check; `test_rejects_crop_that_is_not_first` covers crop at index 1, 2, and mid-list, replacing the test that asserted position independence.

docs/references/websocket-events-reference.md
```diff
@@ -66,6 +67,23 @@ last-updated: 2026-08-12
...
+
+**Sent to**: Asset owner (by userId)
+**Payload**: `{ asset: SyncAssetV2, edit: SyncAssetEditV1[] }` — the refreshed sync asset row plus the complete current edit-action list (empty after a delete). See `AssetEditReadyV2Payload` in `services/websockets.py`.
+
+**Client handling**:
+- **Web**: The editor's apply flow registers a one-shot 10-second wait for this event, filtered on `payload.asset.id`, *before* sending the PUT/DELETE, and treats a timeout as a failed apply — every successful edit write must emit exactly one.
+- **Mobile**: Not used.
```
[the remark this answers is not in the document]
taggartgorman: Fixed — the summary row and Mobile bullet now describe the real handling: mobile's `applyEdits` arms the same 10s `asset.id`-filtered wait, and on receipt dispatches to `syncWebsocketEditV2` (v2 asset upsert + edit-row replace, malformed payloads dropped).

docs/references/websocket-events-reference.md
```diff
@@ -66,6 +67,23 @@ last-updated: 2026-08-12
...
+
+**Companion event**: the same commit also emits `on_asset_update` with the full refreshed `AssetResponseDto`, which the editor panel and timeline use to refresh `isEdited`, dimensions, MIME type, and URLs without a reload.
```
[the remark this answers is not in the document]
taggartgorman: Fixed — the `on_asset_update` summary row and trigger now include the edit-route companion emission (PUT/DELETE, including the idempotent root-current delete); the 'only emission path' claim is gone.

routers/api/assets.py
```diff
@@ -1339,6 +1414,374 @@ async def download_asset(
...
+    background) or the bounded deadline passes, which logs and emits anyway;
+    ``wait_for_refresh=False`` reads once, for a write that stored nothing.
+
+    Returns ``None`` when the read fails: the write is already durable, so
+    the caller skips emission rather than failing the request. Rationale:
+    ``docs/references/websocket-events-reference.md`` § AssetEditReadyV2.
+    """
+    try:
+        loop = asyncio.get_running_loop()
+        deadline = loop.time() + _EDIT_EVENT_ARTIFACT_WAIT_SECONDS
```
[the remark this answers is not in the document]
taggartgorman: Fixed in aa116f8 — the poll deadline is now an absolute loop time captured at route entry (`_edit_event_deadline`, 8.5s with margin for request latency) and passed into `_retrieve_asset_for_edit_event`, so the render, upload, and thumbhash poll all spend from the client's single 10-second budget. A commit inside the window always gets one post-commit read and emits with whatever that read returns rather than starting a fresh wait. Pinned by `test_put_slow_render_spends_the_event_deadline`; the Readiness-wait rule in the event reference explains the trade (an on-time event with a briefly stale cache key beats a failure toast on a durable write).

ternarybits (approved):
Approved, just take care of the remaining merge conflicts and comments.

routers/api/assets.py
```diff
@@ -1339,6 +1414,386 @@ async def download_asset(
...
+            )
+            if not wait_for_refresh or gumnut_asset.thumbhash != previous_thumbhash:
+                return gumnut_asset
+            if loop.time() >= deadline:
+                logger.warning(
+                    "Edit-committed event emitting before the thumbhash refreshed",
+                    extra={"asset_id": gumnut_asset_id},
+                )
+                return gumnut_asset
+            await asyncio.sleep(_EDIT_EVENT_ARTIFACT_POLL_SECONDS)
```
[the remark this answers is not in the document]
taggartgorman: Good catch on the deterministic part. The loop no longer commits to a full poll interval past the deadline — the sleep is now clamped to the remaining budget (`min(POLL, deadline - loop.time())`), so it wakes at the deadline rather than up to 500 ms after it.

I'm deliberately *not* bounding the final `retrieve` itself (wrapping it in `wait_for` and emitting a cached last-asset). The overshoot that remains is one retrieve latency, and the 8.5s deadline sits 1.5s inside the clients' 10s wait specifically to absorb that. The emission path already tolerates a degraded outcome here — if the deadline hits with a stale thumbhash it emits anyway with a warning, and the write is durable regardless — so the only symptom of an overshoot is a transient "apply failed" toast that clears on the next refresh. Bounding the read would trade real loop complexity (cancelling an in-flight SDK request, threading a last-good asset through) for closing a rare, self-healing UX gap that the margin already covers. Not worth it.

